PR #3, opened 2015-11-18
Do not disconnect on suspend

selvanair:
Openvpn no longer terminates but restarts on suspend. This conflicts with
the stop/restart logic in the gui during windows power state change events.
This patch changes the default behaviour so that SIGTERM is not triggered
during windows suspend.


mattock:
Looks good to me. I have no way to test the functionality, but I see that OpenVPN code in "release/2.3" and "master" branches has the required fixes ("Fix termination when windows suspends/sleeps"). Merging.

PR #4, opened 2015-11-18
Widen the Russian About page

SpecLad:
The first line of the page is too long for the current width, so it spills onto the second line, shifting the rest of the text downwards. As a result, the line with the URL is shifted out of the label's bounds and is no longer visible.

### DIFF
--- a/res/openvpn-gui-res-ru.rc
+++ b/res/openvpn-gui-res-ru.rc
@@ -116,7 +116,7 @@ BEGIN
 END
 
 /* About Dialog */
-ID_DLG_ABOUT DIALOG 6, 18, 249, 104
+ID_DLG_ABOUT DIALOG 6, 18, 273, 104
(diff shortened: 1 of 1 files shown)


mattock:
I'm not particularly familiar with this part of OpenVPN-GUI, so I have a question... Do you know if the "About" page static enough to allow hardcoding these values? Or will this break if the system font or system font size is changed?

SpecLad:
The measurements in dialog templates are in [dialog template units](https://msdn.microsoft.com/en-us/library/windows/desktop/ms644994(v=vs.85).aspx#measurements), which scale with the font. Otherwise, every dialog template would break upon a font change. :-)

Plus, these dialogs have a hardcoded font, so it's not going to change anyway.

mattock:
Ok, thanks for the explanation! Merging.

PR #6, opened 2015-12-12
Run with highest privilege available

selvanair:
Request the highest privilege of the invoking user. When started by a
user with admin rights, a UAC prompt may pop up requesting consent.
The current behaviour is preserved for users with limited privileges.

### DIFF
--- a/res/openvpn-gui.manifest
+++ b/res/openvpn-gui.manifest
@@ -21,7 +21,7 @@
     <security>
         <requestedPrivileges>
             <requestedExecutionLevel
-                level="asInvoker"
+                level="highestAvailable"
                 uiAccess="false"/>
         </requestedPrivileges>
     </security>


chipitsine:
the idea behind "highest privilege" is that route manipulation on Windows requires those privileges.

selvanair:
Some clarifications are in order:

Most windows installations have UAC enabled which makes all admin accounts "filtered administrators" with only limited privileges enabled by default. Except for the "Administrator" account which is disabled in most set-ups.  Thus when a windows user (even admin) starts the OpenVPN GUI, by default it launches with limited privileges and all actions that require higher privileges fail. The patch makes the GUI to start with maximum privileges that the invoking user has access to -- this will range from full admin rights to very limited rights depending on the user's profile. In most simple set-ups this is just standard user or full admin, though more fine-grained control is possible.

A vast majority of  windows users (especially home users) do have admin rights (though filtered by UAC). So requesting "highest privileges" will work the same way as "run as administrator" in may cases. Elevation will happen when possible (after a UAC prompt asking for consent unless the prompt is disabled in policy settings).  When elevation is not possible, the process will silently launch with limited user rights.

In contrast, a "run as admin" short cut or "requireAdministrator" request will completely fail for users who have no access to admin credentials.

This option is not totally non-intrusive as there may be users who do not want to give their full privileges to the OpenVPN GUI and other processes that it launches. However, such advanced users normally work using a standard user account,  in which case this is a no-op. So I believe this to be a good compromise to avoid excessive user complaints resulting from misunderstanding of UAC and a highly intrusive "requireAdministrator" setting.

mattock:
Quoting my email to openvpn-users and openvpn-devel mailing lists

>Hi,
>
>I've produced two test installers based OpenVPN 2.3.9 plus an OpenVPN-GUI which requests highest possible privileges of the user, instead of launching as the invoking user:
>
><https://github.com/OpenVPN/openvpn-gui/pull/6>
>
>The above patch should allow users with administrator privileges to launch OpenVPN-GUI without having to "Run as administrator" separately. It should also allow use of OpenVPN-GUI without having full admin rights - it is enough to have whatever OpenVPN/OpenVPN-GUI needs to function properly. A much more detailed description is available beyond the above link.
>
>The test installers are here:
>
><http://build.openvpn.net/downloads/temp/openvpn-install-2.3.9-I901-i686.exe>
><http://build.openvpn.net/downloads/temp/openvpn-install-2.3.9-I901-x86_64.exe>
>
>Let me know if the installers work correctly - or don't. I'd like to merge this functionality into OpenVPN 2.3.10 which is due early next week. If that deadline passes, I will release new Windows installers with this patch a bit later.

mattock:
I tested the x64_64 version of the I901 installer (see above) myself on Windows Server 2012 (r2) and the results were as expected:

- User with administrator privileges -> UAC prompt pops up -> VPN connections work correctly
- Normal user -> No UAC prompt -> VPN connections do not even launch (which is expected)

With the officially release 2.3.9 installers (I601) users with admin privileges do not get the UAC prompt, openvpn-gui.exe _seems_ to work, but adding routes on connection fails, so VPN does not actually work. This was expected. Behavior for normal users is the same as in the test (I901) installers.

I'll merge this fix and include it in 2.3.10 installers that will be released later today.

PR #7, opened 2015-12-15
fix cppcheck warning

chipitsine:
$ cppcheck .
Checking localization.c...
[localization.c:224]: (error) va_list 'args' was opened but not closed by va_end().


selvanair:
ACK.

PR #10, opened 2016-01-02
configure.ac: http://openvpn-gui.sf.net --> https://github.com/openvp…

chipitsine:
also, there's openvpn-gui-5 available at http://openvpn-gui.sf.net
which is a bit obsolete (sf.net says there's 308 downloads per week)
maybe we need to put redirect there ?

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -20,7 +20,7 @@ dnl  59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
 
 AC_PREREQ(2.59)
 define([_GUI_VERSION], [9])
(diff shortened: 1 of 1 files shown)


mattock:
Thanks, this change makes sense. I agree we should redirect people from SF.net to GitHub. I would also prefer disabling file downloads on SF.net, so that people don't download obsolete versions by mistake. Right now the only person who can edit the SF.net project is d12fk on #openvpn-windows channel on Freenode. I'll bug him about these issues. By the way: feel free to join that IRC channel if you're not already there!

PR #15, opened 2016-02-15
Put --log first in the command line

selvanair:
This is needed to avoid early messages going to stdout leaving no
trace of errors when openvpn exits before management interface is up.
It also ensures that any --log directives in the config do not override
the log file location.

Signed-off-by: Selva Nair <selva.nair@gmail.com>

PR #16, opened 2016-02-15
Better error reporting when connection fails to come up

selvanair:
- Handle early errors (openvpn exits before management connection is up)
    with a helpful error message that points the user to view log.
- Include only readable config files in the connection list
- Warn if no connection profiles found

TODO: handle startup errors from interactive service

### DIFF
--- a/main.c
+++ b/main.c
@@ -178,6 +178,9 @@ int WINAPI _tWinMain (HINSTANCE hThisInstance,
     exit(1);
   }
 
+  if (!IsUserAdmin())
+    CheckIServiceStatus();
+
   BuildFileList();
   if (!VerifyAutoConnections()) {
     exit(1);
--- a/manage.c
+++ b/manage.c
@@ -54,7 +54,6 @@ InitManagement(const mgmt_rtmsg_handler *handler)
     }
 }
 
-
 /*
  * Connect to the OpenVPN management interface and register
  * asynchronous socket event notification for it
@@ -205,7 +204,11 @@ OnManagement(SOCKET sk, LPARAM lParam)
             if (time(NULL) < c->manage.timeout)
                 connect(c->manage.sk, (SOCKADDR *)&c->manage.skaddr, sizeof(c->manage.skaddr));
             else
+            {
+                /* Connection to MI timed out. */
+                c->state = timedout;
                 rtmsg_handler[stop](c, "");
+            }
         }
         break;
 
--- a/openvpn-gui-res.h
+++ b/openvpn-gui-res.h
@@ -229,6 +229,8 @@
 #define IDS_NFO_RESTARTED               1708
 #define IDS_ERR_ACCESS_SERVICE_PIPE     1709
 #define IDS_ERR_WRITE_SERVICE_PIPE      1710
+#define IDS_ERR_NOTSTARTED_ISERVICE     1711
+#define IDS_ERR_INSTALL_ISERVICE        1712
 
 /* Registry Related */
 #define IDS_ERR_GET_WINDOWS_DIR         1801
--- a/openvpn-gui-res.h
+++ b/openvpn-gui-res.h
@@ -156,6 +156,8 @@
 #define IDS_ERR_CONN_SCRIPT_FAILED      1248
 #define IDS_ERR_RUN_CONN_SCRIPT_TIMEOUT 1249
 #define IDS_ERR_CONFIG_EXIST            1251
+#define IDS_NFO_CONN_TIMEOUT            1252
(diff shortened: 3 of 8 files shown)


mattock:
@d12fk: can you review this? Meanwhile I will do some testing as mentioned in comments of PR https://github.com/OpenVPN/openvpn-gui/pull/15.

mattock:
@selvanair: was this also supposed to warn the user if OpenVPN-GUI could not connect to the Interactive Service? Based on my testing this is not the case.

selvanair:
No, I left it for later as such a warning won't be appropriate for GUI distributed with 2.3. Could you please make a release branch that could be used for 2.3 distributions to allow continued bug fixes but no new features? Then we could also remove "highestAvailbable" from master but keep it in the release branch.

selvanair:
@mattock We could add a check for version number of openvpn.exe and warn about iservice if minor version >= 3 (i.e >= 2.4). But that wont work for version = 2.3_git. Is it safe to assume that the version number would always be M.m.r or M.m_git ?

selvanair:
Assuming there are no more 2.3 releases, added a commit: "warn if iservice not running" to this PR.
Rebased to master.

mattock:
We need the release branch anyways, so I'll create it tomorrow when I have more time. We can then add the "Warn if interactive service is not installed or not running" to the "master" branch and the "asInvoker" manifest modification to the release branch.

Then we have the versioning problem - we're now at version 10[.0.0.0]. Perhaps we should keep 10.0.0.0 for the release branch and use to 11.0.0.0 for the master branch. Then we can make minor releases to both. Thoughts?

mattock:
I played around with filesystem ACLs and these two features seem to work as expected:
* Include only readable config files in the connection list
* Warn if no connection profiles found
I'll test the last part later.

mattock:
@selvanair : how would I test the "Handle early errors" part?

selvanair:
@mattock: early-errors
- add an illegal option to the config file or use an empty config file (but readable)
OpenVPN will error out before management interface comes up so the GUI will appear to hang until it times out after 15 seconds. Then it should pop up a message saying check the logs. The log file should show option error. Requires the --log option first patch applied as well for this to work.

This is better than before where the message will just say "connection to xxx failed" and logs will be empty. But the user experience is still sucks ---  any suggestions for improvement?

selvanair:
@mattock: Regarding version numbers and branches:

I'm not an expert on this, but what about something like:

set current version = 10.0.x --> new branch "release/10.0"  (version = 10.0.x)
change to version = 11.0.0 on master
and then automatically increment the 3rd digit with each commit on either branch with release getting only bug fixes.

For next release (for 2.4) we could then use release/11.0 (version at that point will be 11.0.x). Then up the master to 11.1.0 ?

mattock:
How about dropping the minor number from the release branch name:
* Create a branch "release/10" for use with OpenVPN 2.3.x. Set version to 10.0 in that branch. Set version to 11.0 on current Git "master".
* When OpenVPN 2.4-alpha1 is release, create a "release/11" branch. Set version to 11.0 in that branch. Set version on the Git "master" to 12.0.

The minor version numbers in "release/10" and "release/11" would only get incremented when a new GUI is released. So far the releases have been synchronous with OpenVPN Windows installer releases.

I think we should also try to keep release branches in sync with Git "master" by cherry-picking (safe) commits.

mattock:
I can confirm that "Handle early errors (openvpn exits before management connection is up) with a helpful error message that points the user to view log." works as expected, when using an empty text file with the .ovpn suffix.

I'm fine with merging this, but I think a C-guy should have a quick look first. I'll ask if someone on the IRC could do the review.

selvanair:
Reg: release branches

Naming branches by just the major version sounds ok to me. 

> I think we should also try to keep release branches in sync with Git "master" by cherry-picking (safe) commits

At least for some time, that would be only major bug fixes for release/10 and almost everything for release/11 when that is out. 
.

mattock:
I'm merging this based on "lazy-ACK" (=nobody cared enough to review the code). Based on my tests the code in this PR works just fine.

mattock:
I updated the version in "master" and created a pull request (https://github.com/OpenVPN/openvpn-gui/pull/19) for it. I'll merge it soon unless someone complains about the versioning strategy we chose.

I also created the [release/10](https://github.com/mattock/openvpn-gui/tree/release/10) branch in my own openvpn-gui fork and cherry-picked a few recent commits into it. Let me know if it looks fine - I'd like to push it to the main OpenVPN-GUI repo.

PR #13, opened 2016-02-05
Registry and admin

selvanair:
First two commits address issues discussed in 
http://article.gmane.org/gmane.network.openvpn.devel/11051
http://article.gmane.org/gmane.network.openvpn.devel/11060

The third commit fixes the path to notepad.exe

PR #17, opened 2016-02-17
Add "Import file" menu option

leobasilio:
Hi. I think an option like this might improve the user experience. I myself didn't know I had to put the ovpn files in a "config" folder when I downloaded OpenVPN the first time. It took me a while to figure this out. The code I'm proposing simply launches an "Open File" dialog and then copy the chosen file to the correct directory.

### DIFF
--- a/README
+++ b/README
@@ -228,4 +228,8 @@ Building OpenVPN GUI from source
 
 * Start a bash shell by running msys.bat.
 
-* Run "make" from the OpenVPN GUI source directory. 
+* Run at the OpenVPN GUI source directory:
+
+$ autoreconf -i
+$ ./configure
+$ make
--- a/main.c
+++ b/main.c
@@ -29,7 +29,6 @@
 #include <prsht.h>
 #include <pbt.h>
 #include <commdlg.h>
-#include <shlobj.h>
 
 #include "tray.h"
 #include "openvpn.h"
@@ -535,11 +534,9 @@ void
 ImportConfigFile()
 {
     
-    TCHAR filter[32];
+    TCHAR filter[37];
 
-    memset(filter, 0, sizeof(filter));
-    _sntprintf_0(filter, _T("*.%s *.%s"), o.ext_string, o.ext_string);
-    filter[2 + _tcslen(o.ext_string)] = _T('\0');
+    _sntprintf_0(filter, _T("*.%s%c*.%s%c"), o.ext_string, _T('\0'), o.ext_string, _T('\0'));
     
     OPENFILENAME fn;
     TCHAR source[MAX_PATH] = _T("");
@@ -559,25 +556,34 @@ ImportConfigFile()
 
     if (GetOpenFileName(&fn)) 
     {
-        
+
         TCHAR destination[MAX_PATH];
         PTCHAR fileName = source + fn.nFileOffset;
 
-        /* make sure full config path exists */
-        SHCreateDirectoryEx(NULL, o.config_dir, NULL);
-
         _sntprintf_0(destination, _T("%s\\%s"), o.config_dir, fileName);
+
+        destination[_tcslen(destination) - _tcslen(o.ext_string) - 1] = _T('\0');
         
-        if (!CopyFile(source, destination, TRUE)) 
+        if (EnsureDirExists(destination))
         {
-            if (GetLastError() == ERROR_FILE_EXISTS) 
+
+            _sntprintf_0(destination, _T("%s\\%s"), destination, fileName);
+            
+            if (!CopyFile(source, destination, TRUE))
             {
-                fileName[_tcslen(fileName) - _tcslen(o.ext_string) - 1] = _T('\0');
-                ShowLocalizedMsg(IDS_ERR_IMPORT_EXISTS, fileName);
+                if (GetLastError() == ERROR_FILE_EXISTS)
+                {
+                    fileName[_tcslen(fileName) - _tcslen(o.ext_string) - 1] = _T('\0');
+                    ShowLocalizedMsg(IDS_ERR_IMPORT_EXISTS, fileName);
+                    return;
+                }
             }
             else
-                ShowLocalizedMsg(IDS_ERR_IMPORT_FAILED);
+                return;
+
         }
+
+        ShowLocalizedMsg(IDS_ERR_IMPORT_FAILED);
 
     }
     
--- a/main.c
+++ b/main.c
@@ -28,6 +28,7 @@
 #include <wtsapi32.h>
 #include <prsht.h>
 #include <pbt.h>
+#include <commdlg.h>
 
 #include "tray.h"
 #include "openvpn.h"
@@ -53,6 +54,7 @@
 LRESULT CALLBACK WindowProcedure (HWND, UINT, WPARAM, LPARAM);
 static void ShowSettingsDialog();
 void CloseApplication(HWND hwnd);
+void ImportConfigFile();
 
 /*  Class name and window title  */
 TCHAR szClassName[ ] = _T("OpenVPN-GUI");
@@ -349,6 +351,9 @@ LRESULT CALLBACK WindowProcedure (HWND hwnd, UINT message, WPARAM wParam, LPARAM
         ShowChangePassphraseDialog(&o.conn[LOWORD(wParam) - IDM_PASSPHRASEMENU]);
       }
 #endif
+      if (LOWORD(wParam) == IDM_IMPORT) {
+        ImportConfigFile();
+      }
       if (LOWORD(wParam) == IDM_SETTINGS) {
         ShowSettingsDialog();
       }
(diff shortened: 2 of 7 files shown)


mattock:
Hi,

This will be particularly useful now, when the [Interactive Service patch](https://github.com/OpenVPN/openvpn/commit/a24dd2e31f196c76594666f37c130817402acb15) has been merged to OpenVPN. It basically allows non-admin users to launch OpenVPN connections. With OpenVPN-GUI PR https://github.com/OpenVPN/openvpn-gui/pull/13, and commit https://github.com/OpenVPN/openvpn-gui/commit/4437ce7a8cd7e0508d26fede84fcaf5d040a8c05 in particular, OpenVPN-GUI can be configured on a per-user basis, so it's not necessarily clear to the user where the config files should reside. Using the import would solve this problem.

That said, @d12fk or @selvanair should review commits 46a1bc7 and 01051c5. I will also need to create yet another batch of test installers / executables and try this out myself.

selvanair:
I too feel this is a useful feature..

But there are some gotchas: the default config_dir is in C:\Program FIles\OpenVPN. So this file copy will fail unless run as admin. Going forward we want to avoid running admin and use interactive service instead.

We could change the default config_dir, but then that may break existing installations. Any thoughts?

leobasilio:
You're right. I'm using Windows 8 and I got the UAC prompt. I thought the admin level was required, so I didn't bother about permissions. Now I see it's actually set to highestAvailable.

Maybe we could keep the default config_dir as a sort of global directory for backward compatibility, which would be available to every user. Then we could have a config_user_dir or something like that, which would be a user directory.

leobasilio:
Regarding the comments on main.c, I chose a fixed number so I would avoid dynamic allocation. I just wanted to keep it simple. I expect an extension to be always a short string. I mean, how long could it be? It's only dangerous if the user deliberately changes the registry key.

Nice point about the _sntprintf_0 line. I'll add it to my next commit.

selvanair:
Simultaneously supporting two config dirs will require quite a lot of changes. One option I thought of was to change the default to %UserProfile%\OpenVPN\config and put a symlink there to the systemwide config_dir (read from HKLM or defaults to the old value). The GUI scans all subdirs inside the config dir, so this may work? If there are multiple configs with same name, the user's one will get chosen over the system one.

We may have to revisit this if/when the service is hardenend with some sort of admin controlled user authorization, but leave that for later.

selvanair:
> I expect an extension to be always a short string. I mean, how long could it be? It's only dangerous if the user deliberately changes the registry key

Sure, ext is going to be short, but better be safe as we allow these to be user_configurable and there are people out there who may run the GUI as admin..  Currently the ext_string var is defined as [16], so a size of 36 is safe, 32 is not.

leobasilio:
> One option I thought of was to change the default to %UserProfile%\OpenVPN\config and put a symlink there to the systemwide config_dir (read from HKLM or defaults to the old value). The GUI scans all subdirs inside the config dir, so this may work?

It sounds good to me.

> `if (!EnsureDirExists(config_dir)) return`
> is better as it handles errors

I didn't see there was this function. It's better indeed.

> Currently the ext_string var is defined as [16], so a size of 36 is safe, 32 is not.

I'll go with 37, because the string must end with 2 null characters.

> This expects the config to be one self-contained file. Will not work if certs, keys etc are in different files..

I see. I didn't know about this possibility. I'm just a guy who downloaded openvpn and thought I could make some improvements. I'm no expert on it. I'll think about an alternative.

selvanair:
>> This expects the config to be one self-contained file. Will not work if certs, keys etc are in different files

> I see. I didn't know about this possibility. I'm just a guy who downloaded openvpn and thought I could make some improvements. I'm no expert on it. I'll think about an alternative.

I am also fairly new here, and there are very few contributors with enough interest or skills to fix/improve the GUI or windows-specific parts of openvpn. Look forward to seeing more contributions and code reviews from you!

This is an excellent start for importing configs. Allowing config+dependencies could be considered later based on feedback..

Symlinks support on XP is minimal to non-existent, so I'm leaning towards your suggestion: use an extra config_dir. We could keep config_dir to point to user's profile (your code needs no change) and add a global_config_dir for system-wide configs. Will work on a patch.

selvanair:
@leobasilio: Could you please check whether your pr works when applied over pr#18 with user as non-admin and interactive service started? You may revert commit 2af8636896.. to remove the highestAvailable request before testing. Thanks.

leobasilio:
Hi. It does work. 

I would like to take the opportunity and ask for some advice. I'm new to git and github and I'm still learning the usual way to do things.

I'm planing to handle external certificates and keys by embedding them in the config file, given the user permission, or by changing relative paths to absolute ones while copying the file. That will require some parsing, and I don't want to get main.c polluted, so I would like to create a specific file for this importation feature. I would call it import.c or something like that.

I have forked this repository and made 3 commits, 2 of which I would discard in favor of the above approach. How should I handle this? Should I make a 4th commit and keep going? Should I go back to the common commit and make a branch? Or what?

Thank you.

selvanair:
Hi,

On Wed, Feb 17, 2016 at 9:10 PM, Leonardo Basilio <notifications@github.com>
wrote:

> Hi. It does work.
>
Thanks.

> I would like to take the opportunity and ask for some advice. I'm new to
> git and github and I'm still learning the usual way to do things.
>
> I'm planing to handle external certificates and keys by embedding them in
> the config file, given the user permission, or by changing relative paths
> to absolute ones while copying the file. That will require some parsing,
> and I don't want to get main.c polluted, so I would like to create a
> specific file for this importation feature. I would call it import.c or
> something like that.
>
Parsing and  embedding appears to be a good option. With interactive
service, openvpn will run as user, so the need for read access to all files
may be assumed. Which means embedding keys etc should work. The old style
of running openvpn as admin will be only for expert users who don't really
need the import feature, I guess..

Alternatively, copying all dependencies + config to a subdir of config_dir
would also an option. Probably a better one. Use one subdir per config, for
example. That was the original purpose of supporting subdirs, I think.. In
fact there are some files that cant be embedded -- up scripts, the pre and
post connect scripts used by the GUI, etc...

> I have forked this repository and made 3 commits, 2 of which I would
> discard in favor of the above approach. How should I handle this? Should I
> make a 4th commit and keep going? Should I go back to the common commit and
> make a branch? Or what?
>
Will write respond to that separately..

> Thank you.
>

selvanair:
Please contact me privately regarding git usage to keep reduce noise here..

leobasilio:
> Alternatively, copying all dependencies + config to a subdir of config_dir would also an option. Probably a better one. Use one subdir per config, for example.

I didn't consider this option because I saw there's a hard coded limit for subdirs. If you feel like that won't be a problem, then subdirs it is.

selvanair:
Max subdirs is 50 which could be increased if needed. Too many configs will anyway make the UI menu pretty unusable.

As for improvements, I suggest to proceed in steps. The current PR can be merged with a few optional changes and be useful for self-contained configs and no scripts (i.e a majority of users?):
  - optionally move the code to a new file as planned
  - copy the config to a subdir even if its just one file (will help with future iterations)
Just push the new version when ready.

Then you could consider improvements to handle more complex cases: say add a dialog so that the user can select configs and optionally dependent files like ca, cert, scripts.. Or even construct a new config base don some inputs. As in the networkmanager plugin for openvpn on linux, for example.  Better do this after reviewing how various OpenVPN-based providers distribute the configs -- as a single file, zip, or whatever.. I suppose most users use some provider -- I have never used one, so no idea.

leobasilio:
All right. It's done then. Thank you.

selvanair:
I did some run-time tests on this. Works ok, but a bit confusing for the end user:
- In case of errors it just says "Import failed" -- including the reason for failure would help.
(In my case I had an old registry entry so it was still trying to write to the system-wide location.)
- In case of success, nothing is shown, the file selection dialog just flashes away. A messagebox saying config imported would be helpful.

A single file in a subdir looks a bit confusing but I take the blame for that idea...I'm used to subdirs as we don't embed certs, keys etc., and also have a few  scripts per config. Let's see what others think.

leobasilio:
> In case of errors it just says "Import failed" -- including the reason for failure would help. (In my case I had an old registry entry so it was still trying to write to the system-wide location.)

That happened because you're using a development build. Before your PR #18, the key on HKCU was created using the system-wide directory. It also happened to me, but the end user won't have this key. Other than that, I thought that failure would be so unlikely that it wasn't worth bothering specifying the error. But that can be done.

> In case of success, nothing is shown, the file selection dialog just flashes away. A messagebox saying config imported would be helpful.

Sure. I'll take care of it.

selvanair:
On Fri, Feb 19, 2016 at 8:00 AM, Leonardo Basilio <notifications@github.com>
wrote:

> That happened because you're using a development build. Before your PR #18
> <https://github.com/OpenVPN/openvpn-gui/pull/18>, the key on HKCU was
> created using the system-wide directory.
>
Well, it happened to me because I have a messed up setup with 10 different
versions installed and I forgot to edit the registry before testing this
one :-)

But my point is that if for any reason an error occurs it would be nice to
show a helpful message to the user.

leobasilio:
I just made a new commit. Please, check it out.

selvanair:
Works well and looks good.

ACK from me.

PR #18, opened 2016-02-18
Support user and global config directories

selvanair:
Collect configs from both locations so that existing installations may not be adversely affected while new users can install configs in their user profile. As a side effect, makes PR#17 work for non-admin users.

- Set default config directory (config_dir) to %UserProfile%\OpenVPN\config
  (saved and read back from HKCU\Software\OpenVPN-GUI\config_dir)

- Add a global_config_dir variable read from HKLM\Software\OpenVPN\config_dir
  or set to OpenVPN-install-path\config

- Scan both directories and their sub directories for connection profiles.
  In case of name conflicts config_dir gets priority over global_config_dir

- Eliminate multiple warnings of duplicate configs
 
### DIFF
--- a/openvpn_config.c
+++ b/openvpn_config.c
@@ -92,7 +92,7 @@ ConfigAlreadyExists(TCHAR *newconfig)
 
 
 static void
-AddConfigFileToList(int config, TCHAR *filename, TCHAR *config_dir)
+AddConfigFileToList(int config, const TCHAR *filename, const TCHAR *config_dir)
 {
     connection_t *c = &o.conn[config];
     int i;
@@ -120,22 +120,18 @@ AddConfigFileToList(int config, TCHAR *filename, TCHAR *config_dir)
 }
 
 
-void
-BuildFileList()
+static void
+BuildFileList0(const TCHAR *config_dir, bool warn_duplicates)
 {
     WIN32_FIND_DATA find_obj;
     HANDLE find_handle;
     TCHAR find_string[MAX_PATH];
     TCHAR subdir_table[MAX_CONFIG_SUBDIRS][MAX_PATH];
     TCHAR fullpath[MAX_PATH];
-    static int warn_no_configs = 1;
     int subdirs = 0;
     int i;
 
-    /* Reset config counter */
-    o.num_configs = 0;
-
-    _sntprintf_0(find_string, _T("%s\\*"), o.config_dir);
+    _sntprintf_0(find_string, _T("%s\\*"), config_dir);
     find_handle = FindFirstFile(find_string, &find_obj);
     if (find_handle == INVALID_HANDLE_VALUE)
         return;
@@ -152,9 +148,17 @@ BuildFileList()
         match_t match_type = match(&find_obj, o.ext_string);
         if (match_type == match_file)
         {
-            _sntprintf_0(fullpath, _T("%s\\%s"), o.config_dir, find_obj.cFileName);
+            _sntprintf_0(fullpath, _T("%s\\%s"), config_dir, find_obj.cFileName);
+
+            if (ConfigAlreadyExists(find_obj.cFileName))
+            {
+                if (warn_duplicates)
+                    ShowLocalizedMsg(IDS_ERR_CONFIG_EXIST, find_obj.cFileName);
+                continue;
+            }
+
             if (CheckReadAccess (fullpath))
-                AddConfigFileToList(o.num_configs++, find_obj.cFileName, o.config_dir);
+                AddConfigFileToList(o.num_configs++, find_obj.cFileName, config_dir);
         }
         else if (match_type == match_dir)
         {
@@ -163,7 +167,7 @@ BuildFileList()
             &&  subdirs < MAX_CONFIG_SUBDIRS)
             {
                 /* Add dir to dir_table */
-                _sntprintf_0(subdir_table[subdirs], _T("%s\\%s"), o.config_dir, find_obj.cFileName);
+                _sntprintf_0(subdir_table[subdirs], _T("%s\\%s"), config_dir, find_obj.cFileName);
                 subdirs++;
             }
         }
@@ -195,7 +199,8 @@ BuildFileList()
 
             if (ConfigAlreadyExists(find_obj.cFileName))
             {
-                ShowLocalizedMsg(IDS_ERR_CONFIG_EXIST, find_obj.cFileName);
+                if (warn_duplicates)
+                    ShowLocalizedMsg(IDS_ERR_CONFIG_EXIST, find_obj.cFileName);
                 continue;
             }
 
@@ -204,9 +209,22 @@ BuildFileList()
 
         FindClose(find_handle);
     }
-    if (o.num_configs == 0 && warn_no_configs)
-    {
-        ShowLocalizedMsg(IDS_NFO_NO_CONFIGS, o.config_dir);
-        warn_no_configs = 0;
-    }
+}
+
+void
+BuildFileList()
+{
+    static bool issue_warnings = true;
+
+    o.num_configs = 0;
+
+    BuildFileList0 (o.config_dir, issue_warnings);
+
+    if (_tcscmp (o.global_config_dir, o.config_dir))
+        BuildFileList0 (o.global_config_dir, issue_warnings);
+
+    if (o.num_configs == 0 && issue_warnings)
+        ShowLocalizedMsg(IDS_NFO_NO_CONFIGS);
+
+    issue_warnings = false;
 }
--- a/main.c
+++ b/main.c
@@ -170,6 +170,8 @@ int WINAPI _tWinMain (HINSTANCE hThisInstance,
   if (!GetRegistryKeys()) {
     exit(1);
   }
+  EnsureDirExists(o.config_dir);
+
   if (!CheckVersion()) {
     exit(1);
   }
--- a/options.h
+++ b/options.h
@@ -125,6 +125,7 @@ typedef struct {
     /* Registry values */
     TCHAR exe_path[MAX_PATH];
     TCHAR config_dir[MAX_PATH];
+    TCHAR global_config_dir[MAX_PATH];
     TCHAR ext_string[16];
     TCHAR log_dir[MAX_PATH];
     TCHAR priority_string[64];
--- a/registry.c
+++ b/registry.c
@@ -73,8 +73,15 @@ GetRegistryKeys()
   if (openvpn_path[_tcslen(openvpn_path) - 1] != _T('\\'))
     _tcscat(openvpn_path, _T("\\"));
 
+  /* an admin-defined global config dir defined in HKLM\OpenVPN\config_dir */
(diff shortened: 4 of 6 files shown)


selvanair:
Note: This changes the default value of config_dir in registry, so to test this, make sure you don't have a predefined registry entry in HKCU\Software\OpenVPN\config_dir

selvanair:
Looks like the comment linked to commit got hidden by the force push. Here goes again:
 
Rebased to master.

Notes on testing:

- Revert the highestAvailable commit
- As this changes the default location of config_dir, please delete HKCU\Programs\OpenVPN-GUI (or just the config_dir key) before testing it. The HKLM keys set during service install are ok (i.e., point to the system wide location.)
- Start the GUI and configs in both system-wide config_dir and users config dir (if any) will appear in the menu. If the user config_dir does not exist it will be created on first use as OpenVPN\config in user's home
- Add some config files (anything.ovpn will do) to the either location and see all appear in the menu
Add a config with same name to both locations and see only one appears in the menu (duplicate warning will be shown only if the GUI is restarted)

selvanair:
@leobasilio: Could you please review this ?

openvpn_config.c
```diff
@@ -153,6 +149,14 @@ BuildFileList()
...
+
+            if (ConfigAlreadyExists(find_obj.cFileName))
+            {
+                if (warn_duplicates)
+                    ShowLocalizedMsg(IDS_ERR_CONFIG_EXIST, find_obj.cFileName);
+                continue;
+            }
+
             if (CheckReadAccess (fullpath))
                 AddConfigFileToList(o.num_configs++, find_obj.cFileName, o.config_dir);
```
leobasilio: It should be "config_dir", not "o.config_dir"

openvpn_config.c
```diff
@@ -153,6 +149,14 @@ BuildFileList()
         if (match_type == match_file)
         {
             _sntprintf_0(fullpath, _T("%s\\%s"), o.config_dir, find_obj.cFileName);
```
leobasilio: It should be "config_dir", not "o.config_dir"

openvpn_config.c
selvanair: Thanks. Bad merge with PR#16. Will force push a fixed version soon.

selvanair:
pushed fixup

leobasilio:
I think it's ok now.

On another topic, I'm getting this message when compiling:

>../openvpn-gui/main.c:186:5: warning: implicit declaration of function 'CheckIServiceStatus' [-Wimplicit-function-declaration]
>     CheckIServiceStatus();
>     ^

CheckIServiceStatus() is missing at service.h.

selvanair:
Squashed all fixups and pushed a clean version.

And, thanks for the pointer to missing declaration -- as its a small fix I included it in the same commit.

mattock:
@snair: we have the release/10 branch now, so you can now safely revert the "highest-available" patch (2af8636) in the "master" branch.

mattock:
This PR seems to work as advertised. Two questions though:
* Is _%UserProfile%\OpenVPN\config_ the "most correct" directory for user config files? It is rather invasive imho. What about _%UserProfile%\.openvpn\config_ instead? It seems that various Windows applications (JRE, SSH) store user configs under a dot directory already, plus that's a standard common practice on *NIX.
* Is a user in local Administrator group supposed to be able to launch connections from his/her home directory? In my case that failed, as was evident from the EventLog. I had to add the  "OpenVPN Administrators" group, and add my user to the group before connections defined under _%UserProfile%_ started to work. I'm just worried about the amount of support requests if user-specific configs do not work by default.

Thoughts?

leobasilio:
* I think the dot notation is usual for cross-platform applications, like the ones you mentioned. This GUI is a Windows only application. In my opinion, the most appropriate path would be %AppData%, but on the other hand it's unknown to most users. After my pull request #17, the user should be able to import config files using the GUI. So storing them in %UserProfile% would keep them visible to the user, making s/he promptly aware of their location.
* I'm new here, so I'm not sure I get what you mean, but as far as I know, the GUI only forwards the config path to openvpn as a command line argument, doesn't it? What exactly failed in your case?

selvanair:
- Location of config: hiddden files/folders are ideal for configurations that a user normally need not access directly. In our case, the user still has to copy files into the config directory: currently "Import File" can handle only single self-contained config with no dependencies, no scripts. Until we have a better import mechanism for configs, hiding the location may not be a good idea? But I don't have a strong preference either way. If we go for a hidden location, a dotfile (folder) in user's profile looks far better than %AppData%

- With the latest interactive service in git master, user in Administrators group should be allowed to start any config. I'm not sure why it failed for you. I'll do some tests again.

- User complaints and confusion due to the policy restrictions added to the interactive service is a concern. I'm working on a few patches for the GUI to mitigate this:

(i) Check group membership and show a helpful message when a connection is attempted on a config that would otherwise fail. The message could be "Add yourself to OpenVPN Administrators" group.
OR Disable non-startable configs -- which is better?
(ii) Listen on the service pipe for errors from interactive service and show them on the status window. Currently only the event log gets them although the service writes errors to both the event log and the pipe. This would have other nice side effects --- e.g.,, no need to wait for 15 seconds to know the connection failed due to an option error.
 
Both (i) and (ii) can be made available in short order.

We could have a permissive mode where all users are treated as "OpenVPN Administrators". This could be achieved by setting ovpn_admin_group (in registry) as, say, "Users" at the service installation time.

mattock:
@selvanair

(i): "Add yourself to OpenVPN Administrators group" would be more educational and would cause less confusiion imho.
(ii) sounds good - the Event Log is probably a bit too deep for the Joe Average.

We probably should ask others developers and core support people about the "permissive mode". In a home enviroment having to (manually) add yourself to a group to use OpenVPN would be a nuisance.

As for failing to use OpenVPN as a user that belong to the administrator group - in my case the group name was a localized one ("Järjestelmänvalvojat"), not "Administrators". Does this make a difference here? I'm pretty sure I used the most recent Interactive service, but I will retest if necessary.

selvanair:
@mattock: Thanks for suggestions on (i) and (ii) -- I'll make some patches this weekend.

Localized or customized group names would have caused the check to fail as currently the service only checks for the group name "Administrators" and whatever is in the ovpn_admin_group key in registry (defaults to "OpenVPN Administrators", if not defined).  I'll do a patch for the service to use the SID of Administrators group for the check.

selvanair:
I've a fork of openvpn  [at github](https://github.com/selvanair/openvpn/tree/admin-group) with support for localized Administrators group (branch named admin-group) . But I've no good way of testing this[*] -- could some one with localized windows test this out please? To test, be admin user but start the GUI without admin priv and remove yourself from "OpenVPN Administrators" group. And try start a config in the user's profile. In case of successful connection the event log should show why the connection was authorized (something like "by virtue of membership in so and so group"..)

If it works, I'll post a patch to the devel list.

mattock:
@selvanair: I can test OpenVPN with the localized fix early next week.

leobasilio:
Hi. I was trying to test this and found 2 other issues. When connecting with config from my user directory, the command line on StartOpenVPN() was something like this:

openvpn --log "C:\Program Files\OpenVPN\log\test.log" --config "test.ovpn" ....

* I think there should be also a global_log_dir, to keep logs and configs both in %UserProfile%.
* --config should be followed by a full file path, as OpenVPN itself doesn't handle user configs.

I can't tell anything about the localized Administrators group yet. Whenever I try to start a connection the status window will pop up for a while and then it closes saying: "Connecting to management interface failed.". I'm still trying to figure this out.

selvanair:
@leobasilio  Thanks for the comments. 
With the latest GUI log defaults to %UserProfile% so for both global and local configs log should go there. Otherwise log will not be writeable when user is not admin. I guess you may have some old settings in the registry which is overriding the default. Please test again after deleting HKCU\OpenVPN key and subkeys so that the new defaults will populate it again. 

That timeout error appears to show the connection failed at the service end -- especially if there is nothing in the log. If so please check the event log. Also can you verify that "net localgroup Administrators" lists your username --- you'll have to use the localized name of "Administrators" in that command..

leobasilio:
I see. Here's what I got on my events log:

When starting the service:

* Error querying registry value: HKLM\SOFTWARE\OpenVPN\ovpn_admin_group

When trying to connect:

* Authorizing user Leonardo by virtue of membership in group Administradores
* openvpnserv error: 0x00000002 CreateProcessAsUser *(some message like: file not found)*

I hope it helps. I can try again tomorrow.

selvanair:
The first error is expected as there wont be any such registry entry by default -- the code will use the "OpenVPN Administrators" in that case which is also irrelevant for this test.

The authorization success shows the patch works as it detects the localized admin group name. But I have never seen the interactive service error at CreateProcessAsUser().. Does the service work on your machine at all?  One way to check that would be to use openvpn from current git master -- it won't recognize the localized Administrators group, so add the user to "OpenVPN Administrators" group. Or use a commit before f3c8a04d602 so that no group membership is needed.

Thanks.

selvanair:
1. Correction: the registry location for the GUI is HKCU\Software\OpenVPN-GUI
2. The file not found error appears to be caused by the service not finding the openvpn executable. Make sure the registry settings made under HKLM\OpenVPN\ during the service install time are correct -- especially exe_path -- the log_dir there will point to the global location, which is fine  as its not used by the GUI.

leobasilio:
Hi. Once again, I had a bad registry key. After fixing it, everything works fine now. So, in the end, my only remark would be to use a full config path when putting together the command line call.

selvanair:
Thanks for the test. 

Not sure I understand the remark about full config path. The GUI passes the config_dir and the config file name relative to it, and the service checks the full path to see whether its authorized. I don't see any benefit of passing the full path. Even if we pass the the full path, the workdir has to be still specified for the config_dir for scripts, non-inlined files etc to work with relative paths.

Am I missing something?

leobasilio:
Oh, I didn't realize that the working dir would be config_dir. I'm still getting to know the code. In that case, it's all right to me.

selvanair:
Yeah, the workdir is set by the service while starting openvpn -- the service takes three inputs from its clients (GUI in our case): the workdir, command line options for openvpn, and any input to pass to stdin. We pass config_dir for work_dir, the usual options including "--config confg_file" and the management password for stdin-input. 

Note that the config_dir used here is the actual root of the config file, not necessarily o.config_dir because subdirectories are allowed.

selvanair:
> @selvanair: I can test OpenVPN with the localized group name fix early next week.

@mattock : Based on leobasilio's tests we got the "localized group name fix" for the service merged into openvpn git master. Could you please recheck this PR with the latest interactive service? Admin user should be allowed to start configs in user profile without the need for the special group membership.

mattock:
@selvanair: yes, I'm building the installers right now. If all goes well, I think we can merge this one.

mattock:
There were some issues in PR https://github.com/OpenVPN/openvpn-gui/pull/26 (see comments), but this one seems solid, so I'll merge this.

mattock:
Travis claimed that after merging this PR the test build failed. Based on Travis' logs it was apt-get timing out at build environment setup time.

PR #19, opened 2016-02-22
Bump version in "master" branch to 11

mattock:
URL: https://github.com/OpenVPN/openvpn-gui/pull/16

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -19,7 +19,7 @@ dnl  distribution); if not, write to the Free Software Foundation, Inc.,
 dnl  59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
 
 AC_PREREQ(2.59)
-define([_GUI_VERSION], [10])
+define([_GUI_VERSION], [11])
 AC_INIT([OpenVPN GUI],[_GUI_VERSION],[openvpn-devel@lists.sourceforge.net],[openvpn-gui],[https://github.com/openvpn/openvpn-gui/])
 AC_DEFINE([PACKAGE_VERSION_RESOURCE], [_GUI_VERSION,0,0,0], [Version in windows resource format])
 AC_CONFIG_AUX_DIR([.])


selvanair:
Naming the branches by major only is fine, but shouldn't the _GUI_VERSION be at 11.0 assuming we may increase the minor number for bug fix releases? Else, programmatically it wouldn't be possible to be figure the full version number of a release.

selvanair:
Forgot to add:
release/10 looks good.

The commit 791aea49e6cdab83c5ac9ca203104859a57f4260 (do not connect to the interactive service as admin) should got to this one too. Though interactive service is not in 2.3, the purpose of that one was to protect against rogue programs acting as iservice.  For clarity, I can make a separate PR for it with an appropriate commit message, if that's better.

mattock:
I think GUI_VERSION should be 11. In configure.ac we would have this:

    define([_GUI_VERSION], [11])
    AC_INIT([OpenVPN GUI],[_GUI_VERSION],[openvpn-devel@lists.sourceforge.net],[openvpn-gui],[https://github.com/openvpn/openvpn-gui/])
    AC_DEFINE([PACKAGE_VERSION_RESOURCE], [_GUI_VERSION,0,0,0], [Version in windows resource format])

The PACKAGE_VERSION_RESOURCE will be shown in the file properties of openvpn-gui.exe. We can simply increment the minor versions at release time, e.g. go from 11.0.0.0 to 11.1.0.0. I assume Microsoft has some documentation on how the four digit versioning scheme is supposed to be used.

I think the only issue with the "major version only" approach is that distinguishing between dist tarballs with differing minor version is not possible. This is not a big deal imho.

Thoughts?

selvanair:
My bad, I didn't see PACKAGE_VERSION_RESOURCE is in config.h and so accessible inside the code.  And also could be queried using GetFileVersionInfo() and related API, I suppose.

Ready to roll then...

> I think the only issue with the "major version only" approach is that distinguishing between dist tarballs with differing minor version is not possible

The dist-gzip tarball for the GUI?  If needed we could decorate the tarball-name in AC_INIT with, say the git tag: openvpn-gui-m4_easycmd_s([git describe --tags]) or something like that. Will need some tags like 10.1 to be present in the branch. Only relevant if the GUI standalone source tarball is distributed.

mattock:
The openvpn-build buildsystem fetches openvpn-gui-\<version\>.tar.gz from a webserver. The \<version\> can be whatever we want, as long as the name in the tarball and the source directory name match. If they don't match, openvpn-build will choke trying to "cd" into the wrong directory.

I'll merge this pull request and then push the release/10 branch.

mattock:
The release/10 branch has been pushed.

PR #20, opened 2016-02-26
Adding static-challenge support

soarqin:
I also copied all usable translations in dialog resources, there is only 1 new string for translation: "Challenge"


soarqin:
okay, I will fix these asap

soarqin:
@selvanair For TCHAR type, I just follow what GetDlgItemTextUtf8() does in misc.c, which treats LPTSTR as LPWSTR, so I will also fix this function.

selvanair:
Hmm.. If there are many such places where TCHAR is assumed to be WCHAR, we could just leave it as is and can do a patch if required later? Samuli may know whether anyone compiles this without -D_UNICODE

@samuli  can we just put a preprocessor check UNICODE and _UNICODE are defined and bail out otherwise?

soarqin:
re-commited, supported ECHO field, moved base64 implementation to win32 cryptoapi (I do think we could also use OpenSSL if don't want crypt32.lib)

selvanair:
Looks good and works as expected. ACK.
Closes [#525](https://community.openvpn.net/openvpn/ticket/525)

Note: The max length of the static-challenge text that displays correctly is about 40 chars depending on word wrap. This should be adequate for most users?

soarqin:
40 chars are enough in most scene, challenge system is usually for MFA-like auth which sends very simple words.

selvanair:
Good enough for me: I use "Enter google authenticator code" :-)

mattock:
@selvanair: I don't know of anyone building openvpn-gui without unicode support. We probably have no way to know that except by requiring unicode.

Anyways, based on my smoketesting ("builds and other functionality does not seem to break") and Selva's more extensive tests I'll merge this.

PR #25, opened 2016-03-05
some minor fixes

chipitsine:
1) openssl-1.0.1r (as in openvpn-build repo)

2) moved common variables to "env" section
3) fixed implicit declaration of function 'CheckIServiceStatus'

main.c:179:5: warning: implicit declaration of function 'CheckIServiceStatus' [-Wimplicit-function-declaration]
     CheckIServiceStatus();
     ^

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,11 @@ os: linux
 language: cpp
 compiler: gcc
 
+env:
+  global:
+    - OPENSSL_CRYPTO_CFLAGS="-I${HOME}/image/include"
+    - OPENSSL_CRYPTO_LIBS="-L${HOME}/image/lib -lcrypto"
+
 matrix:
   fast_finish: true
   include:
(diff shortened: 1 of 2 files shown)


selvanair:
Last three commits should be replaced by a single one that defines

BOOL CheckIServiceStatus(void) in service.h
Having said that, this is handled in PR#18

Aside: On windows BOOL is defined once windows.h is included and is an int. We also have bool defined in main.h (as int) although we should just use BOOL everywhere. Note that bool is a valid C++ type where it need not be an int.

cron2:
Hi,

On Sat, Mar 05, 2016 at 08:35:11AM -0800, selvanair wrote:
> Aside: On windows BOOL is defined once windows.h is included and is an int. We also have bool defined in main.h (as int) although we should just use BOOL everywhere. Note that bool is a valid C++ type where it need not be an int.

Does Windows have <stdbool.h>?  We changed all the "openvpn proper" code
to use that - so compilers can do "proper bool" instead of emulation 
using int - and while it was the right thing to do it uncovered a few
interesting bugs where "legacy code" was storing more than 0/1 in a bool
variable...

gert
-- 
                                                           //www.muc.de/~gert/

chipitsine:
it seems to be ok to use c99 "stdbool.h" instead of defining "bool", however, I didn't check that for MSVC build

selvanair:
stdbool.h is C99, isn't it? I don't use visual studio, but probably VS 2017 has it ? :-) mingw-w64 has it, at least in the version I have here.

Though we have bool = int, BOOL and int used to return true/false, switching to a real "bool" shouldn't be hard. For now, we get BOOL directly from windefs.h (included in windows.h) so that appears to be the easiest choice.

chipitsine:
I mean  https://github.com/OpenVPN/openvpn-build/tree/master/msvc  
it is not VS2017

selvanair:
2017 was a joke :-) I suppose VS2015 supports C99.

Your three commits in which first <stdbool.h> is added, then bool is redefined to int makes no sense to me. All that is needed to fix the missing declaration is "BOOL CheckIServiceStatus(void)" in service.h and that is pending in PR#18. We have BOOL and bool everywhere in the code and neither of them are the same as the C99 bool in <stdbool.h>.

Having said that its fine to use stdbool.h if available, but then the patch should address all uses of BOOL and bool and int to replace them with the C99 bool. And that has to be done conditionally on HAVE_STDBOOL_H is defined or not.

chipitsine:
ok, do whatever you want to. I'm closing this

PR #26, opened 2016-03-07
handle iservice policy restrictions

selvanair:
This is not fully ready for a pull request but I need some feedback. So here goes:

It includes #PR18 as the first commit as without that the rest doesn't make much sense -- so ignore that commit

The second commit adds some functionality to check whether a config will be treated as authorized by the interactive service. If not it offers and option to add the user to the "OpenVPN Administrators" group (or its equivalent). Is this a good idea? If it fails to add the user to this group should we start the connection without using the service (i.e default to the old behavior)? Note that such an action may be useful for some connections that could be established without admin rights.

This is supposed to be tested with the latest openvpn git master and the HKLM registry keys required for the service present. Also the group "OpenVPN Administrators" should exist. Do not add any users to it. This may be easily done using "net localgroup "OpenVPN Administrators" /add".

Thanks.

### DIFF
--- a/access.c
+++ b/access.c
@@ -1,0 +1,312 @@
+/*
+ *  This file is a part of OpenVPN-GUI -- A Windows GUI for OpenVPN.
+ *
+ *  Copyright (C) 2016 Selva Nair <selva.nair@gmail.com>
+ *
+ *  This program is free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation; either version 2 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program (see the file COPYING included with this
+ *  distribution); if not, write to the Free Software Foundation, Inc.,
+ *  59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+ */
+
+#ifdef HAVE_CONFIG_H
+#include <config.h>
+#endif
+
+#ifndef SECURITY_WIN32
+#define SECURITY_WIN32
+#endif
+
+#include <windows.h>
+#include <shellapi.h>
+#include <lm.h>
+#include <stdlib.h>
+#include <security.h>
+#include "main.h"
+#include "options.h"
+#include "service.h"
+#include "localization.h"
+#include "openvpn-gui-res.h"
+
+extern options_t o;
+
+#define MAX_UNAME_LEN (UNLEN + DNLEN + 2) /* UNLEN, DNLEN from lmcons.h +2 for '\' and NULL */
+
+/*
+ * Check whether current user is a member of specified groups
+ */
+static BOOL
+CheckGroupMember(DWORD count, WCHAR *grp[])
+{
+    LOCALGROUP_USERS_INFO_0 *groups = NULL;
+    DWORD nread, nmax, err;
+    WCHAR username[MAX_UNAME_LEN];
+    DWORD size;
+    DWORD i, j;
+    BOOL ret = FALSE;
+
+    size = _countof (username);
+    /* get username in domain\user format */
+    if (!GetUserNameExW (NameSamCompatible, username, &size))
+        return FALSE;
+#ifdef DEBUG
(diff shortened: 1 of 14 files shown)


mattock:
I did pretty extensive testing of this PR combined with PR https://github.com/OpenVPN/openvpn-gui/pull/18. This PR worked as advertised, except for one detail: if the group name was customized, the authorization check failed consistently. The test setup was this:
* An unprivileged user (no admin rights) was used
* _HKLM:\\SOFTWARE\OpenVPN\ovpn_admin_group_ was set to _dummygroup_
* The group ("dummygroup") was created with _net localgroup dummygroup /add_

With these settings all connection attempts failed, and nothing appeared in the connection log (e.g. $HOME\OpenVPN\log\test.log).

If the user was in both "dummygroup" and "OpenVPN Administrators" group, then configs under $HOME\OpenVPN\config worked fine. So the customized group name was basically ignored.

Similarly, if ovpn_admin_group key was missing, and the user was in "OpenVPN Administrators" group, then things worked fine.

OpenVPN-GUI was also able to add the user to the ovpn_admin_group correctly, regardless of what its name was ("dummygroup", "OpenVPN Administrators"...). So it was only the authorization part that failed if ovpn_admin_group was customized.

mattock:
Oh, and the installer containing this PR and PR https://github.com/OpenVPN/openvpn-gui/pull/18 is this:
* http://build.openvpn.net/downloads/temp/openvpn-install-2.3_guipr18and26-I607-x86_64.exe

selvanair:
@mattock Regarding the use of customized ovpn_admin_group : your tests show the GUI recognizes it but the service doesn't. I'll check this, but a quick question. Was the service restarted after the registry change? The registry is scanned by the service only at startup.

selvanair:
> With these settings all connection attempts failed, and nothing appeared in the connection log (e.g. $HOME\OpenVPN\log\test.log).

We currently do not have any support for reading errors from the service. So in case of errors where openvpn is not even started the user gets no feedback. Fixing this is in my TODO list -- hopefully the last piece before we're ready for 2.4 alpha release...

mattock:
@selvanair I probably did not restart the OpenVPNServiceInteractive after the ovpn_admin_group change. That would explain the behavior nicely.

selvanair:
Thanks. I tested it again, and it should work if interactive service and GUI are restarted after registry changes. Group membership is dynamically checked in the both the service and the GUI, but the registry is scanned only once.

More tests by others and a technical review of the code will be useful.

mattock:
@selvanair Great! We need to document that behavior properly. Plus we should probably offer to define the ovpn_admin_group at install time, before enabling and starting the interactive service.

selvanair:
Yes, a number of things to document related to the interactive service and recent changes to the GUI.

Obviously, this patch is meant for users technically-challenged to add themselves to a group. However, it'll unconditionally fail if the group doesn't exist. So, I think, the installer should create the group by default. As for registry keys,  a number of keys are setup as a part of the service installation. We have to add one more (ovpn_admin_group) to that list. Setting up this key and adding the group as such do not grant any new rights to users, so admins need not be concerned. We need not offer to add any user to the group at install time.

Having said that, if desired, we could add the functionality to create the group as well to this patch... That'll complicate the patch a bit, but could be done.

selvanair:
Rebased to master and pushed with some improvements
 - Add the ovpn_admin_group if it doesn't exist
 - Do not allow adding the user to built-in admin group
 - Protect against parallel execution of net command from multiple threads
See the commit message for details

Can anyone provide a code review please?

mattock:
@d12fk @leobasileo @ValdikSS : any takers for a code review?

leobasilio:
I don't feel I know the code that well yet to review this. Sorry.

access.c
```diff
@@ -0,0 +1,312 @@
...
+            {
+                ret = TRUE;
+                break;
+            }
+        }
+        if (ret)
+            break;
+    }
+#ifdef DEBUG
+    PrintDebug(L"User is %s in an authorized gtoup", ret? L"" : L"not");
```
ValdikSS: typo

ValdikSS:
I'm not really a C programmer and very rarely use Windows. It would be great if someone with Windows Security knowledge will check this.

access.c
selvanair: Thanks. Will fix.

mattock:
I asked somebody to review this on openvpn-devel ml. If that fails I'd say we merge this as-is to get things moving forward.

selvanair:
Fixed the typo that valdikss pointed out and also added a sanity check on the group name.

Looks like we may have to merge this as is, and handle any feedback or review as they come..

mattock:
Yes, I'll merge this based on lazy-ACK.

PR #32, opened 2016-03-20
Fix potential out-of-bounds access

Chocobo1:
found by cppcheck.

Before this fix, `o.auto_connect[i]` will be evaluated before `i < MAX_CONFIGS`.


selvanair:
ACK.

PR #33, opened 2016-04-04
Some small bug fixes

selvanair:
Broken into 4 commits: only the 3rd one fixes a bug that is easy to reproduce (try an empty password with static challenge).
@soarqin could you please check it doesn't break the rest of the static challenge code?

### DIFF
--- a/misc.c
+++ b/misc.c
@@ -343,7 +343,8 @@ BOOL IsUserAdmin(VOID)
                                   &AdministratorsGroup);
     if(b)
     {
-        CheckTokenMembership(NULL, AdministratorsGroup, &b);
+        if (!CheckTokenMembership(NULL, AdministratorsGroup, &b))
+            b = FALSE;
         FreeSid(AdministratorsGroup);
     }
 
--- a/misc.c
+++ b/misc.c
@@ -33,27 +33,42 @@
 #include "options.h"
 #include "manage.h"
 #include "misc.h"
+#include "main.h"
 
 /*
  * Helper function to do base64 conversion through CryptoAPI
+ * Returns TRUE on success, FALSE on error. Caller must free *output.
  */
-static void
+static BOOL
 Base64Encode(const char *input, int input_len, char **output)
 {
     DWORD output_len;
+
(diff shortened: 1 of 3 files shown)


soarqin:
This patch works fine. It was my mistake that not to handle empty input.

mattock:
These fixes look very reasonable. I'll ask if someone on #openvpn-devel could do a proper code review and if that fails, I think we should merge these with lazy-ACK.

ffes:
As requested by @mattock I took a look at the changes. Disclaimer in advance. I have little to no experience with the code of openvpn-gui. Only just started to contribute, just translations so far.

If I understand correctly, @soarqin already gave his approval on the base64 code. I haven't got a clue how to test the other code, but it looks good to me.

I am not a fan of the `== FALSE` check, but that is really personally and arbitrarily. On the other hand, in `misc.c` you use `if (!Base64Encode())`.

selvanair:
Thanks for the comments. Sometimes == FALSE is easier to read, but not in this case.  Will change it.

mattock:
@ffes: Thanks for the review! We're quite low on manpower on the Windows front, so your help is most welcome!

@selvanair: so now we have code review from @ffes, and @soarqin tested that the [Handle empty strings in Base64Encode](https://github.com/OpenVPN/openvpn-gui/pull/33/commits/875d9f94578d0c299a04efe317e4d0e89f59c2de) patch does not break anything. Enough testing and review for a merge in your opinion?

selvanair:
@samuli This is pretty harmless and we have a review + test. OK to merge, I think.

ffes:
@mattock I understand your request, but I can't promise much.

Some background. I am a Windows developer and sysadm. I work at a company where we use openvpn on our pfSense firewall. We have ~20 Windows clients. My main interest is the user experience for me and my users, thus openvpn-gui. I am not the "security" guy.

mattock:
@selvanair: Ok, I'll merge this.
@ffes: Ok - we have security guys on the OpenVPN side, so that part is covered pretty well already.

PR #34, opened 2016-04-07
Add all missing texts and translate them.

ffes:
As mentioned in #30 some text where still missing. This adds
all those texts and adds the texts that were added by #26 as well.

PR #36, opened 2016-04-09
Fix typos and reword on some texts

ffes:
As @syzzer correctly noticed in #34 I included two typos in that PR.

And I also did some minor rewording in the sentences.

@syzzer, could you please check my work, so we don't need yet another PR?


syzzer:
Looks good, ACK.

Thanks for following up!

selvanair:
Yay! We have an ACK. Thanks. Merging.

PR #37, opened 2016-04-11
Fill the username field with the current Windows username

ffes:
And after that put the focus on the password field.

### DIFF
--- a/openvpn.c
+++ b/openvpn.c
@@ -236,6 +236,15 @@ UserAuthDialogFunc(HWND hwndDlg, UINT msg, WPARAM wParam, LPARAM lParam)
             if(param->challenge_echo)
                 SendMessage(wnd_challenge, EM_SETPASSWORDCHAR, 0, 0);
         }
+        /* Fill the username field with the current Windows username and put the focus on the password field */
+        WCHAR username[MAX_PATH];
+        DWORD usernamesize = MAX_PATH;
+        if (GetUserName(username, &usernamesize))
+        {
+            SetDlgItemText(hwndDlg, ID_EDT_AUTH_USER, username);
+            SendMessage(hwndDlg, WM_NEXTDLGCTL, (WPARAM) GetDlgItem(hwndDlg, ID_EDT_AUTH_PASS), TRUE);
+        }
+
         if (param->c->state == resuming)
             ForceForegroundWindow(hwndDlg);
         else


selvanair:
Thanks for the patch. I too would like to see the username filled in, but remembering the previously used username would be much more useful.

At least we have several users with username not the same as the windows username and in such cases this could be more of an irritation than having a blank box.

ffes:
I understand your point.

But remember the last one used, it will be "almost always" wrong when you switch a lot between sessions.

Then it would be even better if it remembers the username per connection. But I don't think there are any per connection settings yet. How do we store that?

Or maybe just fill the default username, leave the focus on the username field and make sure the text is selected. So if it is alright, you can press `Tab` otherwise just type the right username.

ffes:
Or maybe introduce an option to control the autofill feature?

0 = disable autofill
1 = always fill with Windows username (current implementation of this PR)
2 = initial fill with Windows username but remember last used name

leobasilio:
I think a yes/no "remember last username" option would be enough. If the username is the windows username, the user would only need to type it once. There's no need for a third possibility. It could be stored in the registry, like the other settings.

selvanair:
What I have in mind is to remember username per connection. The windows username could be made the default if no previous value exists. Once its connection specific, saving it unconditionally would be ok, I think.

We locally use a patch that saves username and optionally password -- it may be a bit too much of code  with encrypted storage of password and not of broad interest. However, saving only the username in plain text is much easier. The data is saved under the OpenVPN-GUI registry key using config-specific keys. I just pushed a cherry-picked and slightly edited version to  https://github.com/selvanair/openvpn-gui (branch named save-pass). If useful feel free to extract code from it.

ffes:
@selvanair I like that idea. I will take a look at the branch later this week.

I can imagine people don't like storing the username for security reasons. So maybe we need to introduce an option to control this behavior.

Having said that, every (UI-based) OS that I am aware of stores the username for you. Putty stores the username per session in plain text in the registry as well, although that is a text you have to enter yourself in the options an is not automatically stored for you.

Don't know what others have to say about storing the username in plain text in the registry. You are a member of the team, so you probably know who is the expert on this.

selvanair:
I do not see any security issues with storing username in registry. Windows usernames are fully exposed on most desktops and may be listed using the net command anyway. If we were storing it in the network that would be a concern but not on the local machine.

@mattock do we have to make this optional? 

The code in my save-pass branch has an encrypted store implemented using CryptProtectData which hides it from other users. But I really think cleartext is fine for username. Anyway, that branch is a quickly put-together patch, not an exact copy of what I use, so some testing may be needed. In real use we also have a button to enable/disable saving of password but that's overkill if its only username.

cron2:
Hi,

On Tue, Apr 12, 2016 at 01:52:53PM -0700, Frank Fesevur wrote:
> Don't know what others have to say about storing the username in plain text in the registry. You are a member of the team, so you probably know who is the expert on this.

We have no formal "expert" on UI questions :-) - if at all, Selva has become
our resident expert due to much work on it in the last months.

I do not think storing the username is a security risk - the password, yes,
but the username is much less sensitive than having stuff like your config
file with key material lying around on your disk...

We could, of course, have a checkbox "[x] remember username"...  but personally
I do not think that the extra code is worth the benefit.

So let's just go for it, and if the feedback on 2.4_alpha with these changes
is "oh my god, we cannot accept that!", we can always add the checkbox.

My $0.2 - I'm not an authority on gui, so just an opinion.

gert

-- 
                                                           //www.muc.de/~gert/

mattock:
I agree with cron2. There is no reason to make this feature optional, and having connection-specific stored usernames sounds quite useful.

ffes:
I have looked at @selvanair branch and think it is relative easily to store the username per session. I will update my PR when I have that done.

fkooman:
I just created an other issue that is very similar to what this PR tries to solve:

https://github.com/OpenVPN/openvpn-gui/issues/39

maybe it has some additional ideas that are helpful :)

mattock:
@ffes : any progress on updating the PR?

selvanair:
With PR 52 merged-in we can now close this.

ffes:
Soon after I opened this one #52 was opened. A much better PR, so I kind of forgot about this PR. Since #52 is already merged in this one is outdated.

PR #38, opened 2016-04-19
Changed typo in include guard.

EagleErwin:


selvanair:
Thanks for the patch. A typo indeed. ACK.

PR #41, opened 2016-04-24
added cppcheck, openssl-1.0.1r

chipitsine:


selvanair:
This looks good, but I am not sure of how the irc notification works: would it not be nuisance with travis builds on forks sending notification to the irc channel? Or is there an option to make forks not send such notifications? If not, sticking with the default email notifications may be better.

chipitsine:
ok, can you commit everything except irc notifications ?

selvanair:
Looks good to me. ACK.

mattock:
@selvanair : I think Travis CI (build failure) notifications on #openvpn-windows could prove useful. Nothing tends to happen on that channel anyways. That said, I'll ask about this on the channel first.

selvanair:
@mattock I was only concerned with notifications from forks that have travis build enabled (mine is one). If that extra noise is tolerable, let's go for it.

PR #42, opened 2016-05-01
reworked packages install to "apt" plugin

chipitsine:


mattock:
Looks good afaics, merging.

PR #43, opened 2016-05-04
Add build description

ffes:
At the moment only building on Cygwin is described.

Other ways to build the project, like [MSYS2](https://msys2.github.io/) or cross-compiling on Linux (maybe even using the very much in beta [Bash on Ubuntu on Windows 10](https://msdn.microsoft.com/en-us/commandline/wsl/about)), can be added in a later stage.

About the text... know that English is not my native tongue.

### DIFF
--- a/BUILD.rst
+++ b/BUILD.rst
@@ -1,0 +1,49 @@
+How to build using Cygwin
+=========================
+
+Cygwin provides ports of many GNU/Linux tools and a POSIX API layer. This is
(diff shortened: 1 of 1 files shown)


mattock:
Looks nice, merging.

Modernizing the README has been on my TODO for a while. I'll do that as a snack hopefully within a week. After that moving the MinGW/MSYS build instructions from README to BUILD.rst makes sense I think.

PR #44, opened 2016-05-04
File 'compile' now removed as well

ffes:
When using `make maintainer-clean` the file `compile` was not removed.


mattock:
Make sense, merging.

PR #45, opened 2016-05-12
fix the following "warning"

chipitsine:
x86_64-w64-mingw32-windres: warning: 5: 230: 1033: duplicate value


selvanair:
Thanks, this fixes the warning. Merging.

PR #46, opened 2016-05-22
Launch on startup setting

leobasilio:
Hi. I've been missing this setting and I think it could be an useful feature for the end user.

### DIFF
--- a/localization.c
+++ b/localization.c
@@ -354,9 +354,54 @@ FillLangListProc(UNUSED HANDLE module, UNUSED PTSTR type, UNUSED PTSTR stringId,
     return TRUE;
 }
 
+static BOOL 
+GetLaunchOnStartup()
+{
+	
+    WCHAR regPath[MAX_PATH], exePath[MAX_PATH];	
+    BOOL result = FALSE;
+    HKEY regkey;
+
+    if (RegOpenKeyExW(HKEY_CURRENT_USER, L"Software\\Microsoft\\Windows\\CurrentVersion\\Run", 0, KEY_READ, &regkey) == ERROR_SUCCESS) {
+
+        if (GetRegistryValue(regkey, L"OpenVPN-GUI", regPath, MAX_PATH) &&
+            GetModuleFileNameW(NULL, exePath, MAX_PATH)) {
+            if (_wcsicmp(regPath, exePath) == 0)
+                result = TRUE;
+        }
+
+        RegCloseKey(regkey);
+
+    }
+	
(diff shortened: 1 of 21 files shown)


selvanair:
Isn't this better delegated to the installer to setup "run on startup"? Actually I was under the impression this to be the default. If not, the nsis script could be patched to do so. Windows 8+ has made it even easier to manage startup programs by users.

That said this patch makes it easier for users to add/remove the GUI to startup.

leobasilio:
I'm not sure the installer does it. Anyway, the user should be able to change his mind later.

localization.c
```diff
@@ -354,9 +354,54 @@ FillLangListProc(UNUSED HANDLE module, UNUSED PTSTR type, UNUSED PTSTR stringId,
     return TRUE;
 }
 
+BOOL 
+GetLaunchOnStartup()
+{
```
selvanair: This and the next function may be defined static.

localization.c
```diff
@@ -354,9 +354,54 @@ FillLangListProc(UNUSED HANDLE module, UNUSED PTSTR type, UNUSED PTSTR stringId,
...
+                SetRegistryValue(regkey, _T("OpenVPN-GUI"), exePath);
+        }
+        else 
+            RegDeleteValue(regkey, _T("OpenVPN-GUI"));            
+
+        RegCloseKey(regkey);
+
+    }
+
+}
```
selvanair: We no longer support non-unicode compilation so avoid TCHAR and _t functions. In any case prefer explicit character types in all new code as far as possible. That is, WCHAR or char (as appropriate) instead of TCHAR, explicitly use functions ending W for wide chars, L instead of T() etc. Also TCHAR in exisitng code may be assumed to mean WCHAR.

leobasilio:
Done.

mattock:
The installer does not seem to do what this patch does, see [openvpn.nsi](https://github.com/OpenVPN/openvpn-build/blob/master/windows-nsis/openvpn.nsi). I'm fine with this patch if @selvanair is fine with the code.

selvanair:
The code looks good to me. One question: the path added to the registry is not quoted. Does it work when the path has spaces (as in C:\Program Files\...)? All startup entries on my machine are quoted.

selvanair:
Works without quotes at least on Windows 7. ACK from me.

leobasilio:
I have some entries without quotes and they work fine too. I think quotes are useful only when you have command line arguments.

PR #47, opened 2016-05-22
Rescan configs

selvanair:
The first patch is a minor bug fix -- add a missing call to CheckReadAccess() and also simplify its usage.

The second patch adds a feature that helps the "import file" option:
Currently if a config is imported when connections are active it does not appear in the config list until all connections terminate or GUI restarts. The patch makes it possible to rescan for configs even if connections are active.

### DIFF
--- a/openvpn_config.c
+++ b/openvpn_config.c
@@ -217,7 +217,13 @@ BuildFileList()
 {
     static bool issue_warnings = true;
 
-    o.num_configs = 0;
+    /*
+     * If no connections are active reset num_configs and rescan
+     * to make a new list. Else we keep all current configs and
+     * rescan to add any new one's found
+     */
+    if (CountConnState(disconnected) == o.num_configs)
+        o.num_configs = 0;
 
     BuildFileList0 (o.config_dir, issue_warnings);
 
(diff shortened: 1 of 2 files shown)


selvanair:
@leobasilio The second commit is supposed to help the "import file" feature. Could you please test it? Thanks.

leobasilio:
After applying the patch, I've imported a config file while I had an active connection and it did show up on the menu.

mattock:
@selvanair : these fixes make sense. One question, though: why was the connection list updated only when no connections where active?

selvanair:
@mattock :  In the original code the connections array gets recreated from scratch each time the user clicks on the tray icon. This is nice as new configs are immediately recognized. But the connection array includes pointers to sockets, window handles etc so such a refresh was obviously possible only when no connections are active, else some valid pointers will be lost. 

The patch changes the rescan logic by keeping all existing configs and adding only any new ones during a rescan. It reverts to the original behaviour if no connections are active.

A better method is to use a linked list instead of an array, but I leave that for a time if/when some feature demands such a functionality.

mattock:
@selvanair : Ok, let's merge this then.

PR #48, opened 2016-05-22
Fix exit handling while in modal loops

selvanair:
**How to test:**
Using the current GUI and interactive service, start a connection
that needs user-auth. While that dialog is open press the exit button 
from the tray menu. The GUI will exit after waiting for a few seconds but 
the openvpn.exe process will be left behind (will have to be killed manually 
before a new connection can be made). This patch should fix it.

**Commit message:**
PostThreadMessage used to trigger exit event gets lost while
in modal loops such as auth dialog. Replace it by PostMessage
and handle it in the status window callback.

Fixes openvpn processes left behind if exit is pressed while
user-auth dialog is active.

### DIFF
--- a/openvpn.c
+++ b/openvpn.c
@@ -848,6 +848,27 @@ StatusDialogFunc(HWND hwndDlg, UINT msg, WPARAM wParam, LPARAM lParam)
     case WM_DESTROY:
         PostQuitMessage(0);
         break;
+
+    case WM_OVPN_STOP:
+        c = (connection_t *) GetProp(hwndDlg, cfgProp);
+        c->state = disconnecting;
+        RunDisconnectScript(c, false);
+        EnableWindow(GetDlgItem(c->hwndStatus, ID_DISCONNECT), FALSE);
+        EnableWindow(GetDlgItem(c->hwndStatus, ID_RESTART), FALSE);
+        SetMenuStatus(c, disconnecting);
+        SetDlgItemText(c->hwndStatus, ID_TXT_STATUS, LoadLocalizedString(IDS_NFO_STATE_WAIT_TERM));
+        SetEvent(c->exit_event);
+        break;
+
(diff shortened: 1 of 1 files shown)


leobasilio:
The issue is real, and the patch does work as intended. There's just this compilation warning:

>../openvpn-gui/openvpn.c:1165:17: warning: passing argument 1 of 'PostMessageW' makes pointer from integer without a cast [-Wint-conversion]
>     PostMessage(o.conn[config].threadId, WM_OVPN_SUSPEND, 0, 0);

selvanair:
On Sun, May 22, 2016 at 9:23 PM, Leonardo Basilio <notifications@github.com>
wrote:

> The issue is real, and the patch does work as intended. There's just this
> compilation warning:
>
> ../openvpn-gui/openvpn.c:1165:17: warning: passing argument 1 of
> 'PostMessageW' makes pointer from integer without a cast [-Wint-conversion]
> PostMessage(o.conn[config].threadId, WM_OVPN_SUSPEND, 0, 0);
>
>
Thanks. That is a bug -- forgot to change the threadId to window handle in
SuspendOpenVPN. It may not show up in normal use as suspend is probably
never used anymore. Will fix and push a new version.

selvanair:
Pushed a fixup.

mattock:
@leobasilio : compile problem gone?

leobasilio:
Yes. It's gone.

chipitsine:
we can add -Werror to gcc/Makefile

selvanair:
@leobasilio has confirmed the bug and tested the patch. What about the code itself -- is it good to be merged?

selvanair:
@mattock : this bug fix has been tested by leobasilio, so may be merge with a lazy ACK?

mattock:
@selvanair: :+1: merging.

PR #50, opened 2016-05-23
cache openssl download

chipitsine:
```
(*) cache openssl download when compiling in Travis CI
(*) stop compilation on every warning
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,8 @@ compiler: gcc
 env:
   global:
     - OPENSSL_VERSION=1.0.1r
+    - OPENSSL_CRYPTO_CFLAGS="-I${HOME}/image/include"
+    - OPENSSL_CRYPTO_LIBS="-L${HOME}/image/lib -lcrypto"
 
 matrix:
   fast_finish: true
@@ -39,8 +41,6 @@ script:
     - ./Configure --prefix=/ --cross-compile-prefix=${CHOST}- shared $m no-multilib no-capieng --openssldir=/etc/ssl --libdir=/lib -static-libgcc >build.log 2>&1 || (cat build.log && exit 1)
     - make install INSTALL_PREFIX="${HOME}/image" INSTALLTOP="/" MANDIR="/tmp" >build.log 2>&1 || (cat build.log && exit 1)
     - cd .. && autoreconf -i -v
-    - export OPENSSL_CRYPTO_CFLAGS="-I${HOME}/image/include"
-    - export OPENSSL_CRYPTO_LIBS="-L${HOME}/image/lib -lcrypto"
     - ./configure --prefix=/ --libdir=/lib --host=$CHOST --build=x86_64-pc-linux-gnu --program-prefix='' && make
     
 cache:
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,10 @@ os: linux
 language: cpp
 compiler: gcc
 
+env:
+  global:
+    - OPENSSL_VERSION=1.0.1r
+
 matrix:
   fast_finish: true
   include:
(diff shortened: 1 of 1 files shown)


Makefile.am
```diff
@@ -39,7 +39,7 @@ dist_doc_DATA = \
 	COPYING
 
 AM_CPPFLAGS = $(OPENSSL_CRYPTO_CFLAGS) -D_UNICODE
-AM_CFLAGS = -municode
+AM_CFLAGS = -municode -Werror
```
selvanair: We have -Wall and -Wextra turned on, adding -Werror to that could become very limiting. The code wont compile until all warnings are eliminated is not where I would like to go... Let's see what others say.

cron2: Hi,

On Mon, May 23, 2016 at 01:45:58PM -0700, Selva Nair wrote:
> > @@ -39,7 +39,7 @@ dist_doc_DATA = \
> >  	COPYING
> >  
> >  AM_CPPFLAGS = $(OPENSSL_CRYPTO_CFLAGS) -D_UNICODE
> > -AM_CFLAGS = -municode
> > +AM_CFLAGS = -municode -Werror
> 
> We have -Wall and -Wextra turned on, adding -Werror to that could become very limiting. The code wont compile until all warnings are eliminated is not where I would like to go... Let's see what others say.

Looking at warnings and getting rid of the reasonable ones is good :-)

Elimitinating *all* warnings might just not be possible with mingw, if the
header files itself trigger warnings (which I've seen compiling on 
ubuntu 14), so -Werror on-by-default for cross-compiled code is problematic.
So I'd not go there.

gert
-- 
                                                           //www.muc.de/~gert/

chipitsine:
it is being built in Travis-CI, there's Ubuntu 14, I do not see any warning in headers. can you please provide details on that ?

selvanair:
Even if the current code can be easily made to pass -Werror doesn't mean that it won't limit us in future. A little more self-discipline to scan through any warnings before submitting a PR  is what I would aim for -- that remark is aimed at myself :)

leobasilio:
I agree with @selvanair.

cron2:
hi,

On Tue, May 24, 2016 at 07:13:38AM -0700, Leonardo Basilio wrote:
> I agree with @selvanair. 

+1

gert
-- 
                                                           //www.muc.de/~gert/

selvanair:
So that's it then. @chipitsine : could you please remove the second patch and force-push? A more meaningful commit summary (headline) would be useful too. Thanks.

selvanair:
Thanks for the edits.
While this may be a good start to implement caching, downloading takes hardly any time or resource these days.  it would be much more useful if the build results of openssl (libs and includes) could be cached as well.  That is likely where bulk of the re-build time is spent. Would it be too hard to cache those results?

Otherwise I leave it to @mattock to decide whether these changes look good.

chipitsine:
I'm playing with ccache in my private branch. No luck yet, ccache is tricky in case of mingw.

cron2:
Hi,

On Thu, May 26, 2016 at 01:00:08PM -0700, Selva Nair wrote:
> Thanks for the edits.
> While this may be a good start to implement caching, downloading takes hardly any time or resource these days.  it would be much more useful if the build results of openssl (libs and includes) could be cached as well.  That is likely where bulk of the re-build time is spent. Would it be too hard to cache those results?

The "normal" openvpn-build (not the gui side) already can do caching
for dependencies.  It's a bit complicated (as far as I remember you have
to tell it "now build and cache dependencies!" and later on "use the
precompiled dependencies!") but does work.

Need to look up how this worked and comment here...  early next week.

gert
-- 
                                                           //www.muc.de/~gert/

chipitsine:
I had a look at that. it seem to be complicated on openssl version upgrade.
however, I keep it in mind.

if I won't be lucky with ccache, I'll use that

chipitsine:
actually, there's very little need of using openssl in openvpn-gui. it is used only for certificate manipulation, i.e. to encrypt or decrypt it.

it could be done with wincrypt api, I think.

selvanair:
Yes, openssl is used only for changing the private key password (a rarely used feature?), replacing it with windows crypto api may be a good idea. That would also eliminate all those queries we often get about cross-compiling openssl.

chipitsine:
I will adress that in separate PR

mattock:
Code-vise this PR looks reasonable, but I would rather focus on using a pre-built OpenSSL version, as that would save a lot more time. We can host the pre-built OpenSSL tar.gz on a webserver if needed.

chipitsine:
compiling openssl takes less than 2 minutes (which is very small time comparing to travis-ci time limit, which is 50 minutes).

also, travis-ci cloud is known to "network name not resolved" or "connection refused" errors pretty often, so I would not like to depend on downloading openssl cache from build.openvpn.net

this patch prevent us from network outages (downloading openssl from openssl.org)

mattock:
Ah, I see. In case of a problematic network this PR could be useful. When building Windows installers for OpenVPN the "compile OpenSSL" step is by far the heaviest. Then again, the build computer I use is pretty low-spec.

chipitsine:
can someone merge it?

mattock:
@cron2, @selvanair : any further comments on this? If not, I'll merge it.

selvanair:
@mattock : no further comments from me..

PR #52, opened 2016-05-30
Save username and optionally passwords

selvanair:
- Username and, optionally, password as well as the private
  key passphrase are saved in config-specific registry keys
- All saved data are kept encrypted using DPAPI
- The passphrase dialog is skipped if a valid saved private
  key password is available. However, the user-auth dialog
  is always presented, prefilled with the saved username
  and password.

Note: A text string "Save password" is added to three dialogs
in all language resource files. Additional text with ids
IDS_MENU_CLEARPASS and IDS_NFO_DELETE_PASS are added to the
STRINGTABLE only in the English language resource file.
All these need translations.


chipitsine:
we tested, works good

selvanair:
Thanks for testing. Can anyone provide a code-review please? A few more eyes looking at least into how the added calls to save_xx and  recall_xx affect the rest of the code would help make sure we are not breaking anything.

chipitsine:
can you, please, merge it?

mattock:
@selvanair: I'll ask for review help on #openvpn-devel again. If that leads to nothing in a few days, I'd say let's just merge this with a "lazy-ACK".

d12fk:
Feature ACK for using Crypt(Un)ProtectData.

mattock:
@d12fk: how about a code-ACK? You're probably the most correct person to review the actual code.

mattock:
@selvanair : I think it's lazy-ACK time again. Feel free to do it :)

selvanair:
Until we have more hands to do code review, it looks like we have to make do with feature ACKs and test reports. Merging.

PR #60, opened 2016-06-22
add --disable-password-change option to configure

chipitsine:


.travis.yml
```diff
@@ -15,13 +15,13 @@ matrix:
   fast_finish: true
   include:
     - compiler: ": Win32"
-      env: CHOST=i686-w64-mingw32 m=mingw
+      env: CHOST=i686-w64-mingw32
```
mattock: What did the "m=mingw" do exactly?

mattock:
Cutting away the last word ("configurations") off of the commit message of https://github.com/OpenVPN/openvpn-gui/pull/60/commits/5c5677a790453400e5174ef81340f2b8de5da097 makes it fit nicely into 72 characters.

.travis.yml
chipitsine: it specifies either mingw or mingw64 for openssl ./Configure

.travis.yml
```diff
@@ -38,6 +42,8 @@ script:
     - cppcheck --force --enable=warning,missingInclude --error-exitcode=1 . >build.log 2>&1 || (cat build.log && exit 1)
     - mkdir image
     - tar zxf download-cache/openssl-${OPENSSL_VERSION}.tar.gz && cd openssl-${OPENSSL_VERSION}
+    - if [ "$CHOST" == "i686-w64-mingw32" ]; then export m=mingw; fi
```
selvanair: As $m is later used as the target spec for openssl compilation, "target=mingw" etc instead of "m=mingw" would be easier to understand for others looking a this. Just nitpicking :)

selvanair:
How does OPENSSL_CRYPTO_CFLAGS and _LIBS get unset for the build with DISABLE_CHANGE_PASSWORD ? If they remain we can't be sure that that the build can succeed without needing those defined, isn't it?

chipitsine:
currently build wil fail without openssl. I'm going to add configure option later, which will skip openssl check during configure and openssl will not be compiled for DISABLE_CHANGE_PASSWORD

chipitsine:
is it good now ?

configure.ac
```diff
@@ -38,6 +38,13 @@ AC_ARG_ENABLE(
 	[enable_distonly="no"]
 )
 
+AC_ARG_ENABLE(
+	[password-save],
+	[AS_HELP_STRING([--disable-password-save], [disable password save support @<:@default=yes@:>@])],
```
selvanair: I suppose you meant to write  "password-change", "password change support" etc.. and not "password save"

chipitsine: oops

chipitsine:
it a typo, I meant "change password" define, not "save password" option

chipitsine:
is it ok now?

configure.ac
```diff
@@ -69,6 +76,7 @@ if test "x$MAKENSIS" = "x"; then
 fi
 AM_CONDITIONAL([HAVE_NSIS], [test "x$MAKENSIS" != "x"])
 
+if test "${enable_password_save}" = "yes"; then
```
selvanair: PKG_PROG_PKG_CONFIG may be also moved inside this "if then"  so that with --disable-password-change users can compile even if pkg-config is not available.

Makefile.am
```diff
@@ -38,7 +38,7 @@ dist_doc_DATA = \
 	COPYRIGHT.GPL \
 	COPYING
 
-AM_CPPFLAGS = $(OPENSSL_CRYPTO_CFLAGS) -D_UNICODE
+AM_CPPFLAGS = $(CRYPTO_CFLAGS) -D_UNICODE
```
selvanair: Why this change? How would this work without also changing the prefix OPENSSL_CRYPTO  to CRYPTO  in PKG_CHECK_MODULES test? Also changing the name will break some existing build scripts, especially when cross-compiling.

configure.ac
```diff
@@ -38,6 +38,13 @@ AC_ARG_ENABLE(
 	[enable_distonly="no"]
 )
 
+AC_ARG_ENABLE(
+	[password-save],
+	[AS_HELP_STRING([--disable-password-save], [disable password save support @<:@default=yes@:>@])],
+	[enable_password_save="yes"],
+	[enable_password_save="no"]
+)
```
selvanair: This does not look right. The default is yes, so the last arg to AC_ARG_ENABLE would be enable_password_xxx ="yes".  And the third arg may be left blank.

Makefile.am
chipitsine: I reverted that part. Do not pay attention to that.

I realised that it will break openvpn/openvpn-build

configure.ac
chipitsine: that makes sense. ok, thank you

configure.ac
chipitsine: I tried it with "yes" in the last arg and it does not work as expected.
on the other side, https://travis-ci.org/OpenVPN/openvpn-gui/builds/148929302 works exactly as expected


I agree that we can omit default value. however, it is not clear about yes/no choice

chipitsine: can you fork it in your private repo and try the way you suggest?

selvanair: On Mon, Aug 1, 2016 at 1:20 PM, Ilya Shipitsin <notifications@github.com>
wrote:

> > @@ -38,6 +38,13 @@ AC_ARG_ENABLE(
> >  	[enable_distonly="no"]
> >  )
> >
> > +AC_ARG_ENABLE(
> > +	[password-save],
> > +	[AS_HELP_STRING([--disable-password-save], [disable password save support @<:@default=yes@:>@])],
> > +	[enable_password_save="yes"],
> > +	[enable_password_save="no"]
> > +)
>
> I tried it with "yes" in the last arg and it does not work as expected.
> on the other side,
> https://travis-ci.org/OpenVPN/openvpn-gui/builds/148929302 works exactly
> as expected
>
> I agree that we can omit default value. however, it is not clear about
> yes/no choice
>
No, the third arg is not the default, the 4th is the default. Set the 4th
to our default (yes) and third as empty, and it should work.

In AC_ARG_ENABLE the last arg defines the default value to be used if
neither --enable-xxx or --disable-xxx is passed to ./configure. So that
should be "yes" in our case as --enable-password-change is the default.

The third arg is used when the option is specified in either--enable or
--disable form -- in this case enable_password_change will already have the
correct value so that arg should be empty unless some other checks are
required.

Selva

chipitsine: ok, what's wrong ?


commit: https://github.com/chipitsine/openvpn-gui/commit/49f8f4edfe0f2096684182854368819eaf33d0d3
build:  https://travis-ci.org/chipitsine/openvpn-gui/builds/148951122

chipitsine: @selvanair, can you produce successful build ?

selvanair: On Mon, Aug 1, 2016 at 2:19 PM, Ilya Shipitsin <notifications@github.com>
wrote:

> In configure.ac
> <https://github.com/OpenVPN/openvpn-gui/pull/60#discussion_r73026272>:
>
> > @@ -38,6 +38,13 @@ AC_ARG_ENABLE(
> >  	[enable_distonly="no"]
> >  )
> >
> > +AC_ARG_ENABLE(
> > +	[password-save],
> > +	[AS_HELP_STRING([--disable-password-save], [disable password save support @<:@default=yes@:>@])],
> > +	[enable_password_save="yes"],
> > +	[enable_password_save="no"]
> > +)
>
> ok, what's wrong ?
>
Looking at your failed travis build log I see that configure is called as

./configure --prefix=/ --libdir=/lib --host=$CHOST
--build=x86_64-pc-linux-gnu --program-prefix='' && make

No --disable-password-change passed? I'm not familiar with travis, but I
guess $CONFIG  (=--disable-password-change, in the last two builds) has to
somehow get into the command line?

Selva


This will set enable_password_save to no if no options are specified and
and to yes if either disable-password-save or enable-password-save is
specified. We want none of that. So write it as


AC_ARG_ENABLE([password-save],[AS_HELP_STRING([--disable-password-save],
              [disable password save support @<:@default=yes@:>@])],
              [],
              [enable_password_save="no"]
)



commit: chipitsine@49f8f4e

selvanair: Please ignore the last two para of last reply -- some crux from cut-n-paste.

chipitsine: sure, I forgot to pass $CONFIG to configure, thank you!

chipitsine:
is it ok now ?

selvanair:
Looks good and cross-compile works as expected, but I have no setup to test with pkg-config on Windows. @mattock : do the travis-ci changes look okay?

chipitsine:
I tested on mingw as well

chipitsine:
up

mattock:
Afaics the Travis-CI changes are ok, but figuring out what is going on there is not trivial due to OpenSSL build being moved to a separate file.. I would be inclined to merging this as-is, and _if_ there are issues, having them fixed later.

chipitsine:
sounds good

chipitsine:
up

PR #69, opened 2016-08-23
fix build --disable-password-change

chipitsine:


selvanair:
Yes, that was my fault. Thanks for the fix.
ACK.

PR #70, opened 2016-08-23
more russian translation

chipitsine:


mattock:
@ValdikSS ACKed this on IRC, merging. Thanks!

PR #72, opened 2016-08-24
No new line for "Starting this connection requires membership in"

ValdikSS:
Because this happens:

screenshot_20160824_194707


selvanair:
Looks good. This change may be applied to other language files too (at least Dutch has this already translated). And in PR 71 for Russian.

mattock:
I take "looks good" from Selva as an ACK. Merging.

PR #74, opened 2016-08-27
removed unused nsis installer

chipitsine:


mattock:
Thanks! The changes make sense, and based on build-complete tests don't seem to break anything. Merging.

Issue #73, opened 2016-08-25
typo in openvpn-gui-res-it.rc


alessiomangoni13:
in line 262:
IDS_NFO_ASSIGN_IP "IP asseganto: %s"

should be
IDS_NFO_ASSIGN_IP "IP **assegnato**: %s"

mattock:
Looks good, merging. Grazie!

PR #75, opened 2016-08-28
Fix Italian typo

ValdikSS:
Fix typo from #73


mattock:
This is clearly a typo, so an ACK. Merging.

PR #76, opened 2016-08-28
Functionality to import configuration file from the command line

ValdikSS:
This functionality allows to import configuration file using `--import` command line option. It is intended to be used as a handler when *.ovpn files are double-clicked. NSIS installer script should be modified to achieve this (https://github.com/OpenVPN/openvpn-build/pull/26).

### DIFF
--- a/main.c
+++ b/main.c
@@ -173,9 +173,10 @@ int WINAPI _tWinMain (HINSTANCE hThisInstance,
 
   GetRegistryKeys();
 
-  EnsureDirExists(o.config_dir);
   /* Parse command-line options */
   ProcessCommandLine(&o, GetCommandLine());
+
+  EnsureDirExists(o.config_dir);
 
   if (already_running)
     {
@@ -540,11 +541,13 @@ ShowConfigFileDialog(PTCHAR config_file)
     int dlgresult = 0;
 
     fileName = PathFindFileName(config_file);
-    dlgresult = ShowLocalizedMsgEx(MB_YESNO, _T(PACKAGE_NAME), IDS_ASK_IMPORT_OR_EDIT, fileName);
+    dlgresult = ShowLocalizedMsgEx(MB_YESNOCANCEL, _T(PACKAGE_NAME), IDS_ASK_IMPORT_OR_EDIT, fileName);
     if (dlgresult == IDYES)
         ImportConfigFile(config_file);
-    else
+    else if(dlgresult == IDNO)
         EditConfig(0, config_file);
+    else
+        return;
 }
 
 
@@ -553,7 +556,7 @@ ImportConfigFile(PTCHAR config_file)
 {
     TCHAR source[MAX_PATH] = _T("");
     PTCHAR fileName;
-    BOOL result = 1;
+    BOOL result = FALSE;
 
     if (config_file == NULL)
     {
@@ -577,10 +580,17 @@ ImportConfigFile(PTCHAR config_file)
 
         result = GetOpenFileName(&fn) && (fileName = source + fn.nFileOffset);
     }
-    else
+    else if(PathFileExists(config_file))
     {
         fileName = PathFindFileName(config_file);
-        _tcscpy(source, config_file);
+        _sntprintf_0(source, L"%s", config_file);
+        result = TRUE;
+    }
+    else
+    {
+        ShowLocalizedMsg(IDS_ERR_PATH_NOT_FOUND, config_file);
+        PrintDebug(L"The file specified for import does not exist: '%s'", config_file);
+        return;
     }
 
     if (result)
@@ -589,9 +599,9 @@ ImportConfigFile(PTCHAR config_file)
 
         _sntprintf_0(destination, _T("%s\\%s"), o.config_dir, fileName);
 
-        destination[_tcslen(destination) - _tcslen(o.ext_string) - 1] = _T('\0');
+        PathRemoveExtension(destination);
         
-        if (EnsureDirExists(destination) && EnsureDirExists(o.config_dir))
+        if (EnsureDirExists(destination))
         {
 
             _sntprintf_0(destination, _T("%s\\%s"), destination, fileName);
@@ -612,11 +622,8 @@ ImportConfigFile(PTCHAR config_file)
             }
 
         }
-
         ShowLocalizedMsg(IDS_ERR_IMPORT_FAILED, destination);
-
     }
-    
 }
 
 #ifdef DEBUG
--- a/main.c
+++ b/main.c
@@ -105,6 +105,7 @@ int WINAPI _tWinMain (HINSTANCE hThisInstance,
   MSG messages;            /* Here messages to the application are saved */
   WNDCLASSEX wincl;        /* Data structure for the windowclass */
   DWORD shell32_version;
+  BOOL already_running = 0;
 
   /* Initialize handlers for manangement interface notifications */
   mgmt_rtmsg_handler handler[] = {
@@ -159,22 +160,28 @@ int WINAPI _tWinMain (HINSTANCE hThisInstance,
   PrintDebug(_T("Shell32.dll version: 0x%lx"), shell32_version);
 #endif
 
-  /* Parse command-line options */
-  ProcessCommandLine(&o, GetCommandLine());
-
   /* Check if a previous instance is already running. */
   if ((FindWindow (szClassName, NULL)) != NULL)
     {
         /* GUI already running */
-        ShowLocalizedMsg(IDS_ERR_GUI_ALREADY_RUNNING);
-        exit(1);
+        already_running = 1;
     }
-
-  UpdateRegistry(); /* Checks version change and update keys/values */
+  else
+    {
+        UpdateRegistry(); /* Checks version change and update keys/values */
+    }
 
   GetRegistryKeys();
 
   EnsureDirExists(o.config_dir);
+  /* Parse command-line options */
+  ProcessCommandLine(&o, GetCommandLine());
+
+  if (already_running)
+    {
+        ShowLocalizedMsg(IDS_ERR_GUI_ALREADY_RUNNING);
+        exit(1);
+    }
 
   if (!CheckVersion()) {
     exit(1);
@@ -352,7 +359,7 @@ LRESULT CALLBACK WindowProcedure (HWND hwnd, UINT message, WPARAM wParam, LPARAM
         ViewLog(LOWORD(wParam) - IDM_VIEWLOGMENU);
       }
       if ( (LOWORD(wParam) >= IDM_EDITMENU) && (LOWORD(wParam) < IDM_EDITMENU + MAX_CONFIGS) ) {
-        EditConfig(LOWORD(wParam) - IDM_EDITMENU);
+        EditConfig(LOWORD(wParam) - IDM_EDITMENU, NULL);
       }
       if ( (LOWORD(wParam) >= IDM_CLEARPASSMENU) && (LOWORD(wParam) < IDM_CLEARPASSMENU + MAX_CONFIGS) ) {
         DisablePasswordSave(&o.conn[LOWORD(wParam) - IDM_CLEARPASSMENU]);
@@ -527,11 +534,26 @@ CloseApplication(HWND hwnd)
 }
 
 void
+ShowConfigFileDialog(PTCHAR config_file)
+{
+    PTCHAR fileName;
+    int dlgresult = 0;
+
+    fileName = PathFindFileName(config_file);
+    dlgresult = ShowLocalizedMsgEx(MB_YESNO, _T(PACKAGE_NAME), IDS_ASK_IMPORT_OR_EDIT, fileName);
+    if (dlgresult == IDYES)
+        ImportConfigFile(config_file);
+    else
+        EditConfig(0, config_file);
+}
+
+
+void
 ImportConfigFile(PTCHAR config_file)
 {
     TCHAR source[MAX_PATH] = _T("");
     PTCHAR fileName;
-    BOOL dlgresult;
+    BOOL result = 1;
 
     if (config_file == NULL)
     {
@@ -553,16 +575,15 @@ ImportConfigFile(PTCHAR config_file)
         fn.Flags = OFN_DONTADDTORECENT | OFN_FILEMUSTEXIST;
         fn.lpstrDefExt = NULL;
 
-        dlgresult = GetOpenFileName(&fn) && (fileName = source + fn.nFileOffset);
+        result = GetOpenFileName(&fn) && (fileName = source + fn.nFileOffset);
     }
     else
     {
         fileName = PathFindFileName(config_file);
-        dlgresult = ShowLocalizedMsgEx(MB_YESNO, _T(PACKAGE_NAME), IDS_ASK_IMPORT_CONFIRM, fileName) == IDYES;
         _tcscpy(source, config_file);
     }
 
-    if (dlgresult)
+    if (result)
     {
         TCHAR destination[MAX_PATH];
 
@@ -570,7 +591,7 @@ ImportConfigFile(PTCHAR config_file)
 
         destination[_tcslen(destination) - _tcslen(o.ext_string) - 1] = _T('\0');
         
-        if (EnsureDirExists(destination))
+        if (EnsureDirExists(destination) && EnsureDirExists(o.config_dir))
         {
 
             _sntprintf_0(destination, _T("%s\\%s"), destination, fileName);
--- a/main.c
+++ b/main.c
@@ -59,7 +59,6 @@
 LRESULT CALLBACK WindowProcedure (HWND, UINT, WPARAM, LPARAM);
 static void ShowSettingsDialog();
 void CloseApplication(HWND hwnd);
-void ImportConfigFile();
 
 /*  Class name and window title  */
 TCHAR szClassName[ ] = _T("OpenVPN-GUI");
@@ -160,6 +159,8 @@ int WINAPI _tWinMain (HINSTANCE hThisInstance,
   PrintDebug(_T("Shell32.dll version: 0x%lx"), shell32_version);
 #endif
 
+  /* Parse command-line options */
+  ProcessCommandLine(&o, GetCommandLine());
 
   /* Check if a previous instance is already running. */
   if ((FindWindow (szClassName, NULL)) != NULL)
@@ -172,8 +173,6 @@ int WINAPI _tWinMain (HINSTANCE hThisInstance,
   UpdateRegistry(); /* Checks version change and update keys/values */
 
   GetRegistryKeys();
-  /* Parse command-line options */
-  ProcessCommandLine(&o, GetCommandLine());
 
   EnsureDirExists(o.config_dir);
 
@@ -364,7 +363,7 @@ LRESULT CALLBACK WindowProcedure (HWND hwnd, UINT message, WPARAM wParam, LPARAM
       }
 #endif
       if (LOWORD(wParam) == IDM_IMPORT) {
-        ImportConfigFile();
+        ImportConfigFile(NULL);
       }
       if (LOWORD(wParam) == IDM_SETTINGS) {
         ShowSettingsDialog();
@@ -528,34 +527,44 @@ CloseApplication(HWND hwnd)
 }
 
 void
-ImportConfigFile()
+ImportConfigFile(PTCHAR config_file)
 {
-    
-    TCHAR filter[2*_countof(o.ext_string)+5];
-
-    _sntprintf_0(filter, _T("*.%s%c*.%s%c"), o.ext_string, _T('\0'), o.ext_string, _T('\0'));
-    
-    OPENFILENAME fn;
     TCHAR source[MAX_PATH] = _T("");
+    PTCHAR fileName;
+    BOOL dlgresult;
 
-    fn.lStructSize = sizeof(OPENFILENAME);
-    fn.hwndOwner = NULL;
-    fn.lpstrFilter = filter;
-    fn.lpstrCustomFilter = NULL;
-    fn.nFilterIndex = 1;
-    fn.lpstrFile = source;
-    fn.nMaxFile = MAX_PATH;
-    fn.lpstrFileTitle = NULL;
-    fn.lpstrInitialDir = NULL;
-    fn.lpstrTitle = NULL;
-    fn.Flags = OFN_DONTADDTORECENT | OFN_FILEMUSTEXIST;
-    fn.lpstrDefExt = NULL;
-
-    if (GetOpenFileName(&fn)) 
+    if (config_file == NULL)
     {
+        OPENFILENAME fn;
+        TCHAR filter[2*_countof(o.ext_string)+5];
 
+        _sntprintf_0(filter, _T("*.%s%c*.%s%c"), o.ext_string, _T('\0'), o.ext_string, _T('\0'));
+
+        fn.lStructSize = sizeof(OPENFILENAME);
+        fn.hwndOwner = NULL;
+        fn.lpstrFilter = filter;
+        fn.lpstrCustomFilter = NULL;
+        fn.nFilterIndex = 1;
+        fn.lpstrFile = source;
+        fn.nMaxFile = MAX_PATH;
+        fn.lpstrFileTitle = NULL;
+        fn.lpstrInitialDir = NULL;
+        fn.lpstrTitle = NULL;
+        fn.Flags = OFN_DONTADDTORECENT | OFN_FILEMUSTEXIST;
+        fn.lpstrDefExt = NULL;
+
+        dlgresult = GetOpenFileName(&fn) && (fileName = source + fn.nFileOffset);
+    }
+    else
+    {
+        fileName = PathFindFileName(config_file);
+        dlgresult = ShowLocalizedMsgEx(MB_YESNO, _T(PACKAGE_NAME), IDS_ASK_IMPORT_CONFIRM, fileName) == IDYES;
+        _tcscpy(source, config_file);
+    }
+
+    if (dlgresult)
+    {
         TCHAR destination[MAX_PATH];
-        PTCHAR fileName = source + fn.nFileOffset;
 
         _sntprintf_0(destination, _T("%s\\%s"), o.config_dir, fileName);
 
@@ -570,7 +579,7 @@ ImportConfigFile()
             {
                 if (GetLastError() == ERROR_FILE_EXISTS)
                 {
-                    fileName[_tcslen(fileName) - _tcslen(o.ext_string) - 1] = _T('\0');
+                    PathRemoveExtension(fileName);
                     ShowLocalizedMsg(IDS_ERR_IMPORT_EXISTS, fileName);
                     return;
                 }
--- a/viewlog.c
+++ b/viewlog.c
@@ -27,6 +27,7 @@
 #include <stdio.h>
 #include <shellapi.h>
 #include <objbase.h>
+#include <shlwapi.h>
 
 #include "tray.h"
 #include "openvpn.h"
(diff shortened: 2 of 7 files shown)


main.c
```diff
@@ -160,6 +159,8 @@ int WINAPI _tWinMain (HINSTANCE hThisInstance,
   PrintDebug(_T("Shell32.dll version: 0x%lx"), shell32_version);
 #endif
 
+  /* Parse command-line options */
+  ProcessCommandLine(&o, GetCommandLine());
```
ValdikSS: I'm not sure if this absolutely correct to move this function. Can somebody please confirm it won't break anything?

selvanair:
On Sun, Aug 28, 2016 at 5:46 PM, ValdikSS <notifications@github.com> wrote:

> This functionality allows to import configuration file using --import
> command line option. It is intended to be used as a handler when *.ovpn
> files are double-clicked. NSIS installer script should be modified to
> achieve this.
>
We cannot change the file association without breaking the current code
where its used to open the editor. I agree its useful to have double-click
start the connection, but a patch does this should also address how to
handle editing in a way user's preferred editor is used -- that was the
reason for implementing config file editing using file association.

Selva

main.c
selvanair: On Sun, Aug 28, 2016 at 5:48 PM, ValdikSS <notifications@github.com> wrote:

> In main.c
> <https://github.com/OpenVPN/openvpn-gui/pull/76#discussion_r76539811>:
>
> > @@ -160,6 +159,8 @@ int WINAPI _tWinMain (HINSTANCE hThisInstance,
> >    PrintDebug(_T("Shell32.dll version: 0x%lx"), shell32_version);
> >  #endif
> >
> > +  /* Parse command-line options */
> > +  ProcessCommandLine(&o, GetCommandLine());
>
> I'm not sure if this absolutely correct to move this function. Can
> somebody please confirm it won't break anything?
>
The only requirement I can think of is that default options should be set
first, then the  registry scanned for user-preferences and only then
command-line options should be set so that the latter over-rides the
former, not the other way. I haven't looked at this patch to see whether
that order is preserved. If its necessary to parse some command-line
options early, some changes to the parsing code (like call it twice) may be
required.

Selva

ValdikSS:
@selvanair I don't think I understood you. Configuration file won't be automatically imported on double click, the user is asked first with a yes/no dialog. I've added an option to edit configuration file with a notepad.exe using a right click menu to an nsis script if you were talking about that.
Do you want me to make a dialog like "What would you like to do:" "Import/edit/cancel"?

mattock:
@ValdikSS : I think @selvanair is speaking of PR https://github.com/OpenVPN/openvpn-gui/pull/62, in particular commit https://github.com/OpenVPN/openvpn-gui/pull/62/commits/964d728a42aad12aed5e5bf3462718faed2b50ef. In [openvpn-build PR 26](https://github.com/OpenVPN/openvpn-build/pull/26/files) you've added a registry key that makes the default .ovpn file edit command "notepad.exe %1". That should not be required now, as file associations determine which editor to use.

selvanair:
As @mattock pointed out, I was referring to PR  [#62](https://github.com/OpenVPN/openvpn-gui/pull/62) where we implemented use of file association for editing. So the change proposed in PR [#26](https://github.com/OpenVPN/openvpn-build/pull/26/files) of openvpn-build should be
(i) do not associate any editor
(ii) Before setting "openvpn-gui.exe -import" as the open command, amend the GUI code where the verb "open" is used for editing -- otherwise the edit config menu in the GUI will spawn openvpn-gui instead of an editor. However, this will take away the user's option to set an editor of their choice. We hav eto somehow remedy that as well. 

To be clear, I'm referring to existing code in openvpn-gui which require some changes to co-operate with this PR 76.

selvanair:
Some comments on the PR:

(i) The correct import location will be set only after registry is parsed, so import should not happen before that. Even otherwise, "ProcessCommandLine" needs to be after scanning the registry. Instead of the reordering, save the result of "already running check", and act on it only after parsing the command line. Also bypass "UpdateRegistry" if another instance is running.
(ii) Add a call to EnsureDirExists(o.config_dir) in the ImportConfigFile function as the latter may now get called before the config directory is created in user profile.
(iii) The way its coded, on double clicking, the config gets copied to the "config_dir" directory and the GUI exits.  This is confusing to the user unless an instance of the GUI is already running. In that case, anyway,  import using the existing "import file.." menu is equally effective. So I suggest some improvements:

(a)  If no other instance of the GUI is running this instance should just run as normal after import instead of exiting.
(b) Intuitively, a user would expect the connection to start on double-clicking, not just import. So trigger an auto-connect after import (in the already running instance, if any or in this new instance).
(c) When double clicked on a config that is already imported, just start the connection.

ValdikSS:
>(a) If no other instance of the GUI is running this instance should just run as normal after import instead of exiting.
(b) Intuitively, a user would expect the connection to start on double-clicking, not just import. So trigger an auto-connect after import (in the already running instance, if any or in this new instance).
(c) When double clicked on a config that is already imported, just start the connection.

I would argue with all three statements.
Imagine I have a configuration file in an email message, and I try to import already imported configuration file. I (personally) would not expect the connection to start or GUI to keep running.

ValdikSS:
>However, this will take away the user's option to set an editor of their choice. We hav eto somehow remedy that as well. 

What do you think is a best way to deal with this? Should I add a path to editor into "advanced" tab or just hard-code notepad.exe again?

selvanair:
> > However, this will take away the user's option to set an editor of their choice. We hav eto somehow remedy that as well.
> 
> What do you think is a best way to deal with this? Should I add a path to editor into "advanced" tab or just hard-code notepad.exe again? Or the registry value probably be fine?

I am inclined towards making this a configurable option again (in one of the settings tab, as you say), but let's first decide how best to handle the --import option. More on that in a separate comment.

ValdikSS:
Or perhaps we should open editor which is associated with *.txt files? I think it's pretty common to associate favourite editor with *.txt and personally I would expect this.

selvanair:
> > a) If no other instance of the GUI is running this instance should just run as normal after import instead of exiting.
> > (b) Intuitively, a user would expect the connection to start on double-clicking, not just import. So trigger an auto-connect after import (in the already running instance, if any or in this new instance).
> > (c) When double clicked on a config that is already imported, just start the connection.
> 
> I would argue with all three statements
> Imagine I have a configuration file in an email message, and I try to import already imported configuration file. I (personally) would not expect the connection to start or GUI to keep running.

If the purpose of this option is to just install the config and don't do anything else, that's fine with me. But then, do not link the default double-click option for .ovpn to this. Instead, add a context menu for .ovpn  named, say, "install" linked to this action and let the user right click and choose "install" to copy the file to the right location. (Here I'm referring to PR 26 for openvpn-build that goes with this PR).

Double-click is best reserved for a more active action (from user's point of view): like starting the config in-place or optionally after an import. We can consider that later.

That said, I just don't understand the need for such an special purpose --import option as the GUI can already import configs using the "import file.." menu.  If this make the life easier for a some users, fine. But please do not make this double click action.

Also note that the way its handled in this patch causes all kinds of option ordering issues. Consider a user doing this:

openvpn-gui.exe --config_dir xxx --import yyy
vs
openvpn-gui.exe --import yyy --config_dir xxx

yyy gets imported to a different location depending on the order or options.

Either the import option should be parsed separately and treated as special and explicitly not allow any other options on the commandline or the  import action should happen only after all command line options are parsed. Unless we want to hide this option from users.

selvanair:
> Or perhaps we should open editor which is associated with *.txt files? I think it's pretty common to associate favourite editor with *.txt and personally I would expect this.

That's an option although its a hackish solution: it could be confusing to a user that the GUI uses a different editor for .ovpn because the user changed the editor for .txt files. 

Anyway, we need not do anything about this until the default action is modified and based on the above discussion, this PR does not need to modify the default action.

ValdikSS:
>If the purpose of this option is to just install the config and don't do anything else, that's fine with me. But then, do not link the default double-click option for .ovpn to this. Instead, add a context menu for .ovpn named, say, "install" linked to this action and let the user right click and choose "install" to copy the file to the right location. (Here I'm referring to PR 26 for openvpn-build that goes with this PR).
>
>Double-click is best reserved for a more active action (from user's point of view): like starting the config in-place or optionally after an import. We can consider that later.

You're absolutely correct from the technical standpoint, but there are some widely used cases where you either can't get right click menu or you just expect import dialog.

Example 1: User got OpenVPN configuration file via email and opens email message via web interface using browser. When the user clicks on the attachment, browser asks should we open or save this file. If user selects "open", it is expected to get import dialog. You can't get right click menu in some browsers, so you have to navigate to the downloads folder manually or press "open containing folder" on that file (which is not available in all browsers also).

Example 2: User uses email software and has 2 different OpenVPN configuration files in 2 different emails, but these configuration files have the same name. If we start connection if the configuration file is already imported, should we check configuration by it's hash or file size, not only by name? If we don't, we'll connect using completely different configuration file (but with the same name).

>That said, I just don't understand the need for such an special purpose --import option as the GUI can already import configs using the "import file.." menu. If this make the life easier for a some users, fine. But please do not make this double click action.

Sometimes it's not handy to navigate to the directory manually. In case of email client or browser, you should first save configuration file somewhere and then run GUI, select "import", go to the directory and shit. Me and you are probably use computers in a different manner and expect different things.

ValdikSS:
>That's an option although its a hackish solution: it could be confusing to a user that the GUI uses a different editor for .ovpn because the user changed the editor for .txt files. 

Do people often change editors? I believe that usual user have maximum of 3 editors: notepad.exe, some IDE and a generic editor like notepad++ or similar, which is associated with *.txt. I don't think that's a hackish solution.

ValdikSS:
So, what do you think? How should we implement it? If you're completely opposite the idea of making import-only functionality as the main action, would it be fine if I make a dialog with multiple actions, like import, connect, edit?

selvanair:
> 
> > That's an option although its a hackish solution: it could be confusing to a user that the GUI uses a different editor for .ovpn because the user changed the editor for .txt files.
> 
> Do people often change editors? I believe that usual user have maximum of 3 editors: notepad.exe, some IDE and a generic editor like notepad++ or similar, which is associated with *.txt. I don't think that's a hackish solution.

It has nothing to do with number of editors. Its non-ideal only because of the reason I mentioned. Anyway, I will do a patch to make this configurable from the settings menu, problem solved. It was indeed an option that could be changed in the registry earlier before I changed it to use file association.

ValdikSS:
>It has nothing to do with number of editors. Its non-ideal only because of the reason I mentioned. Anyway, I will do a patch to make this configurable from the settings menu, problem solved. It was indeed an option that could be changed in the registry earlier before I changed it to use file association.

Well, .ovpn file is a text file, and it's more user-friendly to use *.txt associated editor than notepad.exe. All my configuration files are with UNIX-style newlines and notepad.exe can't handle them, showing me a lot of garbage on the screen which I can't edit. If the user have any other editor, it's better to use it by default than notepad.exe, or, if you're going to make editor configurable, get path to associated *.txt editor and set it by default instead of notepad.exe

I've implemented it, take a look at https://github.com/ValdikSS/openvpn-gui/commit/2090f4fb7240e9dbe7249ffbf3cf8d39923a37bc

selvanair:
> So, what do you think? How should we implement it? If you're completely opposite the idea of making import-only functionality as the main action, would it be fine if I make a dialog with multiple actions, like import, connect, edit?

That's the way I would go. For edit menu, the installer should not overwrite any editor already set. Then we can provide an option to change the editor from the GUI and use the editor for .txt as a fallback as you suggested.

Having connect is useful but it has to be implemented by passing the message to any running instance which is quite different from the approach used here.  That also should take care of more gracefully handling already running instances of the GUI (see also [Trac  #394]
(https://community.openvpn.net/openvpn/ticket/394)) Note that the existing --connect option does not work with full path.

So I would say keep this PR to just implement --import and probably "edit". Let's handle the rest separately. You may also want to let the user force-overwrite an existing config with the same name so that updated configs can be easily distributed.

Finally, what would be the default action then?

ValdikSS:
All right, I'll just implement --import and --edit in this PR and in next PR I'll make an import/edit dialog for double-click (without connecting yet). Is it fine?

selvanair:
@ValdikSS : Sounds good. 

For edit I looked at the repo you linked to (using associated exec of .txt). Looks ok, except that in case log files better use the the default handler for "open" for log files first and fall back to that for .txt only if that fails. Finally "notepad.exe" if all fails. Similarly for editing .ovpn -- if the user has set an edit handler that should get preference, falling back to the one for .txt files otherwise.

mattock:
@ValdikSS : what is the status of this? We should try to get this into OpenVPN-GUIs bundled with 2.4-alphas, and 2.4-alpha1 is going to get release late this week hopefully.

ValdikSS:
@mattock I'm bit busy. Will try to work on weekend on this.

mattock:
@ValdikSS : no problem. We have not discussed how to handle OpenVPN-GUI updates in the OpenVPN 2.4 installers. Probably at some point we will have to create a "release/2.4" branch in OpenVPN-GUI, which will then only receive small fixes and improvements. @selvanair : thoughts?

selvanair:
@mattock We can continue to add new stuff at least until 2.4 beta, I suppose. Even after 2.4 stable, most of the patches could go into both master and release until we add some feature that does not belong to the release version.

To keep track of the GUI version in released binaries, it would be useful to have a build number (could be number of commits past a tag) plus a short commit hash in, say, the About page. Alternatively we could update the build number of PACKAGE_VERSION_RESOURCE (currently set to _GUI_VERSION.0.0.0) whenever a new binary is made. _GUI_VERSION can still stay at the single number (such as 11) so that the build-scripts are not affected.

While at it, we should update the copyright in About page before 2.4 alpha. Any suggestions on how to attribute the copyright for development beyond the original 2004-2005? The copyright line referring to OpenVPN also need to be updated.

mattock:
@selvanair : makes sense. I vote for incrementing PACKAGE_VERSION_RESOURCE on each release, as it will / should show up in the file properties on Windows. The four-digit version number scheme is quite MS-specific:

* https://en.wikipedia.org/wiki/Microsoft_version_numbering
* https://blogs.msdn.microsoft.com/jim_glass/2007/02/16/microsoft-product-build-version-numbers/

I think incrementing the minor version (second digit) on each release would be sufficient in our case.

I created PR https://github.com/OpenVPN/openvpn-gui/pull/88 that updates the About page.

selvanair:
@mattock  For our purposes we could interpret the four numbers in the VERSION_RESOURCE as
"major.minor.build.revision" each number is a short (that's how its named in the microsoft header files although service packs uses a diffferent interpretation)

Anyway, incrementing major.minor would work fine with the new version tracking logic used to trigger a possible registry update (i.e., look for deprecated variables etc.). That said, we should change the version not just with new releases of openvpn but whenever a new windows binary is distributed (which could be in between two releases). For that we could either increase the build number or the minor number, I have no strong preference.

ValdikSS:
@selvanair, @mattock, please check 2 new commits.

Changes:
* `--import` removed
* `--edit` added, tries to open configuration file in the text editor associated with "*.log" first, then, if it failed, with "*.txt" editor, and finally in notepad.
* `--configfile_dialog` added. A dialog which prompts for import/edit. This dialog is for double click action.

Installer could be updated to add "edit" item for right-click "*.ovpn" menu, which would call `openvpn-gui.exe --edit`

main.c
```diff
@@ -160,22 +160,28 @@ int WINAPI _tWinMain (HINSTANCE hThisInstance,
...
-
-  UpdateRegistry(); /* Checks version change and update keys/values */
 
   GetRegistryKeys();
+
+  EnsureDirExists(o.config_dir);
   /* Parse command-line options */
   ProcessCommandLine(&o, GetCommandLine());
 
-  EnsureDirExists(o.config_dir);
```
selvanair: Moving this up is problematic in normal use as o.config_dir can change after ProcessCommandLine. Anyway existence of o.config_dir is ensured in ImportConfigFile() so keep the original order.

main.c
```diff
@@ -528,40 +534,64 @@ CloseApplication(HWND hwnd)
...
-    _sntprintf_0(filter, _T("*.%s%c*.%s%c"), o.ext_string, _T('\0'), o.ext_string, _T('\0'));
-    
-    OPENFILENAME fn;
+
+void
+ImportConfigFile(PTCHAR config_file)
+{
     TCHAR source[MAX_PATH] = _T("");
+    PTCHAR fileName;
+    BOOL result = 1;
```
selvanair: Nitpicking: TRUE is better than 1.

main.c
```diff
@@ -528,40 +534,64 @@ CloseApplication(HWND hwnd)
...
+        fn.lpstrInitialDir = NULL;
+        fn.lpstrTitle = NULL;
+        fn.Flags = OFN_DONTADDTORECENT | OFN_FILEMUSTEXIST;
+        fn.lpstrDefExt = NULL;
+
+        result = GetOpenFileName(&fn) && (fileName = source + fn.nFileOffset);
+    }
+    else
     {
+        fileName = PathFindFileName(config_file);
```
selvanair: We should check that config_file exists -- its not guaranteed to be a valid file as it comes from a command line. fileName is used below to create a subdir, so we want it to be a sane name.

main.c
```diff
@@ -528,40 +534,64 @@ CloseApplication(HWND hwnd)
...
+        _tcscpy(source, config_file);
```
selvanair: This is not safe, use wcsncpy or equivalent and null terminate source.

main.c
```diff
@@ -528,40 +534,64 @@ CloseApplication(HWND hwnd)
...
+    {
         TCHAR destination[MAX_PATH];
-        PTCHAR fileName = source + fn.nFileOffset;
 
         _sntprintf_0(destination, _T("%s\\%s"), o.config_dir, fileName);
 
         destination[_tcslen(destination) - _tcslen(o.ext_string) - 1] = _T('\0');
         
-        if (EnsureDirExists(destination))
+        if (EnsureDirExists(destination) && EnsureDirExists(o.config_dir))
```
selvanair: This change is not needed. If  EnsureDirExists(destination) succeeds, the next call will too as destination is a subdirectory of o.config_dir.  All necessary parent directories are automatically created by the first call. So the original code is fine.

main.c
```diff
@@ -570,7 +600,7 @@ ImportConfigFile()
             {
                 if (GetLastError() == ERROR_FILE_EXISTS)
                 {
```
selvanair: As the main purpose of this is for ease of distributing configs, don't we want to allow optional overwrite of an existing config?

viewlog.c
```diff
@@ -54,13 +57,20 @@ void ViewLog(int config)
...
+  if (AssocQueryString(0, ASSOCSTR_EXECUTABLE, L".log", NULL, assocexe, &assocexe_num) == S_OK)
+    status = ShellExecuteW (o.hWnd, L"open", assocexe, o.conn[config].log_path, o.log_dir, SW_SHOWNORMAL);
   if (status > (HINSTANCE) 32) /* Success */
     return;
   else
-    PrintDebug (L"Opening log file using ShellExecute with verb = open failed"
-                 " for config '%s' (status = %lu)", o.conn[config].config_name, status);
+  {
+    if (AssocQueryString(0, ASSOCSTR_EXECUTABLE, L".txt", NULL, assocexe, &assocexe_num) == S_OK)
+      status = ShellExecuteW (o.hWnd, L"open", assocexe, o.conn[config].log_path, o.log_dir, SW_SHOWNORMAL);
```
selvanair: If we are using ShellExecuteW with an executable name, the rest of the code can be removed and simply replace use assocexe if available or o.editor otherwise. No CreateProcess required.

Same with EditConfig()

mattock:
@ValdikSS : in today's community meeting we agreed to include (if possible) all the current openvpn-gui PRs before OpenVPN 2.4_rc1 release (in ~2 weeks). I'd like to build a test installer that includes as many openvpn-gui PRs as possible to facilitate testing.

When could you fix the few issues pointed out by @snair?

ValdikSS:
@mattock yes I'll try my best on this weekend.

mattock:
@ValdikSS : Excellent! It would be really good to have this functionality in openvpn-gui.exe for testing prior to 2.4_rc1.

main.c
```diff
@@ -528,40 +534,64 @@ CloseApplication(HWND hwnd)
...
-ImportConfigFile()
+ShowConfigFileDialog(PTCHAR config_file)
 {
-    
-    TCHAR filter[2*_countof(o.ext_string)+5];
+    PTCHAR fileName;
+    int dlgresult = 0;
+
+    fileName = PathFindFileName(config_file);
+    dlgresult = ShowLocalizedMsgEx(MB_YESNO, _T(PACKAGE_NAME), IDS_ASK_IMPORT_OR_EDIT, fileName);
```
selvanair: Allow user to cancel the dialog

main.c
```diff
@@ -528,40 +534,64 @@ CloseApplication(HWND hwnd)
...
+    }
 
+    if (result)
+    {
         TCHAR destination[MAX_PATH];
-        PTCHAR fileName = source + fn.nFileOffset;
 
         _sntprintf_0(destination, _T("%s\\%s"), o.config_dir, fileName);
 
         destination[_tcslen(destination) - _tcslen(o.ext_string) - 1] = _T('\0');
```
selvanair: This is now risky as destination string could be anything, so use PathRemoveExtension here too.

viewlog.c
```diff
@@ -40,12 +41,14 @@ extern options_t o;
 void ViewLog(int config)
 {
   TCHAR filename[2*MAX_PATH];
+  TCHAR assocexe[2*MAX_PATH];
+  DWORD assocexe_num;
```
selvanair: size of MAX_PATH for files (paths) and 2*MAX_PATH only for the command line which is made of two paths.

viewlog.c
```diff
@@ -54,13 +57,20 @@ void ViewLog(int config)
 
   /* Try first using file association */
   CoInitializeEx(NULL, COINIT_APARTMENTTHREADED | COINIT_DISABLE_OLE1DDE); /* Safe to init COM multiple times */
-  status = ShellExecuteW (o.hWnd, L"open", o.conn[config].log_path, NULL, o.log_dir, SW_SHOWNORMAL);
-
+  if (AssocQueryString(0, ASSOCSTR_EXECUTABLE, L".log", NULL, assocexe, &assocexe_num) == S_OK)
```
selvanair: This will mis-behave (most of the time fails, could segfault) unless assocexe_num is set to the size of the buffer:
assocexe_num = _countof(assocexe)

viewlog.c
```diff
@@ -54,13 +57,20 @@ void ViewLog(int config)
...
+    status = ShellExecuteW (o.hWnd, L"open", assocexe, o.conn[config].log_path, o.log_dir, SW_SHOWNORMAL);
   if (status > (HINSTANCE) 32) /* Success */
     return;
   else
-    PrintDebug (L"Opening log file using ShellExecute with verb = open failed"
-                 " for config '%s' (status = %lu)", o.conn[config].config_name, status);
+  {
+    if (AssocQueryString(0, ASSOCSTR_EXECUTABLE, L".txt", NULL, assocexe, &assocexe_num) == S_OK)
```
selvanair: same here: define assocexe_num = .. as shown above

viewlog.c
```diff
@@ -91,34 +101,54 @@ void ViewLog(int config)
   CloseHandle(proc_info.hProcess);
 }
 
-
-void EditConfig(int config)
+/* either config or config_path */
+void EditConfig(int config, PTCHAR config_path)
 {
```
selvanair: make config_path const.

viewlog.c
```diff
@@ -91,34 +101,54 @@ void ViewLog(int config)
...
-  TCHAR filename[2*MAX_PATH];
+  TCHAR config_file[2*MAX_PATH];
+  TCHAR fullpath[2*MAX_PATH];
+  TCHAR config_dir[2*MAX_PATH];
+  TCHAR assocexe[2*MAX_PATH];
+  DWORD assocexe_num;
```
selvanair: size MAX_PATH enough for all paths, 2*MAX_PATH only for the command line. Also eliminate config_file as its only used in the debug print which can use the fullpath instead.

viewlog.c
```diff
@@ -91,34 +101,54 @@ void ViewLog(int config)
...
   CLEAR (sd);
 
-  /* Try first using file association */
-  _sntprintf_0(filename, L"%s\\%s", o.conn[config].config_dir, o.conn[config].config_file);
+  if (config_path == NULL)
+  {
+    // using "config"
+    _sntprintf_0(fullpath, L"%s\\%s", o.conn[config].config_dir, o.conn[config].config_file);
+    _tcsncpy(config_dir, o.conn[config].config_dir, _countof(config_dir) - 1);
+    _tcsncpy(config_file, o.conn[config].config_file, _countof(config_file) - 1);
```
selvanair: _tsncopy is not guaranteed to NUL terminate. Add explicit termination.

viewlog.c
```diff
@@ -91,34 +101,54 @@ void ViewLog(int config)
...
+  }
+  else
+  {
+    // using "config_path"
+    _tcsncpy(fullpath, config_path, _countof(config_dir) - 1);
+    _tcsncpy(config_dir, config_path, _countof(config_dir) - 1);
```
selvanair: NUL termination needed here too

viewlog.c
```diff
@@ -91,34 +101,54 @@ void ViewLog(int config)
...
+    PathRemoveFileSpec(config_dir);
+    _tcsncpy(config_file, config_path, _countof(config_file) - 1);
+    *config_file = PathFindFileName(config_file);
```
selvanair: Assigning *config_file to a pointer? That completely messes up that string..

viewlog.c
```diff
@@ -91,34 +101,54 @@ void ViewLog(int config)
...
+  }
 
+  /* Try first using file association */
   CoInitializeEx(NULL, COINIT_APARTMENTTHREADED | COINIT_DISABLE_OLE1DDE); /* Safe to init COM multiple times */
-  status = ShellExecuteW (o.hWnd, L"open", filename, NULL, o.conn[config].config_dir, SW_SHOWNORMAL);
+  if (AssocQueryString(0, ASSOCSTR_EXECUTABLE, L".txt", NULL, assocexe, &assocexe_num) == S_OK)
```
selvanair: Define assocexe_num (see comment above), else the call will fail at best or segfault.

viewlog.c
```diff
@@ -91,34 +101,54 @@ void ViewLog(int config)
...
+    status = ShellExecuteW (o.hWnd, L"open", assocexe, fullpath, config_dir, SW_SHOWNORMAL);
   if (status > (HINSTANCE) 32)
     return;
   else
     PrintDebug (L"Opening config file using ShellExecute with verb = open failed"
-                 " for config '%s' (status = %lu)", o.conn[config].config_name, status);
+                 " for config '%s' (status = %lu)", config_file, status);
```
selvanair: eliminate config_file and just use fullpath in the debug print.

viewlog.c
```diff
@@ -91,34 +101,54 @@ void ViewLog(int config)
...
 
-  _sntprintf_0(filename, _T("%s \"%s\\%s\""), o.editor, o.conn[config].config_dir, o.conn[config].config_file);
+  _sntprintf_0(fullpath, _T("%s \"%s\""), o.editor, fullpath);
```
selvanair: Overlapped destination and argument string in snprintf family is undefined.  And this particular usage is guaranteed to mess up fullpath.

viewlog.h
```diff
@@ -20,4 +20,4 @@
  */
 
 void ViewLog(int config);
-void EditConfig(int config);
+void EditConfig(int config, PTCHAR config_path);
```
selvanair: add const to config_path

ValdikSS:
@selvanair, thanks a lot for your hard work! I'm short in time and can't devote this patch much time. Just applied your patch on top of my, it's in this pull request now.

ValdikSS:
@selvanair is this patch already late to be included into release? If I fix conflict and test it thoughtfully today and tomorrow, can it be still merged?

mattock:
@ValdikSS : OpenVPN 2.4.0 [will be released](https://community.openvpn.net/openvpn/wiki/StatusOfOpenvpn24) on 28th so there is still some time. If the final version of this PR is available (and approved) by, say, 23rd (upcoming Friday), then we'd have ample time to test it fix any issues.

selvanair:
Although the code may be just ok in the current form, the user experience is rather poor because when the config exists it just bails out -- no way to update an existing config. I think it should be smarter in detecting whether the user is clicking on an already installed config and if not offer to update etc. I would say let's leave this for 2.4.1 and try to offer a better UX.

selvanair:
Closing this as obsoleted by recent changes to the import feature.

PR #78, opened 2016-09-09
Add CMake-based build system

evpobr:
This commit brings [CMake](www.cmake.org) support to OpenVPN Gui.
# Generating makefile

It is recommended not to build in source directory, so change it:

```
mkdir build
cd build/
```

Then run command:

```
cmake .. -G"MSYS Makefiles"
```

Debug configuration is enabled by default, to create release configuration:

```
cmake .. -G"MSYS Makefiles" -DCMAKE_BUILD_TYPE=Release
```

You can set install prefix:

```
cmake .. -G"MSYS Makefiles" -DCMAKE_INSTALL_PREFIX=/mingw
```
## OpenVPN GUI options

To change maximum of config files (50 by default) set `MAX_CONFIGS`
variable:

```
cmake .. -G"MSYS Makefiles" -DMAX_CONFIGS=64
```

To disable change password feature, if OpenSSL package was found:

```
cmake .. -G"MSYS Makefiles" -DDISABLE_CHANGE_PASSWORD=ON
```

If OpenSSL package is not found, `DISABLE_CHANGE_PASSWORD` will be set
to 1 automatically.
# Building

Now you have usual makefile, so type:

```
make
```

If you want verbose mode:

```
make VERBOSE=1
```
# Installation

To install OpenVPN Gui type:

```
make install
```

To install stripped files:

```
make install/strip
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,66 @@
+cmake_minimum_required (VERSION 3.0.2)
+
+project (openvpn-gui VERSION 11)
+
+include (GNUInstallDirs)
+
+set (PACKAGE "openvpn-gui")
+set (PACKAGE_BUGREPORT "openvpn-devel@lists.sourceforge.net")
+set (PACKAGE_NAME "OpenVPN GUI")
+set (PACKAGE_STRING "${PACKAGE_NAME} ${PROJECT_VERSION_MAJOR}")
+set (PACKAGE_TARNAME "openvpn-gui")
+set (PACKAGE_URL "https://github.com/openvpn/openvpn-gui/")
+set (PACKAGE_VERSION "${PROJECT_VERSION}")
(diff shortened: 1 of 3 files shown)


cron2:
Hi,

On Fri, Sep 09, 2016 at 12:32:02AM -0700, evpobr wrote:
> This commit brings [CMake](www.cmake.org) support to OpenVPN Gui.

So what is the benefit of this?  We seem to be able to build the GUI
quite nicely with the existing build system...  and maintaining multiple
build systems usually leads to "one of them is forgotten, and breaks".

So unless there are more conving arguments than "because we can" I tend
to feature-NAK this.

gert

-- 
                                                           //www.muc.de/~gert/

mattock:
I'll close this PR as we still don't have answer to the _"why is this needed?"_ question after more than 2 months.

PR #80, opened 2016-09-17
cleaned up unused resource

chipitsine:


mattock:
The resource indeed seems to be unused. The only reason for keeping afaics is to prevent accidental reuse the same error code. Even so the translations could be removed, and the define could just be commented out and marked as "historic". @selvanair : thoughts?

chipitsine:
what is wrong in reusing the same error code ?

mattock:
@chipitsine : I think it's mostly just about code hygiene. Also, Googling for an error code should preferably only return relevant results, not a mix of relevant and irrelevant, depending on how old version of OpenVPN-GUI the author of the webpage had used.

selvanair:
I agree with @mattock -- we can leave the define in openvpn-gui-res.h but remove the translations.

mattock:
Ok, this looks good now. Merging.

PR #84, opened 2016-09-20
Improve the message shown when no config files are found

selvanair:
The popup message shown when no config files are present now includes a pointer to the "Import File.." menu and the location of the user and global config directories.

Note: Translations of the added text is missing.

### DIFF
--- a/openvpn_config.c
+++ b/openvpn_config.c
@@ -235,7 +235,7 @@ BuildFileList()
         BuildFileList0 (o.global_config_dir, issue_warnings);
 
     if (o.num_configs == 0 && issue_warnings)
-        ShowLocalizedMsg(IDS_NFO_NO_CONFIGS);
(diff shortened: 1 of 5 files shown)


schwabe:
Changes look good, ACK from my side. Thanks for fixing it so quick!

ftoledo:
Can i disabled this popup via registry key or some config settings?
We distribute computers to new users with pre installed openvpn gui, i search the way to do not get this message when user still do not have your config. 
thanks

selvanair:
You can suppress it by having a registry value (DWORD) in `HKCU/Software/OpenVPN-GUI/silent_connection = 1` or using `--silent_connection 1` on command line. But that has other side effects like status/log window not popped up during connection, duplicate configs are silently ignored etc.

Note that the reg key is per user (i.e. in HKCU). It may be a useful to take defaults from HKLM when HKCU value is not present to ease such deployments. PRs welcome.

PR #86, opened 2016-10-09
Some bug fixes

selvanair:
See commit messages for details

### DIFF
--- a/openvpn-gui-res.h
+++ b/openvpn-gui-res.h
@@ -296,4 +296,7 @@
 /* Save password related messages */
 #define IDS_NFO_DELETE_PASS             2001
 
+/* Timer IDs */
+#define IDT_STOP_TIMER                  2500  /* Timer used to trigger force termination */
+
 #endif
--- a/openvpn.c
+++ b/openvpn.c
@@ -474,7 +474,8 @@ OnStop(connection_t *c, UNUSED char *msg)
             SetForegroundWindow(c->hwndStatus);
             ShowWindow(c->hwndStatus, SW_SHOW);
         }
-        ShowLocalizedMsg(IDS_NFO_CONN_TERMINATED, c->config_name);
+        MessageBox(c->hwndStatus, LoadLocalizedString(IDS_NFO_CONN_TERMINATED, c->config_file),
+                   _T(PACKAGE_NAME), MB_OK);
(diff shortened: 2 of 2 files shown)


cron2:
Hi,

On Sat, Oct 08, 2016 at 11:48:51PM -0700, Selva Nair wrote:
> -- Commit Summary --
> 
>   * NUL terminate messages received from interactive service
>   * Terminate any openvpn processes that fail to stop
>   * Do not start a connection when a previous thread has not fully exited

Feature-ACK "because it makes sense".

The code looks reasonable, though I'm (still) not a windows coder... so
a true review would be better.

gert
-- 
                                                           //www.muc.de/~gert/

selvanair:
> Does 9892d58 solve some particular issue, or is it just "good practice"?
(nul-terminate strings read from service)

We take the length of the string received from the service using wcslen which gives erratic results (including potential buffer overrun) when not properly nul terminated. As we do not read much from the service, this showed up only when I added support for reading the process PID from service. Not easy to reproduce as it depends on the contents of the buffer past the good data. But required for getting PID from service to work reliably.

>  Commits fcd0efa
(terminate process that does not gracefully stop)

Stale openvpn processes being left behind  happened many times to me due to signals getting overwritten in openvpn etc. But has become is less of an issue after we patched openvpn to fix exit event getting ignored. Though it can still happen in some situations. That said, at this point this is more of a defensive programming.
 
> and ad58766 get a feature-ACK from me. That said, I'd like to test that they work as intended. Any pointers on how to do that?

To test the 3rd commit,  "do not start a new connection before the previous one has exited", try the following:  Stop interactive service, use the current GUI, add some junk directive to the config file, and start a connection. It will fail promptly due to the error in config and a message-box will pop up. Leave the pop up and status window as is (i.e do not press OK), go to the main menu (i.e right click on tray icon) and click connect again for the same config. It will fail again poping up another message box. Now there are two status windows and two message boxes open. Click OK on both the message boxes: one of the status window will close automatically and the other will hang around. Watch the CPU usage by the GUI process -- one core get saturated.

Well, this is an unusual kind of usage, but easy to happen when there are many windows covering the desktop and the user does not notice a pending previous status window that may be hidden behind.

mattock:
I'll merge this with a lazy-ACK so that we get this into 2.4-alpha1, which has already been tagged.

PR #88, opened 2016-10-13
Update About page

mattock:
Proposal based  on discussion in PR https://github.com/OpenVPN/openvpn-gui/pull/76

Right now the text might overflow, but I'll fix that after we've agreed upon the content.

PR #89, opened 2016-10-19
Update About page

mattock:
After some creative use of forced pushes, remotes and branching my original pull request (https://github.com/OpenVPN/openvpn-gui/pull/88) went up in a smoke. So here is an amended version that also incorporates feedback from @selvanair . I also linked to OpenVPN-GUI's GitHub page instead of openvpn.net: that should make it more obvious who the "OpenVPN GUI contributors" are. 

### DIFF
--- a/res/openvpn-gui-res-en.rc
+++ b/res/openvpn-gui-res-en.rc
@@ -187,14 +187,16 @@ BEGIN
     ICON ID_ICO_APP, 0, 8, 16, 21, 20
     LTEXT "OpenVPN GUI v" PACKAGE_VERSION " - A Windows GUI for OpenVPN\n" \
           "Copyright (C) 2004-2005 Mathias Sundman <info@openvpn.se>\n" \
-          "http://openvpn.se/", 0, 36, 15, 206, 26
+          "Copyright (C) 2008-2014 Heiko Hund <heikoh@users.sf.net>\n" \
+          "Copyright (C) 2012-2016 OpenVPN GUI contributors\n" \
+          "https://github.com/OpenVPN/openvpn-gui/", 0, 36, 15, 206, 42
     LTEXT "OpenVPN - An application to securely tunnel IP networks " \
-          "over a single UDP port, with support for SSL/TLS-based " \
+          "over a single TCP/UDP port, with support for SSL/TLS-based " \
           "session authentication and key exchange, packet " \
           "encryption, packet authentication, and packet compression.\n" \
           "\n" \
-          "Copyright (C) 2002-2005 OpenVPN Solutions LLC <info@openvpn.net>\n" \
-          "http://openvpn.net/", 0, 8, 45, 235, 56
+          "Copyright (C) 2002-2016 OpenVPN Technologies, Inc <info@openvpn.net>\n" \
+          "https://openvpn.net/", 0, 8, 61, 240, 64
 END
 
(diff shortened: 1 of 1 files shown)


res/openvpn-gui-res-en.rc
```diff
@@ -187,14 +187,16 @@ BEGIN
     ICON ID_ICO_APP, 0, 8, 16, 21, 20
     LTEXT "OpenVPN GUI v" PACKAGE_VERSION " - A Windows GUI for OpenVPN\n" \
           "Copyright (C) 2004-2005 Mathias Sundman <info@openvpn.se>\n" \
-          "http://openvpn.se/", 0, 36, 15, 206, 26
+          "Copyright (C) 2008-2014 Heiko Hund <heikoh@users.sf.net>\n" \
+          "Copyright (C) 2012-2016 OpenVPN GUI contributors\n" \
+          "https://github.com/OpenVPN/openvpn-gui/", 0, 36, 15, 206, 26
```
selvanair: 0, 36, 15, 206, 26 ---> 0, 36, 15, 206, 42 so as not to clip the text

res/openvpn-gui-res-en.rc
```diff
@@ -187,14 +187,16 @@ BEGIN
...
     LTEXT "OpenVPN - An application to securely tunnel IP networks " \
-          "over a single UDP port, with support for SSL/TLS-based " \
+          "over a single TCP/UDP port, with support for SSL/TLS-based " \
           "session authentication and key exchange, packet " \
           "encryption, packet authentication, and packet compression.\n" \
           "\n" \
-          "Copyright (C) 2002-2005 OpenVPN Solutions LLC <info@openvpn.net>\n" \
-          "http://openvpn.net/", 0, 8, 45, 235, 56
+          "Copyright (C) 2002-2005 OpenVPN Technologies, Inc <info@openvpn.net>\n" \
```
selvanair: That should be "Copyright (C) 2002-2016...." isn't it?

res/openvpn-gui-res-en.rc
```diff
@@ -187,14 +187,16 @@ BEGIN
...
+          "https://openvpn.net/", 0, 8, 45, 235, 56
```
selvanair: 0, 8, 45, 235, 56 -> 0, 8, 61, 240, 64

mattock:
@selvanair : all the issues should be fixed now.

selvanair:
Looks good. Merging.

selvanair:
@mattock: I'm preparing a patch that copies the updated copyright message to all lang files (where missing) -- no translation required. This may be good opportunity to change how version number is shown on the About tab as well. Currently we have 
"OpenVPN GUI v" PACKAGE_VERSION --> OpenVPN GUI v11
We could make the version read $major.$minor using a macro or go full 11.1.0.0 style using PACKAGE_VERSION_RESOURCE_STR or leave as is. Any thoughts?

mattock:
If the version number in "About" tab is generated from data in configure.ac, then I think full four digit version makes most sense.

selvanair:
Yes, its a preprocessor macro, so any define generated by configure (config.h) with a string value will work. Will use the 4-digit dotted decimal denoted by  PACKAGE_VERSION_RESOURCE_STR (soon to be 11.2.0.0)

PR #91, opened 2016-11-03
pkcs11 pin prompt

selvanair:
Support pkcs11 token insertion request and token pin input

(This is built on top of PR 87, so ignore the first commit)


selvanair:
@valdikSS: could you please give a review of this? This one and dynamic CR support are two useful features to have in the GUI.

selvanair:
PKCS11 prompt function tested by @bjoernv on [Trac 740#16](https://community.openvpn.net/openvpn/ticket/740#comment:15)

selvanair:
@mattock I think we may have to use lazy ACK on this to get this into 2.4-rc1 and hopefully some testing in the wild before 2.4. We can ignore PR 87 as that one is the same as the first commit here.

mattock:
@selvanair : agreed. The endless loop thing was fixed at the OpenVPN side, right?

selvanair:
No, the endless loop is not fixed. It seems no one understands that part of the code in OpenVPN -- at least I did not get any response. Anyway it apparently shows up only if the token is removed mind-way through the password dialog (I haven't reproduced it) and is reportedly seen in command line usage as well. Its not a runaway loop, openvpn just re-prompts the user to insert the token even if the user presses cancel.

cron2:
Hi,

On Tue, Nov 29, 2016 at 10:31:35PM -0800, Samuli Seppänen wrote:
> @selvanair : agreed. The endless loop thing was fixed at the OpenVPN side, right?

No, that was another endless loop (much easier to understand, just our
own code, no pkcs11-helper involved).

Maybe @syzzer has some ideas, or @jjk.

gert
-- 
                                                           //www.muc.de/~gert/

chipitsine:
2016-11-30 10:28 GMT+03:00 Gert Doering <notifications@github.com>:

> Hi,
>
> On Tue, Nov 29, 2016 at 10:31:35PM -0800, Samuli Seppänen wrote:
> > @selvanair : agreed. The endless loop thing was fixed at the OpenVPN
> side, right?
>
> No, that was another endless loop (much easier to understand, just our
> own code, no pkcs11-helper involved).
>

can we involve some static analyzer (like clang or coverity) to detect
endless loops ?

or "token was removed" is rather impossible to detect ?

cron2:
Hi,

On Tue, Nov 29, 2016 at 11:57:31PM -0800, Ilya Shipitsin wrote:
> 2016-11-30 10:28 GMT+03:00 Gert Doering <notifications@github.com>:
> 
> can we involve some static analyzer (like clang or coverity) to detect
> endless loops ?

"brains" is more helpful here :-) - because that loop is usually *not*
endless.  

It's likely some sort of retry-on-failure loop that does not handle 
certain error conditions right.

gert
-- 
                                                           //www.muc.de/~gert/

ValdikSS:
What endless loop are you talking about? I have some problems with pkcs11 in OpenVPN in Linux, it deadlocks sometime, so I need to run it with strace, that way the probability of deadlock is much less.

selvanair:
On Wed, Nov 30, 2016 at 3:26 AM, ValdikSS <notifications@github.com> wrote:

> What endless loop are you talking about? I have some problems with pkcs11
> in OpenVPN in Linux, it deadlocks sometime, so I need to run it with
> strace, that way the probability of deadlock is much less.
>
Its not an endless loop like a runaway train. Its what one would expect
from something like --auth-retry interact except that sending "cancel"
instead of "OK" to the token insertion request does not make the process
exit. Instead it just goes back to something like "insert the token now"
like a car navigation system repeatedly asking you to take a u-turn.

Again this does not happen unless the token is removed mid-way through a
password prompt.

See Trac: #740 comment 15:
https://community.openvpn.net/openvpn/ticket/740#comment:15

mattock:
Merging as agreed in today's community meeting. What remains to be fixed cannot be fixed on the OpenVPN-GUI side.

PR #95, opened 2016-11-15
Add missing WINAPI in the definition of HandleServiceIO

selvanair:
WINAPI is __stdcall in 32 bit windows (ignored in 64 bit) causing
this bug to show up in the 32 bit version only.

Also fix out-of-bounds write of ovpn_version[] in openvpn.c

Resolves Trac #758

### DIFF
--- a/openvpn.c
+++ b/openvpn.c
@@ -655,7 +655,7 @@ InitServiceIO (service_io_t *s)
  * Read-completion routine for interactive service pipe. Call with
  * err = 0, bytes = 0 to queue the first read request.
  */
-static void
+static void WINAPI
 HandleServiceIO (DWORD err, DWORD bytes, LPOVERLAPPED lpo)
 {
     service_io_t *s = (service_io_t *) lpo;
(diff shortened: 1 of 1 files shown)


selvanair:
@mattock: This fixes a real bug (my fault) that makes the 32 bit GUI unusable in the current state [(Trac 758)](https://community.openvpn.net/openvpn/ticket/758). Would be nice to get this into _beta1 if we can elicit a quick review.

cron2:
Hi,

On Tue, Nov 15, 2016 at 10:22:11AM -0800, Selva Nair wrote:
> WINAPI is __stdcall in 32 bit windows (ignored in 64 bit) causing
> this bug to show up in the 32 bit version only.
> 
> Also fix out-of-bounds write of ovpn_version[] in openvpn.c
> 
> Resolves Trac #758
> 
> Signed-off-by: Selva Nair <selva.nair@gmail.com>

I only have a rough idea what WINAPI and __stdcall is, but it's
obviously something to do with the calling convention of a callback
function - and if that mismatches, boom.  Casts are wonderful to
hide such misdeclarations :-)

It's a bug of the category "oversight, well-understood cause and fix, 
result has been tested, and the explanation matches the observed
behaviour" (physicists like that) - so, ACK.

gert

-- 
                                                           //www.muc.de/~gert/

selvanair:
Hi,

On Tue, Nov 15, 2016 at 2:56 PM, Gert Doering <notifications@github.com>
wrote:

> It's a bug of the category "oversight, well-understood cause and fix,
>

I would call it reckless oversight, but did penance spending a long time to
hunt it down... Looking back, the cast was a probably a quick fix for a
cryptic compilation error in 32 bit, left to be looked into later, but
completely forgotten.


> result has been tested, and the explanation matches the observed
> behaviour" (physicists like that) - so, ACK.
>

Thanks.

Selva

PR #96, opened 2016-11-16
travis-ci: upgraded to openssl-1.0.2i

chipitsine:


selvanair:
@mattock: Is this good to merge?

.travis.yml
```diff
@@ -39,7 +39,6 @@ script:
     - ./configure --prefix=/ --libdir=/lib --host=$CHOST --build=x86_64-pc-linux-gnu --program-prefix='' $CONFIG && make
     
 cache:
-  apt: true
```
mattock: Why is this line removed? Is it redundant?

mattock:
Besides the one line comment this looks good.

.travis.yml
chipitsine: actually, it does not work (and never supposed to work).
so, I removed it.

mattock: Ok, then I have no further concerns. Merging.

PR #98, opened 2016-11-20
Update version-info resource

selvanair:
- Add PRODUCTVERSION to the VERSIONINFO resource.
  Without this some versions of Windows (e.g., Server 2008 32bit)
  displays Product version = 0.0.0.0, File version = 11.1.0.0

- Also define the file and product versions in the stringinfo block
  as a stringified form of PACKAGE_VERSION_RESOURCE (e.g., "11.1.0.0").
  Windows versions that take the Product version from this block
  will now get the same value as older versions that use PRODUCTVERSION.


mattock:
The changes look reasonable. Merging.

PR #99, opened 2016-11-20
Make the program DPI aware

selvanair:
- Set dpi-awareness to true in the manifest (i.e., "system-dpi aware")
- Check system dpi and scale and/or position widgets and windows
  that depend on the system dpi (only components within the status
  window are affected).

Note: Declaring dpi awareness eliminates automatic rescaling of
windows that causes blurred text on high dpi monitors.
Windows 8.1 and later allow per monitor dpi setting which is
not handled here.

### DIFF
--- a/main.c
+++ b/main.c
@@ -302,6 +302,47 @@ ResumeConnections()
     }
 }
 
+/*
+ * Set scale factor of windows in pixels. Scale = 100% for dpi = 96
+ */
+static void
+dpi_setscale(UINT dpix)
+{
+    /* scale factor in percentage compared to the reference dpi of 96 */
+    if (dpix != 0)
+        o.dpi_scale = MulDiv(dpix, 100, 96);
+    else
+        o.dpi_scale = 100;
+    PrintDebug(L"DPI scale set to %u", o.dpi_scale);
+}
+
+/*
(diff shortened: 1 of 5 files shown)


chipitsine:
shall we also add "supportedOS" fields ?

https://github.com/chipitsine/openvpn-build-old/blob/master/generic/openvpn-gui-5/res/konturvpn-gui.manifest


and, probably, it makes sense to add manifest to openvpn.exe/openvpnserv.exe/openvpnserv2.exe as well ?

selvanair:
I'll add supported OS if this PR gets amended, else you can issue a PR. If you do include Vista, 7, 8, 8.1 and 10 in the list. Its useful to get the right return value for GetVErsion() etc. if we ever use it.

chipitsine:
I'm ok if you will add supported OS

selvanair:
Tested on Windows 10 with dpi =192 (scale = 200%) by @supergregg on Trac. See [Trac 772#2]( https://community.openvpn.net/openvpn/ticket/772#comment:2)

mattock:
Merging this based on discussions in [Trac 772](https://community.openvpn.net/openvpn/ticket/772). Further improvements beyond this PR are also outlined in that ticket.

PR #100, opened 2016-11-24
Copy updated copyright to language files

selvanair:
- Copyright update by commit 32c31ab942d in English
  copied to other language files.
  (except Ukranian for which a PR is pending)
- Changed version displayed in About page to the form
  "a.b.c.d"

### DIFF
--- a/res/openvpn-gui-res-de.rc
+++ b/res/openvpn-gui-res-de.rc
@@ -188,7 +188,7 @@ BEGIN
           "Copyright (C) 2004-2005 Mathias Sundman <info@openvpn.se>\n" \
           "Copyright (C) 2008-2014 Heiko Hund <heikoh@users.sf.net>\n" \
           "Copyright (C) 2012-2016 OpenVPN GUI contributors\n" \
-          "https://github.com/OpenVPN/openvpn-gui/", 0, 36, 15, 206, 42
+          "https://github.com/OpenVPN/openvpn-gui/", 0, 36, 15, 206, 50
     LTEXT "OpenVPN - Eine Applikation zum sicheren tunneln eines IP " \
           "Netzwerks über einen einzelnen TCP/UDP Port, mit Unterstützung " \
           "von SSL/TLS-basierten  Session-Authentisierung und Schlüssel- " \
@@ -196,7 +196,7 @@ BEGIN
           "Paket-Kompression.\n" \
           "\n" \
           "Copyright (C) 2002-2016 OpenVPN Technologies, Inc <info@openvpn.net>\n" \
-          "https://openvpn.net/", 0, 8, 61, 240, 64
+          "https://openvpn.net/", 0, 8, 70, 240, 64
 END
 
 /* Proxy Authentication Dialog */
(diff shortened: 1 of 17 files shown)


selvanair:
Updated with text positioning/clipping fixes for all languages. @stigok : Could you please check this displays OK?  A test executable is [here](https://github.com/selvanair/openvpn-gui/releases). It includes master + all pending PRs.

mattock:
The changes are mostly pretty mechanical and look good on the surface. @selvanair: how did you figure out correct window sizes for all the languages? Did you script it somehow?

selvanair:
I opened the executable in resedit and went through every dialog in each language to find what is clipped and how much to adjust !

stigok:
Clipping looks good to me too!

mattock:
@selvanair : I think the resedit process should be documented, so that translators themselves can make the necessary adjustments.

stigok:
I agree. Sounds like a good shortcut around going through the build process.

selvanair:
What I did was somewhat tedious (see below) and you cannot change the text by opening the exec directly. A better approach for translators would be to open the resource file in resedit and do all the edits in there. Avoid changing vertical positions of widgets as generally only horizontal positions and widths need to be changed to accommodate text. However, our resource files are not fully compliant with resedit (or with visual studio) -- making them open without error in resedit would be useful..

Anyway here is  what I did:

As I needed the translations in before checking for clipped text I had to build and exec first. Then the process is to start resedit, open a new project and point to the executable -- opening an exec instead of resource files is called Hacker mode in resedit in case you want to use the wizard. 

The left hand side pane will show all resources found, click on each dialog to open up a view of the dialog. Visually select a widget that need repositioning, scroll down in the properties (shown on left) to see the x,y, width, height values. Move the widget/text, adjust its width etc and note down the new x position and width and enter them into the resource file. Do not change y position and height unless absolutely essential (Hint: use arrow keys to move text 1 pixel at a time).  Ignore the status window as its re-formatted, buttons repositioned etc. at run time.

Some of our setting dialog tabs have inadequate heights in the resource file which doesn't really matter as all of them get automatically resized to that of the largest tab. But if translators are going to visually edit the resource files it would be good to have those heights normalized to a common value as well. I can add a commit to do that or leave for another PR?

selvanair:
@mattock are you ok with these changes? For a quick look a test exec is [here](https://github.com/selvanair/openvpn-gui/releases)

mattock:
@selvanair : yes, I'm fine with the changes. I tested _openvpn-gui-nov27.exe_ with several languages and could not spot any issues. Feel free to merge.

selvanair:
OK merging.

mattock:
Oh, one more thing. I need to update the Finnish translation before 2.4_rc1 is pushed out. At that point I can test your resedit instructions and then add them to the PR.

selvanair:
If you plan to use resedit first change the dimensions of the settings tab windows, General, Proxy and About, to 252x218 -- that's what the "Advanced" tab uses and the system rescales all of them to that size (being the largest) before display. Resedit doesn't recognize that and will show each of those tabs as an independent dialog with some text clipped unless the window size is adjusted as above. 
By the way 252x218 was chosen from the recommended sizes for property sheet dialogs.

PR #101, opened 2016-11-29
since we do not support WinXP anymore, NTDDI_WINXP --> NTDDI_VISTA

chipitsine:


configure.ac
```diff
@@ -56,7 +56,7 @@ AC_DEFINE_UNQUOTED([MAX_CONFIGS], [$MAX_CONFIGS], [Maximum number of config file
 case "$host" in
 	*-mingw*)
 		CPPFLAGS="${CPPFLAGS} -DWIN32_LEAN_AND_MEAN"
-		CPPFLAGS="${CPPFLAGS} -D_WIN32_WINNT=NTDDI_WINXP"
+		CPPFLAGS="${CPPFLAGS} -D_WIN32_WINNT=NTDDI_VISTA"
```
selvanair: Yes we should change this, but why are we using NTDDI_xxx  to set _WIN32_WINNT? It should be  _WIN32_WINNT_VISTA, isn't it? The former has 8 bytes while _WIN32_WINNT is 4 bytes. Is there some magic by which either form works?

chipitsine: I followed https://msdn.microsoft.com/en-us/library/windows/desktop/aa383745%28v=vs.85%29.aspx

what is _WIN32_WINNT_VISTA ?

selvanair:
@chipitsine The link you posted  lists what is _WIN32_WINNT_VISTA

As explained there it is  NTDDI_VERSION that is to be set to some NTDDI_xxx macro and _WIN32_WINNT to _WIN32_WINNT_xxx . As we do not need the fine grained distinction possible by NTDDI_VERSION so we should just set _WIN32_WINNT to  _WIN32_WINNT_VISTA

By the way NTDDI_VISTA = 0x06000000, _WIN32_WINNT_VISTA = 0x0600. The header files will compare the value of _WIN32_WINNT with _WIN32_WINNT_xxx and NTDDI_VERSION with NTDDI_xxx macros. The two are not interchangeable.

mattock:
So what do we do with this PR?

selvanair:
As I pointed out in the comments _WIN32_WINNT should be defined as a relevant _WIN32_WINNT_xxx macro, not _NTDDI_xxx. So waiting for @chipitsine to act.

chipitsine:
I wished to take a deeper look, but I'm out of time now. so, close this PR

PR #103, opened 2016-12-07
Add 24x24 and 20x20 versions of each icon.

aixxe:
As requested in #29.


selvanair:
The added icons look good, and will not affect the current icon loading at all as we only use the first icon in each file now and those haven't changed. Will help support different dpis when the code is improved as discussed in [Trac 772](https://community.openvpn.net/openvpn/ticket/772)

ACK.

selvanair:
@mattock I think its safe to merge this now.
Shall we also fix icon loading code to support high-dpi displays? My test branch got a positive report, but if you think we should freeze the gui until 2.4.0 is out, we can do that in 2.4.1

PR #107, opened 2016-12-08
Load icons at sizes given by DPI-dependent system metric

selvanair:
- Check system metric for large and small icon sizes and
  try to load the correct size instaed of scaling from one size.

Scaling will still happen if the required size is not available
in the icon resource. As we add more icon sizes they will
get automatically used as needed.

LoadImage scales up from next smallest size available. Revisit this
when LoadIconWithScaleDown (Vista+) becomes available in mingw.

Resolves Trac: #772 (icon scaling issue)

### DIFF
--- a/localization.c
+++ b/localization.c
@@ -236,12 +236,22 @@ ShowLocalizedMsg(const UINT stringId, ...)
     va_end(args);
 }
 
-
-HICON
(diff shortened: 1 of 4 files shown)


mattock:
We decided that this is good enough to merge in the IRC meeting today (14th Dec 2016).

PR #108, opened 2016-12-15
Update _WIN32_WINNT to _WIN32_WINNT_VISTA

selvanair:
The change will help warn against use of API calls not in vista.
The original setting of this macro to  _NTDDI_WINXP was wrong and
permitted use of any API not just those supported by WinXP. Build
is unaffected as we currently do not have any function calls not
present in Vista.


selvanair:
This is a substitute for PR 101 with the correct define for _WIN32_WINNT for vista.

selvanair:
@mattock This is a simple safe patch and will protect us against mistakenly adding API calls not supported by vista. Shall we merge it?

mattock:
@selvanair : yes, let's merge it. This just managed to drop under my radar.

PR #116, opened 2017-01-07
added Windows Vista/Win7/Win8/Win8.1/Win10 to compatibility manifest

chipitsine:


selvanair:
While this looks safe, what does it really mean to say that we fully support all versions from Vista to Windows 10. For example, would declaring support for Windows 10 mean that we handle dynamic changes in DPI which we do not (just an example) or some other feature that we may not?

chipitsine:
I meant "do not use compatibility mode", as I suspect when there's not manifest, windows might run application in some compatibility, which can lead to strange things like https://community.openvpn.net/openvpn/ticket/384

selvanair:
Are you sure384 is fixed by this change. If that's true, the commit message should mention it and Trac updated. I've been unsable to find a  goo fix for Trac 384.

chipitsine:
no, I'm not sure. I'm going to install RTL windows and play with it

selvanair:
We cant merge something based on speculation. At least we have to be sure that this change cant hurt -- are you sure of that?

chipitsine:
it does not hurt. we deploy our own installer with such manifest for 1000+ users

chipitsine:
unfortunately, it does not resolve 384

chipitsine:
I found that "winscp" is not affected by such Hebrew/English bug. will have a look at it

chipitsine:
I've got an idea, what if we will add hebrew resource to openvpn-gui ...

selvanair:
Resource file has nothing to do with input language. The former is for
translations and is chosen by the language from the settings menu. One
should be able to input Arabic or Hebrew to the password prompt without
changing the display language. And in fact one can.

The issue described in Trac #384 is that the input language automatically
changes to RTL (if any are enabled system-wdie) when the password window
comes up although the default language is set to English. Strangely enough
this does not happen for the password change window.

I think it has to do with the per thread input language setting that
Windows keeps track of, but not clear why it does it in a bizarre way and
cannot remember the previous setting.

chipitsine:
I made similar to described setup:

Win7 Rus (user locale = system locale)
three languages: En --> He (default) --> Ru
when I start openvpn-gui, in password prompt it changes language to "Ru", it seems to be "the next language after default"


we can play with LANG_DEFAULT vs LANG_NEUTRAL  ( http://stackoverflow.com/questions/15043624/difference-between-neutral-neutraldefault-and-neutralsys-default ), I'll try to add missing hebrew resource tomorrow

selvanair:
On Sun, Jan 8, 2017 at 2:50 PM, Ilya Shipitsin <notifications@github.com>
wrote:

> Win7 Rus (user locale = system locale)
> three languages: En --> He (default) --> Ru
> when I start openvpn-gui, in password prompt it changes language to "Ru",
> it seems to be "the next language after default"
>
For me with 3 languages, first one default, it changes to the third
language which is RTL. And with with 3 languages and none RTL, it stays at
the default language. As if it has a mind of its own..

selvanair:
Even if this may not fix any of the issues discussed above, looks like its good to have compatibility info in the manifest. If no one objects by tomorrow I'll merge this.

mattock:
The manifest changes look correct based on [this](https://blogs.msdn.microsoft.com/chuckw/2013/09/10/manifest-madness/) MSDN blog post. Basically the supportedOS parameters affect the Program Compatibility Assistant (PCA). No complaints from me.

Issue #105, opened 2016-12-08
Option to disable 'save password'


wimg:
For ISO27001 certification, we are not allowed to let users save their VPN passwords locally. Is there a way to remove or disable the 'save password' box upon authentication ?

selvanair:
There is no option to do this but could be achieved using an up-script [1]. Note that passwords are saved in an encrypted form using CryptProtectData and can be decrypted only by the same user. The OpenVPN core (openvpn.exe) also saves the password unless explicitly disabled using --auth-nocache.

[1] When a connection is up, the GUI runs the script named "configname_up.bat", if found in the same folder as the config file. You could have the script do

reg delete HKCU\Software\OpenVPN-GUI\configs\configname  /v auth-data /f
reg delete HKCU\Software\OpenVPN-GUI\configs\configname  /v key-data /f
exit 0 # ignore any value not found error

That will delete the stored passwords for the key and auth-user-pass, if any. Here configname is the basename of the config file (i.e., if the config file is  "myvpn.ovpn", configname would be "myvpn", and the up-script be named "myvpn_up.bat"

chipitsine:
passwords are actually stored using DPAPI (data protection api), it is not possible to steal those passwords. what is ISO 27001 requirement exactly ?

as for pure disabling, we might implement separate configure option and you (or someone) can build custom installer

selvanair:
As @chipitsine wrote, I'm also curious about the exact wording of the ISO 27001 requirement. I've heard they have requirements like not allowing access to source code etc. which makes little sense.

Anyway, here is a better alternative to my proposal to use GUI up-script: Use an openvpn up-script instead. This is better as it runs even during reconnections, and need not be batch files. To do this use a script or executable that deletes the registry keys listed in my previous reply (do not name the script as configname_up.bat to avoid the GUI running it, though that cant hurt). Then to the connection config file add

script-security 2 # needed to run user-defined scripts
up scriptname  # replace script name with path relative to the config folder or full path
auth-nocache  # openvpn will prompt passwords even for reconnections

Building with the option disabled is not very useful as a user can always install another version of the GUI and save passwords.

wimg:
I'm not aware of the exact text of the ISO certification, it's for one of our customers.

The problem is that the VPN is started by a non-admin, so using an up script apparently doesn't work there :-(

selvanair:
>The problem is that the VPN is started by a non-admin, so using an up script apparently doesn't work there :-(

[...]

PR #117, opened 2017-01-09
Add a system-wide option to disable the password save feature

selvanair:
Issue #105

- A new registry HKLM\Software\OpenVPN\disable_save_passwords
  (32 bit DWORD value) may be set to a non-zero value to
  disable password saving by users. Applies to both auth and
  private key passwords. Usernames are always saved.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -128,6 +128,10 @@ Registry Values affecting the OpenVPN GUI operation
 Parameters taken from the global registry values in
 *HKEY_LOCAL_MACHINE\\SOFTWARE\\OpenVPN\\* key
 
+(Default)
+    The installation directory of openvpn (e.g., *C:\\Program Files\\OpenVPN*).
+    This value must be present.
+
 config_dir
     The global configuration file directory. Defaults to
     *C:\\Program Files\\OpenVPN\\config*
@@ -149,6 +153,10 @@ ovpn_admin_group
     The windows group whose membership allows the user to start any configuration file
     in their profile (not just those installed by the administrator in the global
     config directory). Default: "OpenVPN Administrators".
+
+disable_save_passwords
+    Set to a nonzero value to disable the password save feature.
+    Default: 0
 
 All other OpenVPN GUI registry values are located below the
 *HKEY_CURRENT_USER\\SOFTWARE\\OpenVPN-GUI\\* key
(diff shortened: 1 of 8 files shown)


README.rst
```diff
@@ -128,6 +128,10 @@ Registry Values affecting the OpenVPN GUI operation
 Parameters taken from the global registry values in
 *HKEY_LOCAL_MACHINE\\SOFTWARE\\OpenVPN\\* key
 
+(Default)
+    The installation directory of opebvpn (e.g., *C:\\Program Files\\OpenVPN*).
```
mattock: Typo here: "opebvpn".

selvanair: Thanks.. Fixed.

selvanair:
Any thoughts on whether we should provide this option? Or is it better to disable-save-passwords if --auth-nocache is in the config which would work on a config by config basis?

selvanair:
> Since people seem to get all upset about this, I tend to give them the rope their auditors want... so a global option to disable password saving, which can only be set by "OpenVPN Administrators"...

I guess you mean "can only be set by Administrators". We do not grant any special privilege to the "OpenVPN Administrators" group at the OS level so its members cannot access a system wide registry entry. If we want the setting to be changeable by members of that group, this will have to be implemented differently (not using a registry entry in HKLM, but not sure how..) 

I think keeping this an admin only accessible setting should be good enough?

mattock:
> I think keeping this an admin only accessible setting should be good enough?

It seems enterprises want this, and enterprises have administrators. So :+1:

Deantwo:
So, this would hopefully be something that can be set at the time of installation or regedit only? I think that sounds good to me.
Would love it if a commandline argument added to the setup file for it, so we can define this with silent installations.

All the user would have to do is reinstall the client and toggle it off during setup. But if we assume the user don't have permission to install programs on their computers, then I guess this is an ok solution.

If the user does have permission to install programs on their computer, they can also install 3rd-party software that save passwords for them. And at that point there are plenty of other things that can go wrong, but at least it isn't our responsibility anymore. I hope.

cron2:
Hi,

On Mon, Jan 16, 2017 at 12:52:35AM -0800, Samuli Seppänen wrote:
> > I think keeping this an admin only accessible setting should be good enough?
> It seems enterprises want this, and enterprises have administrators. So :+1:

Right :)

gert
-- 
                                                           //www.muc.de/~gert/

selvanair:
This has been +1 by many but no full review.. @mattock: shall we lazy ACK this?
Anyway I'll first merge the two commits.

mattock:
@selvanair lazy ACK sounds good.

selvanair:
OK, merging.

PR #118, opened 2017-01-10
Check group membership without needing connection to DC

selvanair:
(updated)
See also [Trac #810](https://community.openvpn.net/openvpn/ticket/810)

~Compare the SID of the user with member SIDs in the Administrators
group and ovpn_admin_group. This avoid issues on setups where the
DC may not be reachable and allows validating users logged in with
cached credentials.~
Authorize the user if

(i) admin groups are found in the process token
(ii) or the SID of the user is a member of the admin groups

The second check is needed to support adding the user to the
ovpn_admin_group when GUI is running, as such changes in group
membership will not be reflected in the token.

Same approach is used in the proposed patch (v2) for interactive service
http://www.mail-archive.com/openvpn-devel@lists.sourceforge.net/msg13834.html

### DIFF
--- a/access.c
+++ b/access.c
@@ -42,60 +42,9 @@ extern options_t o;
 
 #define MAX_UNAME_LEN (UNLEN + DNLEN + 2) /* UNLEN, DNLEN from lmcons.h +2 for '\' and NULL */
 
-/*
- * Check whether current user is a member of specified groups
- */
-static BOOL
-CheckGroupMember(DWORD count, WCHAR *grp[])
-{
-    LOCALGROUP_USERS_INFO_0 *groups = NULL;
-    DWORD nread, nmax, err;
-    WCHAR username[MAX_UNAME_LEN];
-    DWORD size;
-    DWORD i, j;
-    BOOL ret = FALSE;
-
-    size = _countof (username);
-    /* get username in domain\user format */
-    if (!GetUserNameExW (NameSamCompatible, username, &size))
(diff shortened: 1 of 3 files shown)


selvanair:
Based on further discussions on Trac: #810, I plan to update the patch for interactive service to check nested membership as well. This PR will be updated to match the service at that point.

selvanair:
I had forgotten to update this with the version that supports nested group membership -- force pushed the updated version. Matching patch for the service is already on the devel ML. A test executable with this version was tested by a couple of users (see [Trac #810](https://community.openvpn.net/openvpn/ticket/810))

cron2:
feature-ACK, since we have this in openvpnserv now, GUI should match.  Code looks familiar though I have not done a full review

selvanair:
We have a feature-ACK but no code ACK yet -- how far is the proposed 2.4.1?

cron2:
Hi,

On Mon, Mar 06, 2017 at 03:14:09PM -0800, Selva Nair wrote:
> We have a feature-ACK but no code ACK yet -- how far is the proposed 2.4.1?

ACK, code changes look reasonable (both patches in this PR) :)

2.4.1 is supposed to happen "real soon now", but we haven't really decided
on a date.  Community meeting tomorrow, mayhaps.

gert
-- 
                                                           //www.muc.de/~gert/

selvanair:
Thanks for the review. Merging.

PR #123, opened 2017-01-18
Enable ASLR & DEP

Chocobo1:
Some related infos:
* https://github.com/rust-lang/rust/issues/16514
* https://sourceware.org/ml/binutils/2014-08/msg00177.html

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -58,6 +58,13 @@ case "$host" in
 		CPPFLAGS="${CPPFLAGS} -DWIN32_LEAN_AND_MEAN"
 		CPPFLAGS="${CPPFLAGS} -D_WIN32_WINNT=NTDDI_WINXP"
 		LDFLAGS="${LDFLAGS} -Wl,--nxcompat,--dynamicbase"
+
+		dnl older mingw doesn't support `--high-entropy-va`
+		_save_ldflags="$LDFLAGS"
+		LDFLAGS="-Wl,--high-entropy-va"
+		AC_LINK_IFELSE([AC_LANG_PROGRAM([])],[AS_VAR_SET(high_entropy_aslr,[yes])],)
+		LDFLAGS="$_save_ldflags"
+		AS_VAR_IF(high_entropy_aslr,[yes],[LDFLAGS="${LDFLAGS} -Wl,--high-entropy-va"],)
 		;;
 	*)
 		test "${enable_distonly}" = "no" && AC_MSG_ERROR([This project is supported for windows only.])
--- a/configure.ac
+++ b/configure.ac
@@ -57,6 +57,7 @@ case "$host" in
 	*-mingw*)
 		CPPFLAGS="${CPPFLAGS} -DWIN32_LEAN_AND_MEAN"
 		CPPFLAGS="${CPPFLAGS} -D_WIN32_WINNT=NTDDI_WINXP"
+		LDFLAGS="${LDFLAGS} -Wl,--nxcompat"
 		;;
 	*)
 		test "${enable_distonly}" = "no" && AC_MSG_ERROR([This project is supported for windows only.])
--- a/configure.ac
+++ b/configure.ac
@@ -57,7 +57,7 @@ case "$host" in
 	*-mingw*)
 		CPPFLAGS="${CPPFLAGS} -DWIN32_LEAN_AND_MEAN"
 		CPPFLAGS="${CPPFLAGS} -D_WIN32_WINNT=NTDDI_WINXP"
-		LDFLAGS="${LDFLAGS} -Wl,--nxcompat"
+		LDFLAGS="${LDFLAGS} -Wl,--nxcompat,--dynamicbase"
 		;;
 	*)
 		test "${enable_distonly}" = "no" && AC_MSG_ERROR([This project is supported for windows only.])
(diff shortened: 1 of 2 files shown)


mattock:
Not sure if this was a coincidence: https://github.com/OpenVPN/openvpn-build/pull/79

With that change OpenVPN GUI also gets ASLR/DEP, but only when built using openvpn-build, which is not the only option. So this PR also makes sense I think.

Chocobo1:
>Not sure if this was a coincidence: OpenVPN/openvpn-build#79

Sometimes happens ;)

---

I was hoping the CI could check my PR, because I'm not sure "Use high entropy ASLR" will work on older mingw version, mine is `gcc version 6.3.0 (Rev1, Built by MSYS2 project)`

Chocobo1:
Now I see the CI check failed, should I remove the `--high-entropy-va` part?

chipitsine:
the best would be "modify configure.ac to check whether that option is supported..."
current CI is xenial mingw

there's also "building openvpn" guide on https://community.openvpn.net/openvpn/wiki/BuildingOpenVPN-GUI

if you want to add some option which is not widely supported, it should be written there

chipitsine:
or you can make another PR to update CI settings/repo/mingw :-)

Chocobo1:
Alright, I added a test to see if the linker supports `--high-entropy-va` flag, now the CI pass.

configure.ac
```diff
@@ -57,7 +57,7 @@ case "$host" in
 	*-mingw*)
 		CPPFLAGS="${CPPFLAGS} -DWIN32_LEAN_AND_MEAN"
 		CPPFLAGS="${CPPFLAGS} -D_WIN32_WINNT=NTDDI_WINXP"
-		LDFLAGS="${LDFLAGS} -Wl,--nxcompat"
+		LDFLAGS="${LDFLAGS} -Wl,--nxcompat,--dynamicbase,--export-all-symbols"
```
selvanair: Is this hack of adding --export-all-symbols still needed in mingw? There has to be a better solution than exporting all symbols

Chocobo1: Sadly yes, I checked in vmmap & procexp.

I also notice the shipped openssl dll doesn't enable aslr.

Another thought, why this project doesn't support building in MSVC? It supports all these security features flawlessly.

cron2:
Hi,

On Wed, Jan 18, 2017 at 11:37:11AM -0800, Mike Tzou wrote:
> Another thought, why this project doesn't support building in MSVC? 

We don't regularily build with MSVC because it's much too cumbersome for
automated snapshot building - and since none of the core developers use
it, we tend to commit C code that breaks MSVC due to stupid restrictions
on their side.  We do try to fix this when we're told, though...

> It supports all these security features flawlessly.

It would be nice if it just supported C99 properly...

gert

-- 
                                                           //www.muc.de/~gert/

selvanair:
To add to that we are not MSVC compliant right now -- our resource files need a lot of changes to satisfy Windows resource compiler which is pretty "primitive" in handling line continuation using "\",  preprocessor macros in stringtable etc.

selvanair:
Oops -- pressed the wrong button!

We are not MSVC compliant right now  -- I had spent some time with R Morris and we did get it compile with a few tweaks except for the resource file which required a lot of edits so we ported only one language file. MS resource compiler is rather primitive in handling line continuation using "\" that we use extensively, preprocessor macros in stringtable etc. Nothing too complicated to change, but not enough motivation to do so..

selvanair:
I did some tests on enabling ASLR using mingw. As pointed out by @Chocobo1, using only -dynamicbase is not enough. In fact it has no effect except that the ASLR flag gets set in the executable[1].  But --export-all-symbols is not required, just exporting one symbol is enough.

Until mingw/binutils fixes this for good[2], we could add a dummy global like ```__declspec(dllexport) char aslr_workaround;``` and then -Wl,--dynamicbase will work.

Notes:
[1] Atleast on Win7 64 bit,  with or without -dynamicbase I find both heap and env block are always randomized, code and stack remains fixed unless reloc section is present. Setting the flag alone is enough for DLLs as reloc is already there.
[2] https://sourceware.org/bugzilla/show_bug.cgi?id=17321

Chocobo1:
>But --export-all-symbols is not required, just exporting one symbol is enough.

I adopted the workaround and gave credit to you in commit msg.

There is still a thing that bothers me, the image base address isn't randomized.
MS linker have a `RandomizedBaseAddress` option[1] for this, but I don't find anything on mingw side.


\[1]: https://msdn.microsoft.com/en-us/library/microsoft.visualstudio.vcprojectengine.vclinkertool.randomizedbaseaddress.aspx

selvanair:
I think setting this property in VS is the dumbed-down way of saying pass /dynamicbase to the linker. See item 5 in the msdn docs for /dynamicbase  (https://msdn.microsoft.com/en-us/library/bb384887.aspx)

Chocobo1:
Yes you're right.
And in my post above I got my mind mixed up, so it's not true.
So is this PR eligible for merging?

selvanair:
DEP appears enabled by default for 64bit programs (tested Win7, Win10), yet the flag -nxcompat may help other platforms.
For ASLR, the --high-entropy-va flag is not supported on 32bit, the check in configure takes care of that. The feature itself appears to work well on Win10 (apparently supported in Win8 as well).
Merging.

Chocobo1:
Thanks to everyone involved.

PR #128, opened 2017-01-31
Fix truncation of usage message shown when --help is used

selvanair:
- Increase the buffer size used to load the usage string to 2048
  chars (defined by USAGE_BUF_SIZE in main.h)
- Make GetGUILanguage() visible outside localization.c

Trac: #833

### DIFF
--- a/localization.c
+++ b/localization.c
@@ -63,7 +63,7 @@ FindResourceLang(PTSTR resType, PTSTR resId, LANGID langId)
 }
 
 
-static LANGID
+LANGID
 GetGUILanguage(void)
 {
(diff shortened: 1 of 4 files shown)


cron2:
Hi,

On Mon, Jan 30, 2017 at 07:38:20PM -0800, Selva Nair wrote:
> - Increase the buffer size used to load the usage string to 2048
>   chars (defined by USAGE_BUF_SIZE in main.h)
> - Make GetGUILanguage() visible outside localization.c

Looks reasonable, so ACK.

gert
-- 
                                                           //www.muc.de/~gert/

Issue #131, opened 2017-02-16
Builds from tarball fail because of a missing resource file


mattock:
Building with openvpn-build fails because _openvpn-gui-res-cs.rc_ is not packaged into the release tarballs:

```
--- snip ---
i686-w64-mingw32-windres -DHAVE_CONFIG_H -I.  -I/home/samuli/buildtest/openvpn-build/windows-nsis/tmp/image-i686/openvpn/include -D_UNICODE -DWIN32_LEAN_AND_MEAN -D_WIN32_WINNT=_WIN32_WINNT_VISTA -DWINVER=_WIN32_WINNT -i res/openvpn-gui-res.rc -o openvpn-gui-res.o
res/openvpn-gui-res.rc:41:33: fatal error: openvpn-gui-res-cs.rc: No such file or directory
 #include "openvpn-gui-res-cs.rc"
                                 ^
compilation terminated.
i686-w64-mingw32-windres: preprocessing failed.
make[1]: *** [openvpn-gui-res.o] Error 1
make[1]: Leaving directory `/home/samuli/buildtest/openvpn-build/windows-nsis/tmp/build-i686/openvpn-gui-11'
make: *** [install-strip] Error 2
FATAL: make openvpn-gui
FATAL: build i686 >&2
FATAL: build-complete >&2
```
This is because the said resource file is missing from Makefile.am. I will issue a PR to fix this.

selvanair:
Yes, this was missed in the commit that introduced the czech language resource. Sorry, before merging I only tested building the GUI in place, not from the release tarball.

mattock:
Fixed by PR https://github.com/OpenVPN/openvpn-gui/pull/132

PR #132, opened 2017-02-16
Fix builds from release tarballs

mattock:
The openvpn-gui-res-cs.rc file was not included in release tarballs, which made
tarball-based builds fail.

URL: https://github.com/OpenVPN/openvpn-gui/issues/131

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -43,6 +43,7 @@ AM_CFLAGS = -municode
 
 openvpn_gui_RESOURCES = \
 	res/openvpn-gui-res.rc \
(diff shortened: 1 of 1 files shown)


chipitsine:
well, it dies not look difficult to do "release tarball build" during travis-ci build.
I'll add this to my task list :)

mattock:
@chipitsine :+1: that would be a great test to have. This kind of breakage tends to be easy to avoid, yet very common.

mattock:
I've tested generating the release tarball, and the tarball contains the missing resource file. As the change is trivial, I'll merge this without an explicit ACK.

PR #135, opened 2017-02-23
AppVeyor support

chipitsine:
AppVeyor stores artifacts, it might be used for "test new feature" builds.
also, it is useful to build in cygwin regularly

@mattock , can you add openvpn-gui to ci.appveyor.com ?

PR #137, opened 2017-02-27
Parse ECHO directives from openvpn

selvanair:
Support the following echo commands

- "echo forget-passwords": delete passwords internally saved by the GUI
  but do not disable the password save feature. Useful when pushed
  from the server so that it gets processed after authentication. Also see
  management-notes.txt in openvpn docs.

- "echo save-passwords": enables private-key and auth-user-pass passwords
  to be saved. Will be effective at startup only if present in the config
  file. If pushed from the server, will get used for subsequent
  password prompts. Essentially this has the effect of presenting the password
  dialogs to the user with save-password checkbox selected. The user may still
  uncheck it during the dialog.

Note: echo commands are processed as and when they are received and in the order
received.

TODO: support "echo setenv", "echo disable-save-passwords"

### DIFF
--- a/main.c
+++ b/main.c
@@ -121,6 +121,7 @@ int WINAPI _tWinMain (HINSTANCE hThisInstance,
       { stop,     OnStop },
       { needok,   OnNeedOk },
       { needstr,  OnNeedStr },
+      { echo,     OnEcho },
       { 0,        NULL }
   };
   InitManagement(handler);
--- a/manage.c
+++ b/manage.c
@@ -322,6 +322,11 @@ OnManagement(SOCKET sk, LPARAM lParam)
                     if (rtmsg_handler[needstr])
                         rtmsg_handler[needstr](c, pos + 9);
                 }
+                else if (strncmp(pos, "ECHO:", 5) == 0)
+                {
+                    if (rtmsg_handler[echo])
+                        rtmsg_handler[echo](c, pos + 5);
+                }
             }
             else if (c->manage.cmd_queue)
             {
--- a/openvpn.c
+++ b/openvpn.c
@@ -108,6 +108,7 @@ OnReady(connection_t *c, UNUSED char *msg)
 {
     ManagementCommand(c, "state on", NULL, regular);
     ManagementCommand(c, "log all on", OnLogLine, combined);
+    ManagementCommand(c, "echo all on", OnEcho, combined);
 }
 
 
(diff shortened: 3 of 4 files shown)


chipitsine:
I recently created (not yet merged) PR, which I hope will simplify creating openvpn-gui.exe for testing

https://github.com/OpenVPN/openvpn-gui/pull/135

what do you think ?

selvanair:
Anyone interested in this functionality? Locally I use this + "echo setenv xxx..." for passing domain name, fileserver name, shared folder names etc to the client (useful for mapping drives in a script, for example). Once the GUI learns to parse echo commands, there are many other functionalities we could add fairly  easily (e.g., the server can suggest to 'disable-save-password').

chipitsine:
as for env variables, how are you going to escape the values ?

there's some old similar "issue":

https://github.com/OpenVPN/openvpn-gui/blob/master/TODO#L3

as for "forget-passwords" and other commands, it is fine!

selvanair:
> as for env variables, how are you going to escape the values ?

The parsing is done by openvpn.exe (client or server if pushed) and passed in through the management interface, so this is unlike other user input. In particular I want to support pushed "echo setenv ..", so I'm thinking of more-or-less the same functionality as --setenv-safe in openvpn.exe (names are mangled to avoid redefining sensitive variables etc..)

mattock:
@selvanair I think the functionality makes a lot of sense. I can't give the code a proper review, although it seems to be correct based on my minimal C-fu. That said, I think we could just merge this using lazy-ACK now after 14 days.

cron2:
Hi,

On Mon, Mar 06, 2017 at 12:03:00PM -0800, Selva Nair wrote:
> Anyone interested in this functionality? Locally I use this + "echo setenv xxx..." for passing domain name, fileserver name, shared folder names etc to the client (useful for mapping drives in a script, for example). Once the GUI learns to parse echo commands, there are many other functionalities we could add fairly  easily (e.g., the server can suggest to 'disable-save-password').

You win this month' round of "finding an interesting and well-hidden
feature" :-) - I had no idea what "echo" was about, but the management-notes
describe it well.

Sounds quite useful to me.

I'll go and have a look at the code.

gert

-- 
                                                           //www.muc.de/~gert/

selvanair:
> Sounds useful & the code looks reasonable, so ACK.
>
> Can't wait to see the more interesting echo bits :-)

Thanks for the review. 'echo setenv' and 'echo disable-save-passwords' are coming :)

PR #147, opened 2017-03-14
Reorder command line arguments check and registry operations

chipitsine:
When openvpn is not installed, HKLM\Software\OpenVPN registry is missing,
however, it should not fail on running "openvpn-gui.exe --help". Such
usage might be useful for smoke testing, for example (build openvpn-gui.exe
on AppVeyor, run it, check whether window is shown or not)

### DIFF
--- a/main.c
+++ b/main.c
@@ -174,11 +174,11 @@ int WINAPI _tWinMain (HINSTANCE hThisInstance,
         exit(1);
     }
 
-  UpdateRegistry(); /* Checks version change and update keys/values */
-
-  GetRegistryKeys();
   /* Parse command-line options */
   ProcessCommandLine(&o, GetCommandLine());
+
+  UpdateRegistry(); /* Checks version change and update keys/values */
+  GetRegistryKeys();
 
   EnsureDirExists(o.config_dir);
 


selvanair:
I see the problem, but this solution is not ideal --- most parameters read from registry are defaults and it should be possible to override them from the command line. So re-ordering the calls is not good. I see two options: (i) split up parsing of global and user registry values and move up only the first or (ii) make registry read errors non-fatal. I prefer the latter (its easier to do too) -- in fact I had made almost all registry entries optional and now the only key that must exist is HKLM\Software\OpenVPN\openvpn_path. We can define a reasonable default for that one too and continue instead of erroring out as done for other values (see registry.c)

chipitsine:
thank you for your feedback, I really missed the case when parameter defined in command line might be overwritten later from registry.

as for situation, I think it is not urgent and we can do things in comfortable manner.
I like the approach when registry value is taken if and only if it is really needed.

i.e. if user specified --exe_path - there's no need to pick value from registry, and so on...
probably we should move the check as close as possible to the usage of certain value.

selvanair:
See PR #197 that does the same without overriding command line params.

@chipitsine : could you please review #197?

chipitsine:
closing this in favour of https://github.com/OpenVPN/openvpn-gui/pull/197

PR #151, opened 2017-03-25
change "%S" to "%ls" to silence mingw warning

chipitsine:
proxy.c:602:32: warning: ISO C does not support the ‘%S’ ms_printf format [-Wformat=]

### DIFF
--- a/proxy.c
+++ b/proxy.c
@@ -599,7 +599,7 @@ OnProxy(connection_t *c, char *line)
     }
 
     char cmd[128];
-    snprintf(cmd, sizeof(cmd), "proxy %s %S %S", type, addr, port);
+    snprintf(cmd, sizeof(cmd), "proxy %s %ls %ls", type, addr, port);
     cmd[sizeof(cmd) - 1] = '\0';
     ManagementCommand(c, cmd, NULL, regular);
 


selvanair:
Looks good but we have more places where %S is used. To be consistent we should we use %ls everywhere to mean wide string and %s for narrow string but that wont work in wprintf -- MS interprets %s and %S with opposite meanings when used in printf and wprintf.. Wonder why mingw complains about %S only in this instance? What's the ISO C compliant format spec for narrow strings in wprintf that will work on windows?

chipitsine:
it is the only place where compiler comlpains

https://ci.appveyor.com/project/mattock/openvpn-gui/build/cygwin-30/job/u9wystmbg0pd1nks#L126

selvanair:
On Sat, Mar 25, 2017 at 1:46 PM, Ilya Shipitsin <notifications@github.com>
wrote:

> it is the only place where compiler comlpains


The idea is to improve the code and not just pacify a particular version of
the compiler, right? I would like to know how to correctly print wide and
narrow strings using printf and wprintf and its variants on Windows and use
it consistently everywhere. Probably replacing %S in wprintf by %hs may
work, but that's not ISO C. Full ISO C compatibility is impossible on
Windows, though..

It seems only for printf it complains about %S, not for wprintf although
the latter use is more of a MS-oddity than the other. Any idea what version
of mingw is this? It doesn't complain on my setup (GCC 4.9.1).

chipitsine:
it's cygwin, I beleive mingw-gcc-5.4.0-4

the idea is to resolve compiler warning first

selvanair:
Anyway a consistent fix is to use %ls everywhere for printing wide strings. That will leave %S in wprintf which, I think, cannot  be fixed in an ISO C compatible way (MS broke it when they decided to interpret %s and %S differently in printf and wprintf). Probably that's why the compiler doesn't warn about it.

If we ever want to support building with cygwin's runtime the remaining %S usage will need some ifdefs...

chipitsine:
building with cygwin is somewhat documented in 

https://github.com/OpenVPN/openvpn-build/blob/master/generic/build#L23-L25

also, after enabling AppVeyor on openvpn-gui repo, I'd say openvpn-gui is built regularly.
not sure it means "supported".

chipitsine:
well, I think we need to do something with that.
either mark AppVeyor as "unsupported"
or provide tests
or add "ifdef"s

selvanair:
As nothing better has come up, let's merge this and get rid of the appveyor warning.

PR #165, opened 2017-05-19
use en dashes where they should be used

nilswende:


selvanair:
The commit message (hyphen to en dash) is misleading here -- this includes a lot of newly translated strings as well and needs review. @cron2 ?

PR #173, opened 2017-06-15
Set focus to password field when username is filled

ffes:
When the username is filled automatically, set the focus to the password field. This way you can enter the password immediately without having the press TAB (or even worst users using the mouse to click on the password field).

### DIFF
--- a/openvpn.c
+++ b/openvpn.c
@@ -289,7 +289,10 @@ UserAuthDialogFunc(HWND hwndDlg, UINT msg, WPARAM wParam, LPARAM lParam)
 
         }
         if (RecallUsername(param->c->config_name, username))
+        {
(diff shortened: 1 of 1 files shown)


selvanair:
This is a useful thing to do, but is it effective? My tests show that when the auth dialog pops up, the focus goes to the task bar where the icon starts to blink indicating a window needs attention. The user has to then click on the auth dialog to be able input anything. This is not a fault of this patch but the current behaviour. 

I think this is caused by the way dialogs are created by  LocalizedDialogBoxParam() with NULL passed as the parent window. Shall we change that to use o.hWnd to have the dialog automatically get focus? That could make the patch really useful.

selvanair:
Correction: I cant anymore reproduce the behaviour I pointed out and this works as expected. Also this is the right thing to do. So ACK. We can address the parent window = NULL vs o.hWnd independently, if necessary.

PR #174, opened 2017-06-15
Fix typo "diffrent"

stvhwrd:


selvanair:
Thanks for the patch. ACK.

PR #176, opened 2017-06-25
harden build script

chipitsine:
when playing with openssl-1.1.0 build, I noticed that build fails on very late stage

https://travis-ci.org/chipitsine/openvpn-gui/builds/246756272

so, adding "set -e" should make it really better

### DIFF
--- a/.travis-build-openssl.sh
+++ b/.travis-build-openssl.sh
@@ -1,5 +1,6 @@
 #!/bin/bash 
 
+set -e
 
 if [ ! -f download-cache/openssl-${OPENSSL_VERSION}.tar.gz ]; then wget -P download-cache/ https://www.openssl.org/source/openssl-${OPENSSL_VERSION}.tar.gz; fi
 


mattock:
I'm wondering if this could have adverse effects. From [Bash man-page](https://linux.die.net/man/1/bash):

> [set] -e
> 
> Exit immediately if a pipeline (which may consist of a single simple command), a subshell command enclosed in parentheses, or one of the commands executed as part of a command list enclosed by braces (see SHELL GRAMMAR above) exits with a non-zero status. The shell does not exit if the command that fails is part of the command list immediately following a while or until keyword, part of the test following the if or elif reserved words, part of any command executed in a && or â”‚â”‚ list except the command following the final && or â”‚â”‚, any command in a pipeline but the last, or if the command's return value is being inverted with !. A trap on ERR, if set, is executed before the shell exits. This option applies to the shell environment and each subshell environment separately (see COMMAND EXECUTION ENVIRONMENT above), and may cause subshells to exit before executing all the commands in the subshell.

The point being that the rules for the failure condition are rather elaborate. Can we be fairly sure that "set -e" does not break the builds in interesting ways?

chipitsine:
we have similar hardening for openvpn itself

https://github.com/OpenVPN/openvpn/blob/master/.travis/build-deps.sh#L2

what am I afraid of is currently we do not fail, because "export" (the last statement) never fails


```
#!/bin/bash 


if [ ! -f download-cache/openssl-${OPENSSL_VERSION}.tar.gz ]; then wget -P download-cache/ https://www.openssl.org/source/openssl-${OPENSSL_VERSION}.tar.gz; fi

mkdir image
tar zxf download-cache/openssl-${OPENSSL_VERSION}.tar.gz && cd openssl-${OPENSSL_VERSION}
if [ "$CHOST" == "i686-w64-mingw32" ]; then export target=mingw; fi
if [ "$CHOST" == "x86_64-w64-mingw32" ]; then export target=mingw64; fi
./Configure --prefix=/ --cross-compile-prefix=${CHOST}- shared $target no-multilib no-capieng --openssldir=/etc/ssl --libdir=/lib -static-libgcc >build.log 2>&1 || (cat build.log && exit 1)
make install INSTALL_PREFIX="${HOME}/image" INSTALLTOP="/" MANDIR="/tmp" >build.log 2>&1 || (cat build.log && exit 1)
cd .. 

export OPENSSL_CRYPTO_CFLAGS="-I${HOME}/image/include"
export OPENSSL_CRYPTO_LIBS="-L${HOME}/image/lib -lcrypto"
```



the alternative is "check every exit status", but I do not really want to.
I think "set -e" would be a good thing here

cron2:
Hi,

On Mon, Jun 26, 2017 at 06:04:05AM -0700, Ilya Shipitsin wrote:
> we have similar hardening for openvpn itself
> 
> https://github.com/OpenVPN/openvpn/blob/master/.travis/build-deps.sh#L2
> 
> what am I afraid of is currently we do not fail, because "export" (the last statement) never fails

Why do we have export as the last statements?  They won't do anything
useful, as the symbol is un-exported as soon as the script ends (no
export "up to parent")...?

gert
-- 
                                                           //www.muc.de/~gert/

chipitsine:
script is included via "dot"

https://github.com/OpenVPN/openvpn-gui/blob/master/.travis.yml#L37

so, export is visible to parent.
I agree, it might be not intuitive. PRs are welcome :)

chipitsine:
closing this in favour of https://github.com/OpenVPN/openvpn-gui/pull/205

PR #181, opened 2017-08-14
Add instruction how to build using MSYS2

ffes:
Added an instruction how to build openvpn-gui using MSYS2

### DIFF
--- a/BUILD.rst
+++ b/BUILD.rst
@@ -57,13 +57,13 @@ One-time preperation
 
 Install MSYS2. Instructions and prerequisites can be found on the official website: https://msys2.github.io/
 
-Once installed use the `mingw64.exe` provided by MSYS2.
+Once installed use the ``mingw64.exe`` provided by MSYS2.
 
 Update the base MSYS2 system until no further updates are available using:
 
 .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


mattock:
Could you mention that the instruction are for Arch Linux? Besides that looks good.

ffes:
No, these **are** instruction to build using MSYS2. AFAIK MSYS2 has nothing to do with Arch Linux.

If you are not sure what MSYS2 is, look at their site (as mentioned in this PR): https://msys2.github.io/ which redirects to http://www.msys2.org/ since a (short?) while.

chipitsine:
@ffes , I think we could interpet @mattock question as "almost nobody knows what msys2 is actually".
and some documentation would help here

I'm going to add msys2 build to app veyor soon (or you can). app veyor build script is documentation itself

there's question about windows builds: https://github.com/OpenVPN/openvpn-gui/pull/151 (see Selva Nair reply). building on windows differ in terms of string formatting. we should do something with that

mattock:
@ffes: ah, I see, my bad. I did not realize pacman was in use outside of Arch Linux. I will merge this as-is, then.

Issue #185, opened 2017-09-18
Traffic goes out from VPN when disconnect happens


gjf:
Sometimes VPN server connection is not stable enough so disconnect happens.
In this case the client shows the message to reconnect providing already stored login and password and waits for the user presses "connect".
In this time when user does not see this message all traffic passes unencrypted outside VPN.

I believe the option (automatically reconnect" would solve this problem, but it is absent.

chipitsine:
I like the way it works in SoftEtherVPN.
it reconnects (and connects) automatically unless user interrupts the connection process

gjf:
Exactly. And also it would be good if the client has a possibility to block all other traffic rather than thru VPN.

selvanair:
> Sometimes VPN server connection is not stable enough so disconnect happens.
In this case the client shows the message to reconnect providing already stored login and password and waits for the user presses "connect".

This should not happen. In case of network glitches openvpn core will reconnect by itself without any prompts unless you have some directives in the config (like --auth-nocache) that interferes with this process.

gjf:
So this feature cannot be done without caching password in memory? OK, will try.

selvanair:
If you want the GUI to automatically use the save password without showing a dialog that is currently not  supported. The assumption is that such prompts are repeated only if --auth-nocache is in use and in such cases the user probably wants to see the prompt.

But it may be a good idea to optionally suppress that prompt: say, if silent_connection is on.

gjf:
Also it would be very nice to be sure there is no possibility to forward traffic through VPN and no way other.
For instance: automatic starting of VPN connection immediately after automatic start of client, blocking other routes etc.

selvanair:
Requirements lik which config, if any, to start automatically or whether to send all traffic through VPN etc vary from user to user, case to case.
- To redirect all traffic through vpn use --redirect-gateway in the config
- To start a connection automatically when the GUI is launched, use --connect <config-file> command line option to the GUI. This may be set on the GUI shortcut.

As for firewalling traffic see also discussion here: #153

selvanair:
PR #186 makes the dialog auto submit after a delay unless interrupted by the user. Should be in next release. This would take care of automatic reconnection.

Closing. Reopen if necessary.

PR #186, opened 2017-09-21
Use saved password with no prompt if silent_connection is on

selvanair:
The behaviour is unchanged if silent_connection is not enabled either on command line or
in the GUI settings.
See also #185 


chipitsine:
Can we merge it?

selvanair:
@chipitsine: There has been some user feedback suggesting this should not be coupled to 'silent_connection' and a separate checkbox that allows to skip this dialog would be more friendly (see #193). E.g.,  a check box on the password dialog window that says don't show this dialog again.

If this is useful as is, I'm open to lazy ACK it. It cant hurt anyway.

chipitsine:
@selvanair , personally I like a bit different way of "silent" reconnection.

I observed it on SoftEtherVPN. If password is saved, so timer is on about 5 sec which let you to interrupt reconnect if you want. If you do not interrupt reconnection, it reconnects with no user interaction.


your way is also good, I think, so, I'd merge it

chipitsine:
almost-silent


something like that (it was another kind of reconnection, due to network error, but idea is the same)

chipitsine:
I just meant UI, not reconnection.
same UI with moving green line can be used for "silent" connections. i.e. it connect in several second unless user wishes to interrupt it

selvanair:
@chipitsine Based on your remarks, I propose the "auto-submit saved credentials if silent" behaviour of this PR coupled with "always submit within a few seconds unless the user interacts". A patch that implements this is in a private branch [here](https://github.com/selvanair/openvpn-gui/tree/no-auth-prompt-v2)

A progress bar looks odd in such a dialog, so it just displays a message saying "Continuing with saved data in 5 seconds.." User can interrupt and submit right away (click OK), or edit any of the fields in which case the scheduled submit is cancelled. 

If this looks promising, please test.

chipitsine:
@selvanair - I'm afraid, I will not provide you any relevant feedback, I will ask our users instead...

as for me, it is great, however, couple of notices

1. "Continuing with prefilled..." should not be red. red means an error, i.e. "pay attention to me".

2. Some interaction with user should be nice, if not a green line, so maybe countdown timer ? 5 sec ... 4 sec ... 3 sec... ? If something is going on, user will treat it as "ok, it is going on, no interaction from me required"

chipitsine:
@selvanair , can you, please enable App Veyor ? it keeps artifacts across builds, that simplifies "here's brand new openvpn-gui, please test" scenario.

I merged your commits https://ci.appveyor.com/project/chipitsine/openvpn-gui-nhep8

chipitsine:
also, I would change

"Continuing with prefilled data in %u seconds..."

with

"Connect automatically in %u sec."

selvanair:
@chipitsine : Thanks for the feedback. Will address all in the next version.

selvanair:
Updated:
When saved username and password are available:
(i) If silent_connection is on bypass auth-user-pass dialog
(ii) Else auto submit after a brief delay unless the user interrupts
(iii) Show a message "Connecting automatically after <time-remaining> seconds"

chipitsine:
I tested it. much better UX, let's merge it

PR #187, opened 2017-09-22
Highlight warning and error messages in status window

selvanair:
Change text colour of log lines flagged as Warning, Nonfatal error and Fatal error.
Colours defined by clr_warning and clr_error introduced in a previous patch are
used. Both of these are currently set as colour "red".

### DIFF
--- a/openvpn.c
+++ b/openvpn.c
@@ -145,6 +145,7 @@ OnLogLine(connection_t *c, char *line)
     message = strchr(flags, ',') + 1;
     if (message - 1 == NULL)
         return;
+    size_t flag_size = message - flags - 1; /* message is always > flags */
 
     /* Remove lines from log window if it is getting full */
     if (SendMessage(logWnd, EM_GETLINECOUNT, 0, 0) > MAX_LOG_LINES)
@@ -158,8 +159,28 @@ OnLogLine(connection_t *c, char *line)
     datetime = _tctime(&timestamp);
     datetime[24] = _T(' ');
 
-    /* Append line to log window */
+    /* deselect current selection, if any */
     SendMessage(logWnd, EM_SETSEL, (WPARAM) -1, (LPARAM) -1);
+
+    /* change text color if Warning or Error */
+    COLORREF text_clr = 0;
+
+    if (memchr(flags, 'N', flag_size) || memchr(flags, 'F', flag_size))
+        text_clr = o.clr_error;
+    else if (memchr(flags, 'W', flag_size))
+        text_clr = o.clr_warning;
+
+    if (text_clr != 0)
+    {
+        CHARFORMAT cfm = { .cbSize = sizeof(CHARFORMAT),
+                           .dwMask = CFM_COLOR|CFM_BOLD,
+                           .dwEffects = 0,
+                           .crTextColor = text_clr,
+                         };
+        SendMessage(logWnd, EM_SETCHARFORMAT, SCF_SELECTION, (LPARAM) &cfm);
+    }
+
+    /* Append line to log window */
     SendMessage(logWnd, EM_REPLACESEL, FALSE, (LPARAM) datetime);
     SendMessage(logWnd, EM_SETTEXTEX, (WPARAM) &ste, (LPARAM) message);
     SendMessage(logWnd, EM_REPLACESEL, FALSE, (LPARAM) _T("\n"));


cron2:
Hi,

On Thu, Sep 21, 2017 at 08:39:30PM -0700, Selva Nair wrote:
> Change text colour of log lines flagged as Warning, Nonfatal error and Fatal error.
> Colours defined by clr_warning and clr_error introduced in a previous patch are
> used. Both of these are currently set as colour "red".
> 
> Signed-off-by: Selva Nair <selva.nair@gmail.com>
> 
>   https://github.com/OpenVPN/openvpn-gui/pull/187

This is definitely a good thing.

Feature-ACK :-)

Code-wise, I'm wondering - where is text_clr reset to "normal" after
an error message?  Or is this automatic?  (I see a "if it's not 0, 
send EM_SETCHARFORMAT" but no "reset to 0 if we had an error").

Also, I do not see where OnLogLine() is called ("callback'ed"), so I'm
not sure if we can rely on "line" being writeable.  But since you've
commented it explicitely, I think I just believe this works :-)

So, code-ACK if this has been taken care of.

gert

-- 
                                                           //www.muc.de/~gert/

selvanair:
Thanks for the review.. 

For those who are wondering how this looks -- [here](https://imagebin.ca/v/3bHOlOEhPFkP) is a screenshot

>Code-wise, I'm wondering - where is text_clr reset to "normal" after
>an error message?  Or is this automatic?  (I see a "if it's not 0,
>send EM_SETCHARFORMAT" but no "reset to 0 if we had an error").

EM_SETCHARFORMAT is sent with flag = SCF_SELECTION which means it applies to the current selection if any (we have none) or to the insertion point until that point changes. We always send a new line at the end which changes the insertion point and that implicitly resets the colour back to default. This could be made more explicit, though..

>Also, I do not see where OnLogLine() is called ("callback'ed"), so I'm
>not sure if we can rely on "line" being writeable.  But since you've
>commented it explicitely, I think I just believe this works :-)

The callback is referred to as `rtmsg_handler[logline]()` in manage.c. On return, the caller moves on to the next line so its safe. Yet I would like to avoid modifying line here but a strdup of every log line looked like an overkill..

Just realized we can use message-flag as the length and then memchr instead of strchr to search for the character. I'll push a commit to do that.

cron2:
Hi,

On Fri, Sep 22, 2017 at 03:44:25PM +0000, Selva Nair wrote:
> For those who are wondering how this looks -- [here](https://imagebin.ca/v/3bHOlOEhPFkP) is a screenshot

I used my imagination, and "visually marking error/warnings in lots of
informative text" sounded good enough :-)

> >Code-wise, I'm wondering - where is text_clr reset to "normal" after
> >an error message?  Or is this automatic?  (I see a "if it's not 0,
> >send EM_SETCHARFORMAT" but no "reset to 0 if we had an error").
> 
> EM_SETCHARFORMAT is sent with flag = SCF_SELECTION which means
> it applies to the current selection if any (we have none) or to the
> insertion point until that point changes. We always send a new line
> at the end which changes the insertion point and that implicitly
> resets the colour back to default. This could be made more explicit,
> though..

Thanks for the explanation.  Good enough :-)

> >Also, I do not see where OnLogLine() is called ("callback'ed"), so I'm
> >not sure if we can rely on "line" being writeable.  But since you've
> >commented it explicitely, I think I just believe this works :-)
> 
> The callback is referred to as `rtmsg_handler[logline]()` in manage.c. On return, the caller moves on to the next line so its safe. Yet I would like to avoid modifying line here but a strdup of every log line looked like an overkill..
> 
> Just realized we can use message-flag as the length and then memchr instead of strchr to search for the character. I'll push a commit to do that.

This would be nicer, neither strdup()'ing nor modifying the caller's 
string.  Thanks.

gert

-- 
                                                           //www.muc.de/~gert/

selvanair:
>This would be nicer, neither strdup()'ing nor modifying the caller's
>string.  Thanks.
> gert

Pushed a commit to do so. (commit 44aa5ac5b445195c996dbeaa91bd40bfa0f08a4c)

cron2:
Hi,

On Sat, Sep 23, 2017 at 12:41:21AM +0000, Selva Nair wrote:
> >This would be nicer, neither strdup()'ing nor modifying the caller's
> >string.  Thanks.
> > gert
> 
> Pushed a commit to do so. (commit 44aa5ac5b445195c996dbeaa91bd40bfa0f08a4c)

Looks good to me.  ACK.

gert

-- 
                                                           //www.muc.de/~gert/

PR #192, opened 2017-10-11
Update german translation

MaxXor:
This PR completes the german translation and fixes some minor things.

### DIFF
--- a/res/openvpn-gui-res-de.rc
+++ b/res/openvpn-gui-res-de.rc
@@ -103,7 +103,7 @@ END
 /* Change Passphrase Dialog */
 ID_DLG_CHGPASS DIALOG 6, 18, 193, 82
 STYLE WS_POPUP | WS_VISIBLE | WS_CAPTION | WS_SYSMENU | DS_CENTER
-CAPTION "OpenVPN – Ändern der Passphrase"
+CAPTION "OpenVPN – Passwort ändern"
 FONT 8, "Microsoft Sans Serif"
 LANGUAGE LANG_GERMAN, SUBLANG_DEFAULT
 BEGIN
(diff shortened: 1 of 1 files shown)


MaxXor:
Can anyone review?

mattock:
@MaxXor sorry this slipped through. I asked @cron2 and @d12fk to review on IRC.

mattock:
It seems we won't be getting any external review. I had a quick look and based on my 2 years of German studies from 25 years ago the PR looks reasonable 😄. In any case I'll merge this based on our lazy-ACK policy.

cron2:
Hi,

On Tue, Jan 23, 2018 at 12:04:03PM +0000, Samuli Seppänen wrote:
> It seems we won't be getting any external review. I had a quick look and based on my 2 years of German studies from 25 years ago the PR looks reasonable ????. In any case I'll merge this based on our lazy-ACK policy. 

OK, sorry for slacking.

I agree to all changes but the *first* one - if that's really the
"change passphrase" dialog (as the comment suggests), calling it a
"Passwort" is confusing at best (because that would be a "password",
like in "username + password" while the passphrase is for the encrypted
key).


So *if* that's the passphrase, I'd change the caption to

 "OpenVPN - Public Key Passphrase ändern"

... and if not, the comment needs to be adjusted.

Neither should stop this from being merged, all the other changes are
a big improvement.

gert


-- 

mattock:
@cron2 excellent and thanks for the proper review! @MaxXor I merged this one already, but the fix proposed by @cron2  can be in a separate PR which, if merged by Thursday, will end up in OpenVPN 2.4.5 installers.

MaxXor:
@mattock Thank you!
@cron2 I agree and will fix this in a separate PR. Thanks.

PR #194, opened 2017-11-18
Validate user input

selvanair:
A very basic validation of user input (username/password, private key passphrase) to avoid errors in communication over the management socket. The patch only disallows embedded '\n', but more exclusions could be easily added as necessary.

Invalid input shows an error-tip as in image below:
image

Commits:
1. An error in parsing of process ID received from the service is corrected
2. Input validation
3. Define a macro used in previous commit missing in some mingw versions --- could be avoided by updating CI builds to use a newer mingw (say ubuntu 16.04)

~~Tested by: @Handle (Trac #958)~~
Tested by: @H4ndl3  (Trac #958)

### DIFF
--- a/manage.c
+++ b/manage.c
@@ -340,6 +340,12 @@ OnManagement(SOCKET sk, LPARAM lParam)
                 }
                 else if (strncmp(line, "ERROR:", 6) == 0)
                 {
+                    /* Response sent to management is not processed. Log an error in status window  */
+                    char buf[256];
+                    _snprintf_0(buf, "%lld,N,Previous command sent to management failed: %s",
+                                (long long)time(NULL), line)
+                    rtmsg_handler[log](c, buf);
+
                     if (cmd->handler)
                         cmd->handler(c, NULL);
                     UnqueueCommand(c);
--- a/misc.c
+++ b/misc.c
@@ -453,3 +453,12 @@ Widen(const char *utf8)
 
     return wstr;
 }
+
+/* Return false if input contains any characters in exclude */
+BOOL
+validate_input(const WCHAR *input, const WCHAR *exclude)
+{
+    if (!exclude)
+        exclude = L"\n";
+    return (wcspbrk(input, exclude) == NULL);
+}
--- a/misc.h
+++ b/misc.h
@@ -39,4 +39,5 @@ BOOL CheckFileAccess (const TCHAR *path, int access);
 BOOL Base64Encode(const char *input, int input_len, char **output);
 int Base64Decode(const char *input, char **output);
 WCHAR *Widen(const char *utf8);
+BOOL validate_input(const WCHAR *input, const WCHAR *exclude);
 #endif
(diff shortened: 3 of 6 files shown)


H4ndl3:
Not that it would be that important, but I'm the one who reported the bug and tested this patch - someone was just faster than me in taking that nickname.

selvanair:
Thanks @H4ndl3 : commit message updated.

chipitsine:
common mistake also is CAPS LOCK. if we can warn a user when it is pressed, it would be nice.
also, less common is punto switcher (nasty russian stuff which automatycally switches between keyboard layouts, however, we can add an application to its white list)

selvanair:
@chipitsine  : the validation done here happens when the user tries to submit the input (presses OK) and not during key-press events. CAPS lock detection and warning would need a different logic. Let's handle that in a separate patch.

chipitsine:
@selvanair , it seems that "wrong character detection" would be nice during input as well.

but, I'd leave it up to you

selvanair:
@chipitsine : I think checks during input and during submission are both useful and could play slightly different roles. In case of masked fields like passwords and, generally, to alert errors in cut-and-paste text, a final validation is useful and that's what this patch does. Only absolutely unacceptable input is warned against here as this stops the submission from going through.

If/when we add something like cap lock warning, other dynamic input checks could be added to it.

chipitsine:
if considering copy-paste errors, I'd consider "show me" button.
for example, when people use bitwarden vault, it is ok to copy password to buffer, but when google docs are used, any random character might be included either before of after the actual password

selvanair:
This all started with Trac #958 where the error was triggered by an embedded '\n' in copy-pasted text. Show me will not help there. Currently we do not disallow any "show-able" characters; if that changes, a show me may help. At the same time I do not want this to be a knee-jerk reaction and make a "look for '\n' patch" and instead want to  use the opportunity to catch any other plausible input errors as well.

Hence the question about what other characters could cause issues in password or username.

mattock:
@selvanair : can we drop commit "Add a define for TTI_ERROR_LARGE missing in some mingw versions" now that Travis-CI uses Ubuntu 16.04 version of mingw-w64? If I recall correctly we're no longer interested in supporting the Ubuntu 14.04 version as it requires a patches mingw-w64 version anyways.

selvanair:
@mattock : Thanks for reminding to purge the extra commit. And it builds now without the define.

cron2:
HI,

On Sat, Nov 18, 2017 at 06:43:51PM +0000, Selva Nair wrote:
> -- Commit Summary --
> 
>   * Correct parsing of the process ID returned by interatcive service
>   * Check for invalid characters in user inputs
>   * Add a define for TTI_ERROR_LARGE missing in some mingw versions

Looks good to me (without the TTI_ERROR_LARGE bit that already got
removed).  So, feature- and code-ACK.

gert
-- 

selvanair:
@cron2 : Thanks for the review. Merging.

PR #195, opened 2017-11-19
switch travis-ci mingw to ubuntu 16.04

chipitsine:


selvanair:
As we have already done this with openvpn-build and we've been working around defines not in older mingw, its useful & timely to move to 16.04/xenial. 

@chipitsine : Is appveyor already using a matching mingw version?

But I have tough time testing this: my travis builds fail with some apt repo errors -- see, e.g.: https://travis-ci.org/selvanair/openvpn-gui/jobs/304400351 Any idea what is wrong? mingw-w64 does get updated and a missing define I was after are resolved, so that part is good.

chipitsine:
probably we should move it under "before_script" section.
if it fails, build is "errored" and it is not passing to "script"

I wonder why it is passing to "script" after failed "install" section

chipitsine:
I see "i686-w64-mingw32-gcc (GCC) 6.4.0" on AppVeyor

selvanair:
@chipitsine : See here: https://travis-ci.org/selvanair/openvpn-gui/jobs/304420575#L1497

mingw does get updated, but on adding a gcc version check still shows:
x86_64-w64-mingw32-gcc --version
x86_64-w64-mingw32-gcc (GCC) 4.8.2  <-- should be 5.3.1

And, as expected, compilation fails to detect some defines that should be there in the updated mingw.

How does it work in openvpn-build, but not here? I tried to reproduce the same steps on a google compute instance (trusty with mingw from xenial) and it appears to work.

chipitsine:
seem to be my fault. thank you for catching that

selvanair:
OK, that has done the trick. Thanks. My test is happy at last.... 
https://travis-ci.org/selvanair/openvpn-gui/builds/304436472
Going by this, we may want to move nsis installation out of the defaults in openvpn-build so we get a newer version that comes with xenial.

If @mattock does not object this may be merged.

mattock:
@selvanair : looks like the commits have been squashed. Merging.

PR #196, opened 2017-11-19
AppVeyor: use proper way to resolve project path

chipitsine:
APPVEYOR_PROJECT_SLUG point to project path properly, when
repo is cloned, deleted and cloned again


mattock:
Looks reasonable based on Appveyor's [environment variable documentation](https://www.appveyor.com/docs/environment-variables). I guess this worked before because project name and project slug happened to match. Merging.

PR #197, opened 2017-12-03
Add a default for openvpnpath

selvanair:
Allow the GUI to run without any registry keys present:
the location of openvpn installation defaults to
C:\Program Files\OpenVPN if not found in registry.

Useful for testing builds or to display --help etc.

See also: PR #147

### DIFF
--- a/registry.c
+++ b/registry.c
@@ -93,45 +93,47 @@ GetGlobalRegistryKeys()
       != ERROR_SUCCESS)
     {
       /* registry key not found */
+      regkey = NULL;
       ShowLocalizedMsg(IDS_ERR_OPEN_REGISTRY);
-      return(false);
     }
-  if (!GetRegistryValue(regkey, _T(""), openvpn_path, _countof(openvpn_path)))
+  if (!regkey || !GetRegistryValue(regkey, _T(""), openvpn_path, _countof(openvpn_path)))
     {
       /* error reading registry value */
-      ShowLocalizedMsg(IDS_ERR_READING_REGISTRY);
-      RegCloseKey(regkey);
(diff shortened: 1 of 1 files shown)


registry.c
```diff
@@ -93,45 +93,47 @@ GetGlobalRegistryKeys()
...
-      return(false);
     }
-  if (!GetRegistryValue(regkey, _T(""), openvpn_path, _countof(openvpn_path)))
+  if (!regkey || !GetRegistryValue(regkey, _T(""), openvpn_path, _countof(openvpn_path)))
     {
       /* error reading registry value */
-      ShowLocalizedMsg(IDS_ERR_READING_REGISTRY);
-      RegCloseKey(regkey);
-      return(false);
+      if (regkey)
```
mattock: Just verifying the logic here: how do we get inside this "if" block? The conditions of the parent "if" block (line 99) seem contradictory to this one and nothing is explictly setting a value for "regkey". Am I missing something?

selvanair: @mattock: The block is entered under two conditions: either (i) regkey is NULL or (ii) GetRegistryValue() failed. (i) happens if the key is not present (ii) if its default value is not set. In case (ii) we show the error, in case (i) we have already shown an error saying key not found...

selvanair:
@chipitsine : Thanks for the approval.

registry.c
mattock: Ah yes, I see it now. The "regkey" parameter can indeed be defined at this point if GetRegistryValue failed on line 99. I guess I did not realize that "||" was actually a logical OR. Sorry about the noise.

PR #199, opened 2017-12-10
Activate urls in status log and open them on left button click

selvanair:
To test use a client config with no server cert verification to get a log
message with an embedded URL (see screenshot below).
 
- Only http/https urls are opened and shell open method
  is used. This should launch the user's default browser.

- Single click opens the url on left mouse button up event.

Currently few log messages contain urls. Hopefully this feature
will encourage more embedded urls in logs.

capture

### DIFF
--- a/misc.c
+++ b/misc.c
@@ -29,6 +29,7 @@
 #include <string.h>
 #include <stdlib.h>
 #include <malloc.h>
+#include <shellapi.h>
 
 #include "options.h"
 #include "manage.h"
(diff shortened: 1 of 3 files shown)


chipitsine:
@selvanair , thank you for keeping adding weird stuff to openvpn-gui :)

as for UX, I have couple of thoughts (maybe, you will want to add something like that in further development)

1) we definetly need a better way of "give me logs". I kind of fed up with telling people "open explorer ... browse to c:\users\...\Openvpn\logs ... see files there ? send them to me, please). something like "zip logs and ... ?" would be really nice

2) that particular http link is useless. it will lead user to wrong direction. I think we should point users to right direction (I do not know what direction is right for "TLS key negotionation failed to occur within 60 sec...")

selvanair:
@chipitsine : If there are urls in logs, its useful to be able to click on them. Having proper urls depend on what we write to logs and that is obviously beyond this patch. By the way you are reading the sample log image wrong -- the link that appears in there is about server certificate verification and does indeed point to a relevant page. For the TLS renegotiation error was deliberately caused to generate some red lines.

As for getting logs from user, there is a "view log" menu which opens the log. The user can copy and paste the text from there? Or save to a file and send it off. A link for reporting problems could be useful, but report to who?

chipitsine:
@selvanair - I assume that user was provided with working config, so TLS authentication timeout means something like "no connection", "user cert is revoked", "additional TLS script check failed"

nothing of that can be found on the http link provided

as for "view logs" - I did not think of using that button. Let me try and see what will happen

"but report to who?" - for example, we observed an issue recently, from server logs everything is ok. however, from client point of view something failed (it was somewhat with virtualbox ethernet adapter), so, I'd like to get logs if client connection failed. something like "connection failed, here are logs ... let send them..." or http POST to predefided location or something like that.

selvanair:
@chipitsine : Let me try again: the only place the pasted log window shows an http link is for the warning "Server certificate verification not enabled." The link that comes with that line is valid and relevant. Even if the link were in error, that's unrelated to the patch.

As for TLS verification error, the log currently does not provide any link for it just as for a numerous other errors as well. The patch is about displaying urls in logs in a way user can click on  them.

To have more and more errors accompanied by helpful urls, please submit patches to openvpn.

View logs menu is the way all users view the logs, or so I thought.

selvanair:
This is better handled when url parsing is included in planned "echo msg" feature. So closing this PR.

PR #200, opened 2017-12-22
Support for 'setenv name var' using echo

selvanair:
- Implement connection specific env variables. These are merged
  with the process environment strings and passed to scripts.

- To set an env variable, use 'echo setenv name value' in the config
  or push from the server. This will set "OPENVPN_name=value" in
  the connections's env set. Note that "name" is mangled as
  "OPENVPN_name" to avoid servers overwriting sensitive variables
  such as PATH. Names are set in the order received and same name
  overwrites any previously set value.

- Environment variable names are allowed to contain only alpha numeric
  characters and underscore as in openvpn.exe. But, unlike openvpn.exe,
  invalid names are ignored, not sanitized.

v2 changes (Dec 16, 2017):

- If value is missing, the directive is interpreted as a delete command
  and the env var with matching name in the connection's env set is removed.

- Windows needs env block to be ordered:
  While merging connection specific env vars with process env block, order
  the entries 'alphabetically' (locale independent, case insensitive
  unicode ordinal order). In case of duplicates, the value in connection
  env set replaces the one in process env.


selvanair:
Anyone interested in pushing env vars from server to GUI scripts, please test. 32 and 64 bit executables may be found in the appveyor builds (link to 64 bit: https://ci.appveyor.com/project/mattock/openvpn-gui/build/cygwin-113/job/c343mix656kibyvk/artifacts)

Thanks.

mattock:
@selvanair I think this is another good candidate for a lazy-ACK after the merge conflicts have been resolved.

selvanair:
Agreed, and conflicts resolved.

PR #202, opened 2018-01-04
Do not auto submit username/password after an auth failure

selvanair:
Without this, after an auth-failure, the next password prompt could cause an automated submission of an empty password unless the user interrupts.

- This case was missed by commit 5fb23f6ad9 that introduced automatic username/password submission


chipitsine:
I encountered the same issue.

thank for the fix!

mattock:
As per email discussions with @selvanair and @chipitsine this can be merged as-is. This is trivial, has been tested and fixes a bug.

PR #204, opened 2018-01-06
better alignment of warning message

chipitsine:
@selvanair , was "74" specified in purpose ?

if not, "80" seem to be better

74:

screenshot from 2018-01-06 23-14-40

80:

screenshot from 2018-01-06 23-18-07

### DIFF
--- a/res/openvpn-gui-res-cs.rc
+++ b/res/openvpn-gui-res-cs.rc
@@ -50,7 +50,7 @@ BEGIN
     CHECKBOX "Uložit heslo", ID_CHK_SAVE_PASS, 6, 42, 100, 10
     PUSHBUTTON "OK", IDOK, 20, 58, 50, 14, BS_DEFPUSHBUTTON | WS_TABSTOP | WS_DISABLED
     PUSHBUTTON "Zrušit", IDCANCEL, 90, 58, 52, 14
-    LTEXT "", ID_TXT_WARNING, 6, 74, 150, 10
+    LTEXT "", ID_TXT_WARNING, 6, 80, 150, 10
 END
 
 /* Auth Username/Password/Challenge Dialog */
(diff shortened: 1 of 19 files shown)


selvanair:
I reused the same control as the one used to show the red warning text on auth failure (as the two are mutually exclusive). And no particular reason for the original poor choice of positioning. Indeed your choice looks better.

While at it we could also increase the width of the window to a more pleasing ratio -- apart from better looks, that would also reduce chances of clipping in the title. 

Most of UI layouts can benefit from some repositioning and resizing for a more standard look.
These guidelines for spacings of controls is a good reference: 
https://msdn.microsoft.com/en-us/library/windows/desktop/dn742486(v=vs.85).aspx 

That said, your change looks good to me.

chipitsine:
as for "we could increase the width" part - I do not agree on that. I think we should shorten "OpenVPN - User Authentication (profile)" to "profile", because that window appears in certain context, no need to specify "OpenVPN" there.

or we can discuss it (with our users of course)

selvanair:
Yeah, profile name is the most useful info in there (which used to be missing in older versions) especially when one has multiple connections. And that's the one that often gets clipped. We could remove "OpenVPN" and save space. Also the context (user/auth) of this one and the static challenge dialog are pretty clear because "username" and "password" strings are in the dialog. Still, get some feedback from users.

Certificate password dialog definitely needs  "Certificate" in addition to profile name in the title  as it just asks for password.

BTW, I find a slightly wider aspect ratio more pleasing (will also need to widen the text boxes to match), but I'm no UI expert.

chipitsine:
multiple connections (you referenced as an argument of having profile name as the most important part) is something we should handle better.

openvpn installer creates single tap adapter during installation, when you connect to the second profile, the error message will not be perfect from UI / UX point of view.

I think we should create separate tap adapters per profile (we can even give proper names to those adapters) and specify adapter name in command line, it will improve UX a lot (users here still disable tap adapters for no reason, connect to several profiles, etc, etc, ...)

mattock:
In an email discussion with @selvanair and @chipitsine it was agreed that this should be merged as-is. The other changes suggested here can be done in a separate PR.

PR #205, opened 2018-01-14
travis-ci: add openssl-1.1.0 support

chipitsine:
also include changes from PR#176 to avoid merge conflict

### DIFF
--- a/.travis-build-openssl.sh
+++ b/.travis-build-openssl.sh
@@ -1,14 +1,15 @@
 #!/bin/bash 
 
+set -e
 
 if [ ! -f download-cache/openssl-${OPENSSL_VERSION}.tar.gz ]; then wget -P download-cache/ https://www.openssl.org/source/openssl-${OPENSSL_VERSION}.tar.gz; fi
 
 mkdir image
 tar zxf download-cache/openssl-${OPENSSL_VERSION}.tar.gz && cd openssl-${OPENSSL_VERSION}
-if [ "$CHOST" == "i686-w64-mingw32" ]; then export target=mingw; fi
-if [ "$CHOST" == "x86_64-w64-mingw32" ]; then export target=mingw64; fi
-./Configure --prefix=/ --cross-compile-prefix=${CHOST}- shared $target no-multilib no-capieng --openssldir=/etc/ssl --libdir=/lib -static-libgcc >build.log 2>&1 || (cat build.log && exit 1)
-make install INSTALL_PREFIX="${HOME}/image" INSTALLTOP="/" MANDIR="/tmp" >build.log 2>&1 || (cat build.log && exit 1)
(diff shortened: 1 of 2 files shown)


.travis-build-openssl.sh
```diff
@@ -1,14 +1,15 @@
...
 if [ ! -f download-cache/openssl-${OPENSSL_VERSION}.tar.gz ]; then wget -P download-cache/ https://www.openssl.org/source/openssl-${OPENSSL_VERSION}.tar.gz; fi
 
 mkdir image
 tar zxf download-cache/openssl-${OPENSSL_VERSION}.tar.gz && cd openssl-${OPENSSL_VERSION}
 if [ "$CHOST" == "i686-w64-mingw32" ]; then export target=mingw; fi
 if [ "$CHOST" == "x86_64-w64-mingw32" ]; then export target=mingw64; fi
-./Configure --prefix=/ --cross-compile-prefix=${CHOST}- shared $target no-multilib no-capieng --openssldir=/etc/ssl --libdir=/lib -static-libgcc >build.log 2>&1 || (cat build.log && exit 1)
-make install INSTALL_PREFIX="${HOME}/image" INSTALLTOP="/" MANDIR="/tmp" >build.log 2>&1 || (cat build.log && exit 1)
+./Configure --prefix=/ --cross-compile-prefix=${CHOST}- shared $target no-capieng --openssldir=/etc/ssl --libdir=/lib -static-libgcc >build.log 2>&1 || (cat build.log && exit 1)
+make install $([[ ${OPENSSL_VERSION} == "1.0."* ]] && echo "INSTALL_PREFIX" || echo "DESTDIR")="${HOME}/image" INSTALLTOP="/" MANDIR="/tmp" >build.log 2>&1 || (cat build.log && exit 1)
```
mattock: Wouldn't
```
"${OPENSSL_VERSION}"
```
with full quotes be more proper? Besides that I think this look good and should work fine.

chipitsine: it depends.

I'd be more happy without quotes (script will fail if variable is not defined), if I would want to mute "variable is not defined", I'd use quotes

mattock: Indeed, you're correct. One thing, though. Double square brackets are a bashism:

    [[ ${OPENSSL_VERSION} == "1.0."* ]]

As the file has #!/bin/bash at the top, I don't consider this a blocker. But our usage seems to be mixed. For example, in other places single square brackets are used:

    if [ "$CHOST" == "i686-w64-mingw32" ] ...

I think sticking to one syntax would be best, whatever that is.

chipitsine: I'm ok with bashism within travis-ci


ok, I'll change to one style soon

mattock: Ok, thanks! Besides that this PR looks good and Should Work(tm).

mattock:
This looks reasonable, so let's get this in.

PR #206, opened 2018-01-22
simplify caption on user/password auth window

chipitsine:
based on UX sessions with our users, it turned out that the most important part is "config name", while "OpenVPN" and "User Authentication" are obvious here. so

before (en)

screenshot from 2018-01-23 02-10-27

before (rus)

screenshot from 2018-01-23 02-10-57

after (rus)

screenshot from 2018-01-23 02-13-31


selvanair:
Agreed config name is the most important part (infact, we did not have it in the title until a year or so ago!) but I think some more context is required. OpenVPN is not the only program that may popup a user/pass dialog on user's desktop.

So I suggest to either keep OpenVPN or add the small app icon to the title bar.

chipitsine:
as for "OpenVPN" word - ask your users ? my users say "I know already what config name means, so OpenVPN is not needed"


as for the icon, which one ?

sh2

chipitsine:
https://github.com/OpenVPN/openvpn-gui/blob/master/res/openvpn-gui-res-fr.rc#L41

french already skipped "OpenVPN" word in caption :)

mattock:
I think having the OpenVPN logo (lock icon) in the titlebar would be good. Assuming the user has the openvpn-gui.exe shortcut on his/her desktop he/she should be somewhat familiar with it. Also, in most cases the authentication dialog pops up in response to user's actions, so the possibility for confusion is fairly small.

chipitsine:
ok, I'm taking a timeout with loading an icon. Hope to resolve it today

chipitsine:
sh3

res/openvpn-gui-res-cs.rc
```diff
@@ -38,8 +38,7 @@ END
 
 /* Auth Username/Password Dialog */
 ID_DLG_AUTH DIALOG 6, 18, 160, 95
-STYLE WS_POPUP | WS_VISIBLE | WS_CAPTION | DS_CENTER | DS_SETFOREGROUND
-CAPTION "OpenVPN - Ověření uživatele"
```
selvanair: The caption line may be left as is --  it will overwritten in INITDIALOG as you have above.

res/openvpn-gui-res-cs.rc
```diff
@@ -38,8 +38,7 @@ END
...
+STYLE WS_POPUP | WS_VISIBLE | WS_CAPTION | WS_SYSMENU | DS_CENTER | DS_SETFOREGROUND
```
selvanair: Adding SYS_MENU style is needed to get icon displayed, but that also adds a close button. Either disable that button or add StopOpenVPN() to WM_CLOSE (the same way as how ID_CANCEL is handled). Otherwise close will lead to the management interface stuck at "waiting for hold-release".

res/openvpn-gui-res-cs.rc
chipitsine: pardon ? why to leave caption line if it will be overwritten ?

chipitsine:
@selvanair , what do you think, should we remove "Cancel" button in favour of "X" ?

selvanair:
IMO, "Cancel" is a natural action for that dialog. I would say remove/disable the X or wire X to do the same thing as Cancel.

chipitsine:
Disabled "X" looks confusing, as for removal, I could not figure out how to remove it


So, let's keep both Cancel and "X"

selvanair:
Removing may be hard (custom window class, may be?) and not worth the trouble. The close button behaving like the cancel is quiet acceptable in my view.

chipitsine:
I'm done with this PR

(probably will address custom window class later, but not sure)

mattock:
@chipitsine @selvanair is this PR ready for merging?

chipitsine:
from my point of view - yes. I tested it, I (and our users) are satisfied with UX

selvanair:
Code looks right and a useful change.
@chipitsine: how does your sample image show the title as centered? In this case centered title is better but it display like that by default, does it..

chipitsine:
it is centered by default

screenshot from 2018-01-29 23-29-46

selvanair:
Not on my Windows 7! When the config name is short it looks strange when not centered. Is that a user configurable thing on WIndows?
capture

chipitsine:
well, even left align is good

(I've not found how to center it)

mattock:
Based on quick search the alignment of the title comes directly from Windows:

* https://superuser.com/questions/542495/how-can-i-left-align-title-bar-text-in-windows-8
* https://stackoverflow.com/questions/11824181/how-to-left-align-window-title-in-windows-8-2012

So there's not much we can/should do about it. Apparently the title is left-aligned if the app is running in Windows XP compatibility mode, otherwise it will be center-aligned on Windows 8 and later. Or you can tweak themes to relocate the title.

I will close the Windows title thingy as "notabug" and merge this as-is 😄 .

PR #211, opened 2018-01-28
Show assigned IPs, bytes in/out and GUI/core versions on status window

selvanair:
Commit 1 as its same as PR 210 (needed here for ipv6 address)

Commit 2 handles bytecount messages from OpenVPN and 3 adds the extra text fields to the status window for displaying IPs/stat/version (screenshot below). Only one new string needs translation: IDS_TXT_BYTECOUNT which by default displays as "Bytes in: %s out: %s"

capture


cron2:
Hi,

On Sun, Jan 28, 2018 at 06:26:33PM +0000, Selva Nair wrote:
> Commit 1 as its same as PR 210 (needed here for ipv6 address)
> 
> Commit 2 handles bytecount messages from OpenVPN and 3 adds the extra text fields to the status window for displaying IPs/stat/version (screenshot below). Only one new string needs translation: IDS_TXT_BYTECOUNT which by default displays as "Bytes in: %s out: %s"
> 
> capture
> 
> 
>   https://github.com/OpenVPN/openvpn-gui/pull/211

I won't clain to understand the windows gui stuff, but the parts
I do understand look good to me - so, feature-ACK, and an ACK for 
"no obvious badness in the string/formatting bits of C code".  

I just assume that the GUI bits work :-)

(As a side note, I noticed that parts of the code use TCHAR, while other
parts use wchar_t - which is the same when compiled with -DUNICODE, but
will differ when compiled without.  Is this a historic legacy?)

gert
-- 

selvanair:
Hi

On Sun, Jan 28, 2018 at 2:24 PM, Gert Doering <notifications@github.com> wrote:
>
> Hi,
>
> On Sun, Jan 28, 2018 at 06:26:33PM +0000, Selva Nair wrote:
> > Commit 1 as its same as PR 210 (needed here for ipv6 address)
> >
> > Commit 2 handles bytecount messages from OpenVPN and 3 adds the extra text fields to the status window for displaying IPs/stat/version (screenshot below). Only one new string needs translation: IDS_TXT_BYTECOUNT which by default displays as "Bytes in: %s out: %s"
> >
> > capture
> >
> > You can view, comment on, or merge this pull request online at:
> >
> > https://github.com/OpenVPN/openvpn-gui/pull/211
>
> I won't clain to understand the windows gui stuff, but the parts
> I do understand look good to me - so, feature-ACK, and an ACK for
> "no obvious badness in the string/formatting bits of C code".
>
>
> I just assume that the GUI bits work :-)
>
>
> (As a side note, I noticed that parts of the code use TCHAR, while other
> parts use wchar_t - which is the same when compiled with -DUNICODE, but
> will differ when compiled without. Is this a historic legacy?)


Yeah, I prefer explicit char and wchar and want to aviod TCHAR in new
code. We do a check for !defined(UNICODE) in main.c and error out
during build, so should be fine -- still have to be on the look out
for type mismatch warnings.

Thanks,

Selva

chipitsine:
can we use arrows instead of "in" and "out" ? it reads better, no need to translate

selvanair:
Something like   &#11015;500 B &#11014;400 B ?
The word "Bytes" may also need some explaining or "translation" in certain circles: in fact "Traffic" or "Transfer" may be preferred by some even in English? I'll leave that to translators: in/out, RX/TX, down/up, arrows,  whatever that clicks.

chipitsine:
arrows are just great.

as for "Bytes", we'd better to stay with abbrevation like KiB, no need to translate them

chipitsine:
also, I'd avoid duplicating like "Bytes in: 12494 (12.2 KiB)", I think 

"(arrow) 12.2 KiB" would be perfect

chipitsine:
TX/RX is something from telecom world. not sure we should use it in UI

selvanair:
I wouldn't call it duplicating; as a user I would want to see both. Once the number gets large (e.g., 3.2 GiB), it will take a lot of transfer before the display changes --- so bytes show the progress, short form gives an easy to grasp value. 

Before you say it: one could get fancy and show transfer rate instead, but I'll leave that to others.

chipitsine:
well, I did not consider transfer rate (it is also nice idea)

I meant something like "df -h", i.e. "human readable"

chipitsine:
another UI idea would be use commas, like in ms windows


nic-status

mattock:
👍

I've seen &darr; and &uarr; in some other apps and I think it looks pretty neat when of proper size and color. That said, I think the arrows can be implemented in another PR so that this one does not get delayed unnecessarily.

selvanair:
Will rebase and test again and then we can merge this.

selvanair:
Removed the already merged extract-ipv6-address commit from this and rebased.
- Arrows using unicode is easy to do and look ok, but fancier (coloured?) arrows may be better and is left for future.
@chipitsine: adding commas is easy: I would be tempted to use a locale-dependent number format API for this (like 967 295,000 vs 967.295,000 vs 967,295,000 vs 96,72,95,000) but in practice even native users may not understand their locale-dependent formats when applied to things like bytes... Need to look into it further, so leaving it for future.
@mattock: I think we can merge this based on Gert's quick review and my own tests.

PR #212, opened 2018-01-29
Set a fall back for editor if file association fails

selvanair:
- When registry keys were simplified, the log viewer and editor
  entries were removed in favour of file associations that the
  user can independently control.

  Yet, there are times when an associated application fails to start
  or no associations are present etc., and its useful to have a fall
  back editor setting to open .ovpn and .log files.
  This patch sets that default as "notepad.exe"

- Also remove the unused profile_dir[] in registry.c


mattock:
Afaics this looks good. Lazy ACK'ing and merging.

PR #214, opened 2018-01-31
add EXSTYLE WS_EX_TOPMOST to login/auth window

chipitsine:
(same way it is already done for key password dialog)


selvanair:
This will keep the user-pass dialog at the top, isn't it? Sounds like the right thing to do -- I will give it a test.
@chipitsine  : your recent improvements/suggestions to user-pass dialog has not been applied to the static challenge dialog. Does that imply your users do not use one-time password? If they do is there a different approach than using static challenge?

chipitsine:
@selvanair - our users just use login/password 
I will have a look at static challenge a bit later

selvanair:
I'm confused. If there is no OTP you do not need auth-token. So why use auth-token?

chipitsine:
I didn't enable it, it was enabled by default.
also, I find it useful, because it speeds up password check

selvanair:
Interesting.. 

Auth-token is not enabled by default. Its created either by using --gen-auth-token in the server config or pushing a token from client-connect script or plugin. Of course it will speed up user authentication if your verify script is slow only because that script cant be skipped (if --gen-auth-token is in use) during renegotiations. But I wouldn't have guessed someone would use it for that purpose.

One can never guess how end users make use of an option.

chipitsine:
@selvanair , I'll try to disable auth-token (I checked, they were enabled by some other person for the reason not clear to me)

.... and yes, I considered auth-token as something "generally useful, because it saves time on re-auth"

selvanair:
This keeps the auth-user-pass dialog at top without blocking access to other windows. ACK.

PR #218, opened 2018-02-11
Save username only if/when saving auth password

selvanair:
There has been a few requests to do this:
Eg., Trac #866 https://community.openvpn.net/openvpn/ticket/866
Note that this ties saving username to saving password -- there is no option to save one and not the other. Also if disable_save_passwords is active, the username is also not saved.

Commit message:

- Currently username is always saved. This changes that to save
  username only when auth password is saved.
- Usernames saved by previous versions are automatically migrated
  if password is also saved, else cleared, to enforce the new behaviour.
- Username and password are saved as encrypted by DPAPI.

Note: Setups in which saving of password is not enabled, any previously
saved username will be forgotten. However, the migration or clearing of
username is attempted only when a connection is started. So previously
saved usernames may stay in the store if a config is unused.

### DIFF
--- a/openvpn.c
+++ b/openvpn.c
@@ -511,6 +511,7 @@ UserAuthDialogFunc(HWND hwndDlg, UINT msg, WPARAM wParam, LPARAM lParam)
             else
             {
                 DeleteSavedAuthPass(param->c->config_name);
+                DeleteSavedUsername(param->c->config_name);
                 Button_SetCheck(GetDlgItem (hwndDlg, ID_CHK_SAVE_PASS), BST_UNCHECKED);
             }
             AutoCloseCancel(hwndDlg); /* user interrupt */
@@ -524,7 +525,6 @@ UserAuthDialogFunc(HWND hwndDlg, UINT msg, WPARAM wParam, LPARAM lParam)
                     show_error_tip(GetDlgItem(hwndDlg, ID_EDT_AUTH_USER), LoadLocalizedString(IDS_ERR_INVALID_USERNAME_INPUT));
                     return 0;
                 }
-                SaveUsername(param->c->config_name, username);
             }
             if (GetDlgItemTextW(hwndDlg, ID_EDT_AUTH_PASS, password, _countof(password)))
             {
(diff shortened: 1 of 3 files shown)


chipitsine:
there are couple of warnings

https://travis-ci.org/OpenVPN/openvpn-gui/jobs/340037416#L1842

https://ci.appveyor.com/project/mattock/openvpn-gui/build/cygwin-163/job/ox1ihid404e9htad#L219


should we turn warning into errors ?

selvanair:
We have ruled out -Werror a number of times in the past. -Werror is no good in production or for CI just because not all warnings are fixable.

AFAIK, there is no fix[*] for that particular warning about the "I64" format spec except for dropping 
`-pedantic`. 

[*] Well, overriding printf and friends in Windows runtime with -D__USE_MINGW_ANSI_STDIO is probably an option. But I've no idea how well-tested is mingw's replacement i/o functions.

mattock:
@selvanair is this something we could lazy-ACK before 2.4.6? I've started the release machinery now, but won't be able to finish the release until tomorrow.

selvanair:
@mattock Some users were forcefully asking for this (the "don't save my username without asking" complaint) but it seems none of them have responded to this PR. And, one user has given a thumps down.

So I'm conflicted as to whether users want this or would prefer just to have an option to disable caching of username, leaving the default behaviour unaffected.

cron2:
Hi,

On Mon, Apr 23, 2018 at 08:06:34AM -0700, Selva Nair wrote:
> @mattock Some users were forcefully asking for this (the "don't save my username without asking" complaint) but it seems none of them have responded to this PR. And, one user has given a thumps down.
> 
> So I'm conflicted as to whether users want this or would prefer just to have an option to disable caching of username, leaving the default behaviour unaffected.

I'm not sure I understand the PR correctly.

Is this:

old:
  always save username
  sometimes save password (if permitted)

new:
  sometimes save (username + password) (if permitted)
  never save username alone

if yes, I think this makes 2FA setups less convenient - I am a lazy one, 
and I would like the GUI to retain my username (because that's always the 
same), and never save password (because that is non-useful for 2FA).

So I'd rather not see the change, at least not unconditionally.

gert

-- 

selvanair:
cron2: Your understanding of the PR is correct.

TLDR; Let's skip this PR and wait for a better one that provides an option to switch off saving username to satisfy those who are offended by that. 

I agree the current behaviour is convenient in that username is always cached. That said, for 2FA, a better approach would be to use the static challenge dialog (which has separate boxes for OTP and password) and always save the password. Unless OTP is the only password in use. But those who are already invested into server setups that want password+OTP as a single string would want to save username but not password.

selvanair:
I'll close and re-submit when ready.

PR #219, opened 2018-02-12
Do not allow "echo save-passwords" to override disable_save_passwords

selvanair:
- As disable_save_passwords may be enforced system-wide by an Administrator,
  "echo save-passwords" should not be allowed to over-ride it. This was
  overlooked in commit d4090a8842990ba992d843c49eaec1760ac03b0e.
  Fix it by ignoring this echo directive if disable_save_passwords is in
  effect. Also write a log message to the status window.

### DIFF
--- a/openvpn.c
+++ b/openvpn.c
@@ -956,7 +956,11 @@ OnEcho(connection_t *c, char *msg)
     }
     else if (strcmp(msg, "save-passwords") == 0)
     {
-        c->flags |= (FLAG_SAVE_KEY_PASS | FLAG_SAVE_AUTH_PASS);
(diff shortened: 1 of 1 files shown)


selvanair:
This is a minor logic fix.. can we get an ACK or resort to another lazy-ACK?

cron2:
Hi,

On Sat, Mar 10, 2018 at 04:04:14PM +0000, Selva Nair wrote:
> This is a minor logic fix.. can we get an ACK or resort to another lazy-ACK?

ACK.  Code makes sense, logic fix doubly so.

You've been flooding us with interesting things to review :-)

gert

-- 

selvanair:
Thanks, Gert.  Merging.

PR #225, opened 2018-02-27
Update TODO

Prockwin:
Verified support for usernames and passwords containing the symbol "

### DIFF
--- a/TODO
+++ b/TODO
@@ -1,7 +1,5 @@
 TODOs for OpenVPN-GUI
 
-* support usernames and passwords that contain a " character
-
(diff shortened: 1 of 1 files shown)


selvanair:
That TODO file is obsolete. We should just delete it. TODO is a personal thing and better kept track of in one's local repos -- I have my own TODO.local.

If there is anything worth saving in that file, convert them to issues.

Prockwin:
I checked the support for the symbol ". I think that this line can be deleted.

selvanair:
Thanks for the patch. But this is no more relevant, so closing.

Issue #227, opened 2018-03-09
Password field is blank after session expiration


NullSquatch:
# Main Issue

To provide some background, the server I connect to uses `static-challenge`.  I prefer to leave my connection active during the day and put my workstation into sleep mode each night.  When I wake my workstation the next morning, the OpenVPN connection is no longer active and I'm presented with an authentication popup.  This by itself is not an issue.

The problem is that the `Password` field on the popup is blank, despite the fact that I have the `Save password` option enabled.  I would only expect `Password` to be blank if the authentication failure was specifically due to an incorrect username/password.

The client log shows these relevant lines:

```
MANAGEMENT: >STATE:1520524314,GET_CONFIG,,,,,,
SENT CONTROL [OpenVPN Server]: 'PUSH_REQUEST' (status=1)
AUTH: Received control message: AUTH_FAILED,SESSION: Your session has expired, please reauthenticate
SIGUSR1[soft,auth-failure] received, process restarting
MANAGEMENT: >STATE:1520524314,RECONNECTING,auth-failure,,,,,
```

This shows that the session expired, so it makes sense that the popup is being displayed.   If I had to guess, I'd say that the GUI code isn't differentiating between session expiration failures and failures caused by an incorrect username/password.

# Additional Observations

1) In this particular situation, the cursor is in the `Password` field, but I would expect the cursor to instead be in the `Response` field, as that is the field that requires new input in order to successfully connect.
1) In this particular situation, the bottom of the dialog does not state the reason for the previous disconnection.  I would expect it to show the reason returned by the server, i.e. `Your session has expired, please reauthenticate`.
1) If I happen to enter the correct username and password but use an incorrect value for the `Response` field, the dialog clears the `Password` field and shows `Wrong username or password. Try again...` at the bottom of the popup, despite the log showing that the reason for the failure was `AUTH: Received control message: AUTH_FAILED,Google Authenticator Code is incorrect`.  I would expect this to be handled similar to session expiration, i.e. the `Password` field should not be cleared, the cursor should be in an empty `Response` field, and the message at the bottom should read `Google Authenticator Code is incorrect`.

selvanair:
[...]

Issue #217, opened 2018-02-11
On handling saved password after an auth error


selvanair:
The GUI currently clears the saved password if an authentication error is received. Some have pointed out that automatically deleting data saved by the user is inappropriate.

The behaviour of RDP has been mentioned as an example. Based on that I've made a patch that does the following. Saved password is not deleted on auth-error but the subsequent user-pass dialog will show up with username filled-in but password empty (same as now). The error text that says something like "wrong username or password. Try again.. " is also unchanged.  The user can type a new password and try again or press cancel. If cancel is pressed, the saved password is preserved and will get used the next time.

This way, user's data is not forcefully deleted, but at the same time the password will not appear prefilled with failed data. Prefilling would probably be ok if we have an option to reveal the hidden characters in the password field (hint: patches welcome to add a Windows 10-like eye).

If there is sufficient interest I'll issue a PR.

PR #228, opened 2018-03-10
Do not clear saved passwords on verification failure

selvanair:
After a failure the auth-pass dialog is shown with the password
field prefilled but highlighted. This allows the user to easily
overwrite the password or resubmit the old password if the
failure was temporary.

After a private key passphrase failure, the dialog is not
prefilled with saved password as this failure happens locally
and in such cases the password is very likely wrong.

If the user aborts the dialog by pressing cancel, the saved
password will get used during the next connection attempt.

Wrong username or password warning text is changed to: "Wrong
credentials".

See also issues: #227 #217

### DIFF
--- a/openvpn.c
+++ b/openvpn.c
@@ -307,12 +307,6 @@ OnStateChange(connection_t *c, char *data)
                 c->failed_auth_attempts++;
             else if (strcmp(message, "private-key-password-failure") == 0)
                 c->failed_psw_attempts++;
-
-            if (strcmp(message, "auth-failure") == 0 && (c->flags & FLAG_SAVE_AUTH_PASS))
-                SaveAuthPass(c->config_name, L""); /* clear saved password */
-
-            else if (strcmp(message, "private-key-password-failure") == 0 && (c->flags & FLAG_SAVE_KEY_PASS))
-                SaveKeyPass(c->config_name, L"");  /* clear saved private key password */
         }
 
         c->state = reconnecting;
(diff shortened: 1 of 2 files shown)


NullSquatch:
I've been using this for the last week without issue.  My server uses static-challenge with session expiration, and this is a nice improvement because I don't have to re-type my password each time.

selvanair:
Thanks for the feedback. We'll try to get this into the next release.

selvanair:
@mattock this may be a good candidate for getting merged before the upcoming release? It should also help those who were complaining of lost saved password on auth-token errors. A fix for the auth-token handling in the core is still in the review process.

cron2:
Hi,

On Fri, Mar 09, 2018 at 08:49:56PM -0800, Selva Nair wrote:
> After a failure the auth-pass dialog is shown with the password
> field prefilled but highlighted. This allows the user to easily
> overwrite the password or resubmit the old password if the
> failure was temporary.
> 
> After a private key passphrase failure, the dialog is not
> prefilled with saved password as this failure happens locally
> and in such cases the password is very likely wrong.
> 
> If the user aborts the dialog by pressing cancel, the saved
> password will get used during the next connection attempt.
> 
> Wrong username or password warning text is changed to: "Wrong
> credentials".

Sounds good, so definitely feature-ACK.  Haven't stared at the code yet.

gert
-- 

mattock:
Feature-ACK also from me. If we want wider testing we could send a link to a test installer to openvpn-devel list. I think this feature might be interesting enough for people to try it out.

mattock:
I will merge this based on lazy-ACK, so that we get this into OpenVPN 2.4.6.

PR #188, opened 2017-09-24
Support sending commands to a running instance of the GUI

selvanair:
A test executable is [here](https://github.com/selvanair/openvpn-gui/releases/tag/v11_testing1)

There are a number of commits here --- the main commit (df07beb) does what the heading says. Commit 2467640 adds as button to the menu for reconnect (a convenience as the button in the status window is hard to reach). This one is now referred to by the rest of the commits so leaving it in here instead of opening a separate PR for it.

All other commits are improvements based on feedback and comments by those who tested an earlier version.
 
- New option --command <action> <params> to send commands to a running instance of openvpn-gui.exe

  Supported actions are
      connect, disconnect, reconnect
  each of which takes the name of the config (with or without the extension .ovpn) as a parameter;
      disconnect_all, exit
  which take no parameter and
      silent_connection
  which takes an optional parameter = 0 or 1 (1 is the default)

  Examples: with the gui running, start a new instance as

  `openvpn-gui.exe --command disconnect myvpn` : ask running instance to disconnect myvpn if connected
  `openvpn-gui.exe --command status myvpn`    : ask running instance to show the status window for myvpn if available
 `openvpn-gui.exe --command disconnect_all`   : ask running instance to disconnect all active connections

- The second instance exits after issuing a SendMessage to the already running instance. If no action is specified, the running instance is notified to show a balloon to alert the user

- These messages may also be sent from scripts as COPYDATA messages with the wData element specifying the action to execute and lpData  a pointer to the parameter. The dwData param must be one of WM_OVPN_xxx with xxx = START, STOP, RESTART, STOPALL, EXIT or SILENT. See main.h for their values.

v2: Bug fixes based on test reports from larson0815 here: selvanair#5
and cron410 here: #104

Signed-off-by: Selva Nair <selva.nair@gmail.com>

PR #230, opened 2018-03-16
remove unused resource

chipitsine:
after https://github.com/OpenVPN/openvpn-gui/pull/188 that string
is not used anymore


mattock:
Looks good. I have a vague recollection that at some point we agreed that old defines should be left in place, which is what this PR does. On that premise I will ACK and merge this now.

PR #231, opened 2018-03-16
travis-ci: switch to openssl-1.1.0 by default

chipitsine:


mattock:
Is there any reason to keep testing OpenSSL 1.0.2 builds? If not, then this PR looks good.

chipitsine:
I think that person who actually build and release an installer should know :)

selvanair:
We use OpenSSL only for privatekey password change and do not plan to add more dependency on it (all new features use native API for digests, wrapping passwords, base64 etc.), so testing the version used for release builds (1.1.0 for now) is good enough.

Issue #237, opened 2018-03-26
Multiple GUI instances no error message displayed


TinCanTech:
Tiny issue, steps to reproduce:

1. Exit running openvpn-gui
2. From cmd prompt: \progra~1\openvpn\bin\openvpn-gui --help (do not close help window)
3. From cmd prompt: \progra~1\openvpn\bin\openvpn-gui --connect my_profile.ovpn
**No error is given** and the gui does not start my_profile.ovpn
Now close help window and start (as in 3 above) and all works as expected.

selvanair:
Thanks for reporting this. Looks like a bug introduced by the recent change to show a friendly message pointing to the running instance when a second GUI instance is started. In case of --help the GUI is not fully started up (tray icon not created), so the second instance has no icon to point the user to, but yet it exits as the GUI is already running. Will fix. 

By the way, the opposite order should work -- i.e., have the GUI running fully started up (with tray icon) and then execute another instance with --help.

selvanair:
@TinCanTech Can you check this fixes the problem?

TinCanTech:
Sorry, I do not currently have a way to test.

selvanair:
Anyone else able to test this? If pre-built exe is needed, see the appveyor artifacts: https://ci.appveyor.com/project/mattock/openvpn-gui/build/cygwin-179/job/rkhsl4v24jynfj23/artifacts
I believe it is linked against openssl-1.0 so if those dlls are in the path, it should run from any folder.

What to test: start as "openvpn-gui.exe --help" from command line. Keeping the help window open, try to start another instance either from command line or by clicking on the GUI icon. The second instance should startup and minimize to the tray as expected.

TinCanTech:
I tested the binary from your link and the issue appears to be solved.
I started it in multiple different ways and they all behaved as I would expect.

selvanair:
Wonderful! Thanks.

mattock:
Fixed by PR https://github.com/OpenVPN/openvpn-gui/pull/241 -> closing.

PR #241, opened 2018-04-02
Fix detection of an already running instance

selvanair:
When openvpn is run with --help option it pops up a help
message and exits when that window is closed. Such instances
cannot accept any commands and should not be treated as a
running instance.

Fixes issue: #237

Additionally log errors to the event log when sending a command to a 
running instance fails. This supplements the non-zero exit-code
set in such cases.


mattock:
As discussed in https://github.com/OpenVPN/openvpn-gui/issues/237 this PR works as advertised. Merging based on that testing and a lazy-ACK.

PR #243, opened 2018-04-12
Added shortcuts (mnemonics)

### LINKED ISSUE #242: Improve dialogs - add shortcuts to input fields/labels
In order to improve dialogs is it possible to add shortcuts (e.g. "&Username") to the input field labels and/or other controls? This would help interaction with Keepass auto typing, for example (an auto type sequence could contain Alt-U to select the username input field first, etc.)

woodychuck:
Added keyboard shortcuts (input fields) to enable direct selection of input fields, e.g. for deterministic auto type sequences with password managers.
Fixes #242


woodychuck:
Done

selvanair:
I tested this and noticed a few glitches (on Windows 7):
In settings menu:
English: Alt-a jumps to &Append to Log as expected, but doesn't toggle the check mark
German: The radio button for balloon: once moved to another radio button, moving back to On connect fails. Eg., Alt-e followed by Alt-b.
(Some text is clipped but that's not this PR's fault).

Nice to have: 
(i) navigate between tabs of the settings menu using keys -- is this possible?
(ii) bind keys for OK and Cancel in the setting menu

woodychuck:
This is weird - I am not a GUI expert, and all I did was adding the ampersands to the resource file expecting consistent results, which obviously aren't what we get.

However I was able to resolve the first issue (English) by changing the mnemonic to A&ppend, i.e. Alt-p - this worked. Maybe Alt-A is somehow reserved? Also, changing the balloon radio button in the German resource file from "Ver&binden" to "Verb&inden" worked for me - again, I have no idea what causes Alt-b to fail. Would you suggest to go forward with these little changes?

Concerning the nice to haves: navigating between tabs generally is done with Ctrl-PgUp and Ctrl-PgDn, and also works for me in the settings dialog. Bind keys for OK and Cancel should be added in the same way, but for the settings they are not part of the resource file apparently. I looked into main.c / ShowSettingsDialog() but frankly I have no idea how this is handled by Windows devs ;-)

selvanair:
You are right, changing the  OK and Cancel text in property sheets is beyond the scope of this patch. Those are built-in buttons (not from resources) so I think we'll have to make it modeless to get a handle and send messages to change the text. Currently its created as a modal dialog.

Using  A&ppend and Verb&inden is fine with me. I still wonder what's the logic behind the original not working..

woodychuck:
OK I changed the misbehaving mnemonics as suggested, hope it's working now as expected.

selvanair:
Look good. ACK. Please squash these commits for a cleaner history and force push.

Note to self: merge has to wait as the last release tags are not yet here..

woodychuck:
Done (with a little help from @florianbeer 😀)

Issue #235, opened 2018-03-25
Only 50 configurations allowed - need unlimited!


user18x:
Hello,

i use openvpn 2.4.4 (now openvpn 2.4.5) but only 50 configurations allowed, i need more!
Can it be "fixed" to unlimited in a newer version?

I've more than 50 config files in the openvpn config-folder.
So if i click on openvpn-icon (gui; graphical user intercae) in tray a message "pops up" with only 50 configurations allowed (sorry the message is in german). 
![nur 50 konfigurationen](https://user-images.githubusercontent.com/37771000/37874149-a1f90134-3029-11e8-901a-a26c2c52d2ad.jpg)

user18x:
Hmm, will it be fixed?

mattock:
@user18x can try with 49 configurations? Does the UI work ok with that many? Can you provide a screenshot?

user18x:
49 configs and less does work!
But i need more than 50!

chipitsine:
@mattock , @selvanair - what if we change default with 250 ?
UI wouldn't be worse

user18x:
Why a limit?
If it needs a limit, then 250 sounds good.

chipitsine:
@user18x - historically, there's limit. please look at the code, it is written such way.
if you are not happy with that limitation, you can always send PR (or hire a developer)
if that PR is good, it will be merged and limitation will go away

selvanair:
Technically, in the current design one can push to about ~6000 configs. But, IMO, anything more than a few dozen will be hard to use with popup menus, and a list view is better and more versatile for such large lists. But such a change is some work and unlikely to happen unless someone sends in a patch.

I suggest we build the release binaries with MAX_CONFIGS = 1000. It wont affect the look, feel or performance[*] for most users who have a few configs. And those with 100's can still have their fun.

There is also a max sub-directories limit of 50 but that can be lifted by changing the way config files are scanned for -- easy to fix.

[*] Except for a couple of MB wasted memory (bss) which can be avoided with some code changes but not worth the effort? The data size doesn't change so the executable will remain the same size.

chipitsine:
I recall some old thread with enabling mouse wheel on config list.
I tried on my test 49 configs, mouse wheel is not working.

can we enable it as well ? (I'll do some investigation myself on that)

selvanair:
[...]

PR #247, opened 2018-04-21
Implement a nested menu for configs in subdirectories

selvanair:
By default, enabled only when number of configs > 50 but can be forced using `--config_menu_view 2` on command line. TODO: make this configurable from the settings menu
 
As an example the following directory structure 

CUSTOMERS/agony/dreamcity/dreamcity.ovpn
CUSTOMERS/agony/boonies/boonies.ovpn
CUSTOMERS/woeful/woeful-gw/woeful-gw.ovpn
PROVIDERS/dont-trust/washington/washington.ovpn
PROVIDERS/dont-trust/london/london.ovpn
PROVIDERS/broken-cipher/bypass-netflix/bypass-netflix.ovpn
WORK/scinet/scinet.ovpn
sample.ovpn
broken/broken.ovpn

Displays as

CUSTOMERS> agony>  dreamcity> connect,disconnect
                                       boonies>   connect...
                         woeful> woeful-gw> connect...
PROVIDERS> dont-trust>    washington>     connect...
                                            london>         connect...
                       broken-cipher> bypass-netflix> connect...
WORK> scinet> connect...
sample> connect...
broken> connect...

(for so few configs, need to enable the nested view using `config_menu_view 2`

I prefer to keep each config in its own directory with the same name (the GUI's import command does the same too). To reduce nesting in such cases the display squashes single config directories one level up.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -49,12 +49,6 @@ AC_ARG_ENABLE(
 	[enable_password_change="yes"]
 )
 
-AC_ARG_VAR([MAX_CONFIGS], [specify the maximum number of configs @<:@default=50@:>@])
-if test -z "$MAX_CONFIGS"; then
-	MAX_CONFIGS=50
-fi
-AC_DEFINE_UNQUOTED([MAX_CONFIGS], [$MAX_CONFIGS], [Maximum number of config files supported.])
-
 case "$host" in
 	*-mingw*)
 		CPPFLAGS="${CPPFLAGS} -DWIN32_LEAN_AND_MEAN"
--- a/main.c
+++ b/main.c
@@ -879,3 +879,20 @@ MsgToEventLog(WORD type, wchar_t *format, ...)
     msg[1] = buf;
     ReportEventW(o.event_log, type, 0, 0, NULL, 2, 0, msg, NULL);
 }
+
+void
+ErrorExit(int exit_code, const wchar_t *msg)
+{
+    if (msg)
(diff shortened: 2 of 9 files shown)


selvanair:
A bug fix commit added (see #235) and rebased to master.

selvanair:
@mattock This has been widely tested so could possibly be given a lazy ACK in time for 2.4.7?

As it stands, nested display is enabled by default but we could make it turned on only if number of configs is > 5 or so.

cron2:
Hi,

On Thu, Feb 07, 2019 at 07:58:48AM -0800, Selva Nair wrote:
> @mattock This has been widely tested so could possibly be given a lazy ACK in time for 2.4.7?
> 
> As it stands, nested display is enabled by default but we could make it turned on only if number of configs is > 5 or so.

Looking at the patch set, it definitely brings desirable improvements.

I've tried to review the code, and while I can't claim to understand
all the Windows details, the approach and code looks reasonable to me - 
so together with the testing seen, this is good enough for me.

So, ACK.  Please merge.

gert

PS: and "ouch!" on the "bug fix config-group data structure" commit :-)

-- 

selvanair:
> So, ACK. Please merge. gert 
>PS: and "ouch!" on the "bug fix config-group data structure" commit :-)

yeah, embarassing storage of pointers and realloc ! Dangers of extending exisitng code instead of writing from scratch -- a linked list would have been more natural here.

Thanks for the ACK.

PR #249, opened 2018-04-24
Change layout of text in --help output dialog

### LINKED ISSUE #221: Output of `openvpn-gui --help` is messy and inconsistent
When I was checking if some translations fit in the HMI, I noticed the output of `openvpn-gui.exe --help` (see left image below). The line breaks cause the text to be mixed with the actual command line arguments, some of the semicolons are not aligned and some 1/0 options are explained for both cases, some only for the '1' option.

The first thing I wandered, is why a `MessageBoxEx` is used to show this output, and not stdout. I think the only situation that you are calling `openvpn-gui.exe --help` is when you are already in command line mode so you will see the output when it is on stdout.

If it is desired that the output stays a MessageBoxEx (which seems to have a fixed width) I think the text must be adapted so it fits better in the MessageBox (see right image below). If you want, I can provide a PR for the changes I did to create the screenshot, but eventually it requires changes in the resource files of all languages.

help

EagleErwin:
Fixes #221 for en, de, fr, es and nl.

I have chosen to only modify the dialog for these languages, because these are the only languages that I more or less feel comfortable for.

See the attached screenshots.
en
de
es
fr
nl

### DIFF
--- a/main.h
+++ b/main.h
@@ -40,7 +40,7 @@
 #define MAX_LOG_LENGTH      1024/* Max number of characters per log line */
 #define MAX_LOG_LINES		500	/* Max number of lines in LogWindow */
 #define DEL_LOG_LINES		10	/* Number of lines to delete from LogWindow */
-#define USAGE_BUF_SIZE		2048	/* Size of buffer used to display usage message */
+#define USAGE_BUF_SIZE		2560	/* Size of buffer used to display usage message */
 
 /* Authorized group who can use any options and config locations */
 #define OVPN_ADMIN_GROUP TEXT("OpenVPN Administrators") /* May be reset in registry */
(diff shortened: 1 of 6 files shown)


selvanair:
Thanks for the patch. 

Improving the layout in help message is a good idea, but not sure whether this is really better than original. See below for what I get (window on left is current, right is with this patch). The long and narrow layout extends beyond the desktop (this is with a remote desktop of only 1200x700  so may not be an issue for many) and some lines like "connect cnn" are better aligned in the original.

For a definitive improvement, we may have to replace the built-in message box by a custom dialog, add scroll bars if needed, and not use tabs for alignment.

capture

EagleErwin:
This gives me some new insights. I was not aware of the fact that a MessageBoxEx increases the width if the vertical space is not sufficient to display everything on the 'default' width. I totally agree with you that the changes in this PR make the situation worse for smaller resolutions.

I don't mind rejecting this PR for this reason.

selvanair:
As discussed earlier this is not really an improvement. If there are no objections let's close this.

Issue #270, opened 2018-06-21
Shall we disable scripts (run by openvpn.exe) by default ?


selvanair:
Considering the recent report on newer ways to exploit scripts in the ovpn file, I propose to disable all such scripts when a connection is started using the GUI (easy to do). We can still allow scripts that are executed by the GUI itself as those require explicitly named batch files to exist in the config folder.

Note that, in 2.4.x releases, we use the interactive service to start openvpn.exe, so privilege escalation though scripts is not a concern[*]. But exploits through commands embedded in the config (.ovpn) files is a valid threat as we currently do not have a way of signing config files.

[*] Some other GUI implementations disable scripts because of the possibility of privilege escalation as they run openvpn.exe with high privileges even if the user does not have those rights.

chipitsine:
we do not use interactive service if user is administrator

selvanair:
On Thu, Jun 21, 2018 at 1:15 PM, Ilya Shipitsin <notifications@github.com>
wrote:

> we do not use interactive service if user is administrator
>
As I tried to explain in the post "admin or not" is unlreated to why we may
want to disable scripts. Sounds like I shouldn't have said anything about
interactive service.

Anyway, not using the iservice when admin is by design. We could revisit
this as Vista is no longer supported but let's discuss that in a separate
thread.

Selva

cron2:
Hi,

On Thu, Jun 21, 2018 at 09:43:58AM -0700, Selva Nair wrote:
> Considering the recent report on newer ways to exploit scripts
> in the ovpn file, I propose to disable all such scripts when a
> connection is started using the GUI (easy to do). We can still allow
> scripts that are executed by the GUI itself as those require
> explicitly named batch files to exist in the config folder.

Short version: yes.


Long version: I have been thinking how to tackle this, and one of the
ideas floating around was to have a dial in the GUI that influences
a --script-security command line being passed to the openvpn.exec
process, at the *end* of the command line (overriding whatever is in
the config).  

Default would be --script-security 0 or 1 ("nothing" or "netsh.exe is 
permitted", according to the docs) but if a user really feels like they 
know what they are doing they can set it to 2 or 3.


Longer version: David wants full blown policy config ("this instance is
allowed to do <list of things, nothing else>") which I find too heavy :-)


What are your ideas how to tackle this?

gert
-- 
"If was one thing all people took for granted, was conviction that if you 
 feed honest figures into a computer, honest figures come out. Never doubted 
[...]

PR #271, opened 2018-07-03
Script security

selvanair:
Ref: issue #270 

See individual commit messages for details. Here is a summary.

1. Enforce --script-security 1 but allow other legal values (0 to 3) or -1 to mean do not override the value in the config file
2. Use a sha1 hash of the file to detect changes and revert the setting to the default
3. A helper patch to easily handle user preferences for password save (needed for next commit)
4. Add a dialog for setting the preferred script-security value. Also move some config-specific preferences such as save password to this dialog.

capture

5. Add a checkbox to Settings->General dialog for enabling a global kill-switch for this feature. I was reluctant to add this but it will be missed by users who know what they are doing.


selvanair:
My tests show this could be very annoying if scripts cause a fatal error when script-security is overridden. So we should consider this only if we can merge something like the patch I proposed for OpenVPN core: [patch-395](https://patchwork.openvpn.net/patch/395/)

selvanair:
> My tests show this could be very annoying if scripts cause a fatal error when script-security is overridden. So we should consider this only if we can merge something like the patch I proposed for OpenVPN core: patch-395

As discussed in openvpn-devel the fatal error seems to be the right thing to do so this could be merged without needing any changes to the core.

Anyone ready to test this patch?

TinCanTech:
With regard to the script security drop down box, this is not intuitive.
I recommend the drop down box list the numeric value of the setting.
eg:
```
0. No external commands (Very restrictive, may cause config to malfunction)
1. Built in commands only (Recommended)
2. All scripts (Unsafe)
3. All scripts and access to passwords (Unsafe)
4. Use config file settings (Very unsafe)
```
I understand `4` is technically `-1` but users are not going to understand that.

selvanair:
Not convinced of the value of including a numerical number as these options do not really correspond to the number in the code (eg., there is no "use config file option" in the core).

The reason for the recommended setting at the top is that is the one selected by default and the best option. But agreed that  probably we should re-order at least the rest -- whatever users will find convenient is fine with me.

By the way -- "use config file setting" is a valid option and potentially "extremely unsafe" at the same time: in the end, if the setting the GUI imposes does not match what is in config, the user is very likely going to get a FATAL error -- recall the discussion with Steffan which led to keeping the FATAL error when config wants to run a script but we override it in the GUI.

TinCanTech:
Only with regard to the drop down list with a numeric, does the default option have to be the first on the list ? (I mean, is that required by the windows widget behind it or can the list appear numerically ordered but select `1` by default)

TinCanTech:
> Not convinced of the value of including a numerical number as these options do not really correspond to the number in the code (eg., there is no "use config file option" in the core).

They do correspond, other than the "use Config file" option (which can be made clear in the GUI). Also, that they do correspond reinforces the core values.

TinCanTech:
All my tests completed with expected results.

Other than the list being out of logical order (which I presume can be corrected) and the possible change to a more appropriate gadget for that list (perhaps an up/down slider or something), I could not find any way to subvert script-security.

selvanair:
This implementation needs a fix. If the user has limited privileges and the config is in the system-wide location, adding the --script-security option from the GUI will cause interactive service to wrongly reject the connection attempt as not authorized.

Options: for configs is in system-wide location, do not enforce a safe script security OR white-list "--script-security" as an option that any user can override. I prefer the latter as on Windows admins are often lay users who could unwittingly install insecure configs in the central location.

selvanair:
I'm in the process of updating this PR like this:

Minimize the dialog about script security setting to just two options: (i) built-in commands only (recommended) and (ii) Use config file setting (potentially unsafe).

The rationale being the purpose of this patch is to force a safe setting unless the user explicitly authorizes to use the value in the config file.  The drop-down menu is not for choosing all possible values of script-security  --- config file is the place for that.  Choices inconsistent with the config is only going to cause an error at runtime.

cron2:
Hi,

On Thu, Feb 07, 2019 at 03:45:57PM +0000, Selva Nair wrote:
> I'm in the process of updating this PR like this:
> 
> Minimize the dialog about script security setting to just two options: (i) built-in commands only (recommended) and (ii) Use config file setting (potentially unsafe).
> 
> The rationale being the purpose of this patch is to force a safe setting unless the user explicitly authorizes to use the value in the config file.  The drop-down menu is not for choosing all possible values of script-security  --- config file is the place for that.  Choices inconsistent with the config is only going to cause an error at runtime.

Sounds reasonable to me.

gert

-- 

cron2:
Hi,

On Mon, Jan 21, 2019 at 09:41:27AM -0800, Selva Nair wrote:
> Options: for configs is in system-wide location, do not enforce a safe script security OR white-list "--script-security" as an option that any user can override. I prefer the latter as on Windows admins are often lay users who could unwittingly install insecure configs in the central location.

Just for cross-reference - the "white list --script-security" patch was
merged to master and release/2.4, so this particular TODO is covered.

gert

-- 

selvanair:
Rebased to master with some changes:

1. script-security setting in the menu reduced to two choices: "use config file setting (unsafe)" or "built-in scripts only (recommended)" and defaults to the latter as before.

2. Added a commit that saves the last fatal error from OpenVPN and appends it to the message box shown when a connection fails. In case of script error if the setting is overridden by the GUI the message is reworded as a localizable string that defaults to:

`External scripts are disabled for this connection as a security measure.  If you must allow such scripts, open the connection specific "Options"menu and select to use config-file setting for "Allow script execution". Note that this is potentially unsafe unless you trust the origin of this connection profile.`

For this to work properly OpenVPN requires patch 684 (https://patchwork.openvpn.net/patch/684/).

selvanair:
The more I test this, the more I dislike it.  Running a config with a script is no longer easy:

TLDR: there is a question at the end.

1. Start with a config that has "--up xxx.bat", but no --script-security. Connection fails and GUI shows a message that points the user to the config-specific options menu

2. User goes to options menu and chooses "Use config file setting (unsafe)" and tries to connect again

3.  Connection fails with an error saying --script-security >= 2 is required

4. Those few who manages to figure out the reason for that edits the config and adds --script-security 2 and tries to connect again

5. Connection fails with the GUI showing a message about safe script-security enforced and points the user to the Options menu

6. Most give up at that point. A few patient ones go to the Options menu and once again reverts the allow script execution to "use config file setting" and tries to connect again.
This time the connection succeeds.

Eventually everyone discovers the global "do not override script-security" option and enables that and happily tunnels ever after, oblivious of scripts infiltrating their configs.

How can we make this less frustrating and still safe?

selvanair:
Amended the last commit to match the proposed error message from OpenVPN in [patch #688](https://patchwork.openvpn.net/patch/688/)

cron2:
Hi,

On Wed, Feb 20, 2019 at 09:06:15AM -0800, Selva Nair wrote:
> The more I test this, the more I dislike it.  Running a config with a script is no longer easy:
> 
> TLDR: there is a question at the end.
> 
> 1. Start with a config that has "--up xxx.bat", but no --script-security. Connection fails and GUI shows a message that points the user to the config-specific options menu
> 
> 2. User goes to options menu and chooses "Use config file setting (unsafe)" and tries to connect again
> 
> 3.  Connection fails with an error saying --script-security >= 2 is required

Maybe we can make this easier by pointing both at the config file and 
the gui setting?

Like

   "Your config tries to run external scripts.  This is disallowed for 
    security reasons.  To allow it, the config file must contain the
    option 'script-security' with a value >= 2 AND it must be allowed 
    by permitting scripts in the GUI - which should never be done for
    configs received from an untrusted source, as malicious scripts
    run from an openvpn config can seriously harm your computer.  If
    the config file is changed, the GUI setting will revert to its 
    default setting and must be re-enabled"

Besides this, I couldn't come up with a more user-friendly way
that does not involve "parsing the config file" from the GUI.

Maybe we need to do this, at least for trivial cases (no nested
configs)?

Mmmmh.

What if we extend openvpn to have a "--disallow-all-scripts" setting 
that will override --script-security but retain the original value, so 
the error message (from your new patch :-) ) can print out 

  "--script-security in config has sufficient level, but scripts are
   disallowed by the GUI"

vs. our current

  "disallowed by script-security setting"

error message for "--script-security in the config is not sufficient"?


(Yes, a new option.  Maybe we need to return to David's original idea
about "--security-profile <dummy|user|expert>" which can affect a number 
of possibly-dangerous config settings at the same time?)

gert
-- 

selvanair:
Parsing the config is in my TODO as its more generally useful.

I guess the fundamental problem is requiring settings in two different places to match one of which automatically changes in some non-intuitive ways. So, how about always set script-security from the GUI? That is, allow the user to choose all valid script-security settings from the GUI (as I had in a previous version, but no option to use the setting in the config file). And always pass --script-security on the command line. Then the user has to change this only in one place (in the GUI) and whatever is in the config file is never used. Unless globally disabled using the disable-script-security-override checkbox in the main settings.

Whenever the config file is edited we revert the script-security value back to 1. We could show that detail as a tool-tip for the dialog item and also could show a message when its reverted so that the user can edit the setting and then connect or go ahead as is.

Would such an improvement be worth the trouble..?

cron2:
Hi,

On Thu, Feb 28, 2019 at 11:12:59AM -0800, Selva Nair wrote:
> Parsing the config is in my TODO as its more generally useful.
> 
> I guess the fundamental problem is requiring settings in two
> different places to match one of which automatically changes in
> some non-intuitive ways. So, how about always set script-security
> from the GUI? That is, allow the user to choose all valid script-security
> settings from the GUI (as I had in a previous version, but no option
> to use the setting in the config file). And always pass --script-security
> on the command line. Then the user has to change this only in one
> place (in the GUI) and whatever is in the config file is never used.
> Unless globally disabled using the disable-script-security-override
> checkbox in the main settings.

I hear what you're saying, but I'm not sure this is helpful - that way,
a GUI user needs to decide, on an .ovpn file given to him by his admin,
which value exactly is the right one.  "Do I need to select 2 or 3
here?  why?"...

Maybe with more explanation what the impact is ("no scripts", "scripts",
"scripts with full access to password data")... so users have an idea
what the effect is.   Yes, this might work :-)

gert

-- 

selvanair:
Yeah, I assumed users have already graduated from OpenVPN U :) But anyway, overriding OpenVPN options can work effectively only if we  expect some level of knowledge on part of the users. Unless we just want to ensure that we don't run any scripts without explicit consent from the user but do not care much whether that was informed consent or not..

Back to the drawing board, it seems.

selvanair:
Based all the discussion this needs re-work and I haven't found time for it. For now withdrawing the PR to reduce clutter..

Issue #281, opened 2018-11-21
Insufficient permissions to route the traffic


gjf:
Latest OpenVPN client, Windows 10 Pro x64 1809.

After install with default settings when running with service the client cannot perform routing:
Wed Nov 21 14:20:51 2018 open_tun
Wed Nov 21 14:20:51 2018 TAP-WIN32 device [Ethernet 2] opened: \\.\Global\{F3029CBA-57CA-44D1-88F6-8E5CC53220A7}.tap
Wed Nov 21 14:20:51 2018 Successful ARP Flush on interface [24] {F3029CBA-57CA-44D1-88F6-8E5CC53220A7}
Wed Nov 21 14:20:51 2018 Blocking outside dns using service succeeded.
Wed Nov 21 14:20:56 2018 env_block: add PATH=C:\WINDOWS\System32;C:\WINDOWS;C:\WINDOWS\System32\Wbem
Wed Nov 21 14:20:56 2018 ERROR: Windows route add command failed: returned error code 1
Wed Nov 21 14:20:56 2018 env_block: add PATH=C:\WINDOWS\System32;C:\WINDOWS;C:\WINDOWS\System32\Wbem
Wed Nov 21 14:20:56 2018 ERROR: Windows route add command failed: returned error code 1
Wed Nov 21 14:20:56 2018 env_block: add PATH=C:\WINDOWS\System32;C:\WINDOWS;C:\WINDOWS\System32\Wbem
Wed Nov 21 14:20:56 2018 ERROR: Windows route add command failed: returned error code 1
Wed Nov 21 14:20:56 2018 env_block: add PATH=C:\WINDOWS\System32;C:\WINDOWS;C:\WINDOWS\System32\Wbem
Wed Nov 21 14:20:56 2018 ERROR: Windows route add command failed: returned error code 1
Wed Nov 21 14:20:56 2018 env_block: add PATH=C:\WINDOWS\System32;C:\WINDOWS;C:\WINDOWS\System32\Wbem
Wed Nov 21 14:20:56 2018 ERROR: Windows route add command failed: returned error code 1
Wed Nov 21 14:20:56 2018 Initialization Sequence Completed

When running client as admin everything works.

cron2:
Hi,

On Wed, Nov 21, 2018 at 03:28:03AM -0800, gjf wrote:
> Latest OpenVPN client, Windows 10 Pro x64 1809.
> 
> After install with default settings when running with service the client cannot perform routing:
[...]

PR #282, opened 2018-11-22
Ignore pushed --route-method when interactive service is in use

selvanair:
When using interactive service, route addition should use the
service. The user may not have privileges to set routes
otherwise.  We already override any --route-method set in the
config file as openvpn.exe is started with --msg-channel as the
last option which sets route-method to ROUTE_METHOD_SERVICE.

This patch extends that to pushed --route-method

Also change _T("") to L"" in the edited lines to be explicit about
wide and narrow strings. We no longer support non-unicode builds.

No change when interactive service is not used.

Ref: issue #281

### DIFF
--- a/openvpn.c
+++ b/openvpn.c
@@ -1931,8 +1931,13 @@ StartOpenVPN(connection_t *c)
 
         c->hProcess = NULL;
         c->manage.password[sizeof(c->manage.password) - 1] = '\n';
-        _sntprintf_0(startup_info, _T("%s%c%s%c%.*S"), c->config_dir, _T('\0'),
-            options, _T('\0'), sizeof(c->manage.password), c->manage.password);
+
+        /* Ignore pushed route-method when service is in use */
+        const wchar_t *extra_options = L" --pull-filter ignore route-method";
+        size += wcslen(extra_options);
+
+        _sntprintf_0(startup_info, L"%s%c%s%s%c%.*S", c->config_dir, L'\0',
+            options, extra_options, L'\0', sizeof(c->manage.password), c->manage.password);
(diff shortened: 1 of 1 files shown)


cron2:
Hi,

On Wed, Nov 21, 2018 at 08:20:16PM -0800, Selva Nair wrote:
> When using interactive service, route addition should use the
> service. The user may not have privileges to set routes
> otherwise.  We already override any --route-method set in the
> config file as openvpn.exe is started with --msg-channel as the
> last option which sets route-method to ROUTE_METHOD_SERVICE.
> 
> This patch extends that to pushed --route-method
> 
> Also change _T("") to L"" in the edited lines to be explicit about
> wide and narrow strings. We no longer support non-unicode builds.
> 
> No change when interactive service is not used.

Looks good.  So ACK from me.

I'm wondering a bit about overflowing startup_info though - it's 
big enough today (1024), but I see no explicite check.  

Is this done implicitely in _sntprintf_0()?

gert
-- 

selvanair:
Thanks.
Yes, _sntpritnf_0 is a wrapper that will not overwrite beyond _countof( startup_info), so there is no buffer overflow concern.  Still the data has to fit into those 1024 characters for correct operation which is the case so far.

cron2:
Hi,

On Thu, Nov 22, 2018 at 05:46:28AM -0800, Selva Nair wrote:
> Thanks.
> Yes, _sntpritnf_0 is a wrapper that will not overwrite beyond _countof( startup_info), so there is no buffer overflow concern.  Still the data has to fit into those 1024 characters for correct operation which is the case so far.

Thanks for the explanation.  So, ACK :-)

gert

-- 

PR #285, opened 2018-12-06
Update English and German language files

chaosflaws:
- synchronized the german resource files with the english version
- corrected hyphens that should have been En dashes


mattock:
@cron2 or some other German speaker: do the changes look reasonable?

mattock:
This is also slightly old, maybe some other strings are needed as well...

schwabe:
Looks good to me

PR #295, opened 2019-03-05
Optionally allow overwrite when importing a config

selvanair:
- Prompt the user for permission if import may overwrite
  an existing config.

- Also raise an error if the import file source matches the
  global or local config directory. Reimporting a config on to
  itself is not supported. This also  avoids ERROR_SHARING_VIOLATION
  in CopyFile() when source and destination are the same.

### DIFF
--- a/main.c
+++ b/main.c
@@ -761,6 +761,14 @@ ImportConfigFile()
         TCHAR destination[MAX_PATH];
         PTCHAR fileName = source + fn.nFileOffset;
 
+        /* check if the source points to the config_dir */
+        if (wcsncmp(source, o.global_config_dir, wcslen(o.global_config_dir)) == 0
+            || wcsncmp(source, o.config_dir, wcslen(o.config_dir)) == 0)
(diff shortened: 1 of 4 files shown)


selvanair:
Can someone take a look at this (or give it a test)? This allows updating configs using the import menu -- currently no possible because we do not allow overwriting.

philomory:
I would love to see this PR merged, is anyone able to take a look at this?

selvanair:
> Could you rebase this on top of master and add following lines
> 
> IDS_NFO_IMPORT_OVERWRITE "A config named ""%s"" already exists. Do you want to replace it?"
> IDS_ERR_IMPORT_SOURCE "Cannot import file <%s> as it is already in the global or local config directory"
> 
> to **openvpn-gui-res-en-msvc.rc** ?
> 
> Otherwise looks good.

Done, and force pushed. The only change is addition of the above to openvpn-gui-res-en-msvc.rc.

PR #296, opened 2019-03-08
travis-ci: switch to xenial image

chipitsine:


chipitsine:
@selvanair 

as for openssl-1.1.0j:

https://travis-ci.org/chipitsine/openvpn-gui/jobs/504097256#L1145-L1157

I guess some patch like https://github.com/OpenVPN/openvpn-build/blob/master/generic/patches/openssl-1.1.0j-issue7653.patch should be applied. I will have a look later

chipitsine:
I will have a look at travis-ci cache tomorrow

chipitsine:
@selvanair , I decided to remove cache. I could not figure out how to make it work again

chipitsine:
@selvanair , I thought the "while" loop was

```
            while (parent)
            {
               ...
                parent = PARENT_GROUP(parent);
                CheckMenuItem(parent->menu, pos, MF_BYPOSITION | MF_CHECKED);
            }
```


really, cppcheck did strange thing here.
xenial comes with cppcheck-1.72, in turn cppcheck-1.87 does it better.

so, I disabled cppcheck for a "while". I'm done with this PR

sure, please provide patches for other errors found

.travis.yml
```diff
@@ -38,7 +38,3 @@ script:
 # build from tarball
     - make distdir=dist dist && tar xvf dist.tar.gz && cd dist
     - ./configure --prefix=/ --libdir=/lib --host=$CHOST --build=x86_64-pc-linux-gnu --program-prefix='' $CONFIG && make
-    
-cache:
-  directories:
-  - download-cache
```
selvanair: Removing the cache will make travis runs take much longer as it has to build openssl each time. As it is travis has been failing often lately because of timeouts and such errors unrelated to our code base. So I think we should try harder to get the cache working again.

chipitsine: we do have just "download" cache. I added it back.

you are right, we can implement "build" cache as well. I will do it in future PR

chipitsine:
@selvanair , thanks!
I will prepare "openssl build caching" feature soon

Issue #216, opened 2018-02-08
Localization not working


ohmer1:
I configured OpenVPN-GUI to use the french localization in the settings, but the apps is still all in english. I'm using version 11.9.0 (bundled with openvpn 2.4.4) on Windows 10.

selvanair:
What do you see if you go back to the settings menu after changing the language? In case of french, you should see group box captions like "User Interface" has changed to "Interface utilisateur" etc. If not, what language shows  up in the "Language:" or "Langue:" box?

Note that messages in the status window will still be in English as openvpn.exe's logs are not localized. Only the GUI dialogs and menus will appear in the chosen language.

ohmer1:
If I change it to english and then back in french, the app switch to french.

I said I set the language to french in the settings, but now I think about it, the language was set to french by default (probably because my OS is configured with the french language) but the settings was not applied until I change it manually.

I replicated the issue on two different computers.

selvanair:
Its supposed to work without need for manually switching back and forth. In any case what the settings show must match what is used by the rest of the program, so what you see is unexpected. Will try to reproduce.

mcornella:
I can confirm. This is a screenshot taken right after installing OpenVPN 2.4.7 (the latest build)
![default lang screenshot](https://user-images.githubusercontent.com/1441704/54914938-69797680-4ef6-11e9-8a1b-14613efffc31.png)

As you can see, the language selected is the right one, but there is no `ui_language` entry under `HKCU\Software\OpenVPN-GUI`, so the UI language doesn't pick the correct value.

After selecting another language and going back to Spanish it does set the `ui_language` entry and the UI shows the correct language:
![spanish lang screenshot](https://user-images.githubusercontent.com/1441704/54915196-e4db2800-4ef6-11e9-8f65-7de58e1f0ff5.png)

It looks like the `HKCU\Software\OpenVPN-GUI` registry path is created on the first start of the GUI program but `ui_language` is not set.

mcornella:
This is an easy fix: it should be enough to do the work of `SetGUILanguage` in [`UpdateRegistry`](https://github.com/OpenVPN/openvpn-gui/blob/7bfb9508526adcc1878d679452bd6f268ff4e26b/registry.c#L327), `langId` being equal to the returned value of `GetUserDefaultUILanguage()`:
[...]

PR #298, opened 2019-03-26
Fallback to default sublanguage when loading resources

selvanair:
When a resource is not found in user's preferred language, first
try the primary language with SUBLANG set to default before
falling back to English.

See issue #216


cron2:
Explanation makes sense, code looks good. ACK.

PR #304, opened 2019-06-30
Promptly close pipe handles passed to child

selvanair:
Parent keeping the handle to write end of child's stdout will
cause ERROR_BROKEN_PIPE not signalled if/when the child exits.

Also add a wrapper for CloseHandle()

Fixes the GUI process hanging in read from child
if the latter unexpectedly dies due to some error.
Trac #1203


cron2:
Looks good to me, so another ACK.

PR #312, opened 2019-09-05
set focus to challenge when password already filled

e2e8:
- set focus to auth challenge field if it is present and the password has already been filled
- remove unnessary cygwin package mingw64-x86_64-gcc-g++

### DIFF
--- a/BUILD.rst
+++ b/BUILD.rst
@@ -22,7 +22,6 @@ their dependencies. You can install these packages using the standard
 - pkg-config
 - make
 - mingw64-x86_64-gcc-core
-- mingw64-x86_64-g++
 - mingw64-x86_64-openssl
 
 
--- a/openvpn.c
+++ b/openvpn.c
@@ -470,6 +470,10 @@ UserAuthDialogFunc(HWND hwndDlg, UINT msg, WPARAM wParam, LPARAM lParam)
             {
                 SendMessage(GetDlgItem(hwndDlg, ID_EDT_AUTH_PASS), EM_SETSEL, 0, MAKELONG(0,-1));
             }
+            else if (param->flags & FLAG_CR_TYPE_SCRV1)
+            {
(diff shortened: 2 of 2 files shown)


BUILD.rst
```diff
@@ -22,7 +22,7 @@ their dependencies. You can install these packages using the standard
 - pkg-config
 - make
 - mingw64-x86_64-gcc-core
-- mingw64-x86_64-g++
+- mingw64-x86_64-gcc-g++
```
selvanair: As per cygwin package list ( https://cygwin.com/packages/package_list.html#m) this change is correct but why are we requiring c++/g++?  Could anyone using cygwin confirm we need this, and `mingw64-x86_64-gcc-core` is not enough?

Anyway, this chunk is unrelated and should be a separate commit or PR.

openvpn.c
```diff
@@ -470,6 +470,10 @@ UserAuthDialogFunc(HWND hwndDlg, UINT msg, WPARAM wParam, LPARAM lParam)
             {
                 SendMessage(GetDlgItem(hwndDlg, ID_EDT_AUTH_PASS), EM_SETSEL, 0, MAKELONG(0,-1));
             }
+            else if (param->flags & FLAG_CR_TYPE_SCRV1)
+            {
+                SetFocus(GetDlgItem(hwndDlg, ID_EDT_AUTH_CHALLENGE));
+            }
```
selvanair: Looks good.

BUILD.rst
chipitsine: I can check it tomorrow

e2e8: I removed the g++ package and the project rebuilt successfully.

selvanair: Thanks for the follow up.
Yes, please cleanup the first commit (i.e., remove the editing of Build.rst from it) and force push. Then we can merge this.

chipitsine: I confirm, build is ok without g++ installed

Issue #319, opened 2019-11-14
Add --update command to fetch newly added config files


jodel9227:
It would be nice to have the ability to tell the gui to fetch newly added config files. Right now, when the gui is running with an opened connection and we add config files, we can't connect a second connection because it doesn't recognize the .ovpn files until we interact directly with the trayicon.

selvanair:
@jodel9227 Are you referring to use cases like the following?

With an instance of GUI running
(i) copy a new config file (foo.ovpn) to the config folder
(ii) run `openvpn-gui --command rescan`   <-- feature request
(iii) connect foo using `openvpn-gui --command connect foo`

If so, please test the implementation [here](https://github.com/selvanair/openvpn-gui/tree/command-rescan).

jodel9227:
Yes that works like a charm, thanks a lot for the quick response !

One thing though. Would it be possible to make this work for config files located in Local AppData ? Right now multiple connections only work when config files are in Program Files/OpenVPN or Users/user123/OpenVPN. We'd rather not use these locations if at all possible. Thanks !

selvanair:
You can change the user's config directory in the settings menu of the GUI
see: settings->Advanced 
Full paths or variable such as %AppData%\foo\bar are accepted. It defaults to %UserProfile%\OpenVPN\config

The global config diretory is set in HKLM\Programs\OpenVPN\ and can only be changed by an admin. Scroll down the repo landing page and you'll find more details on these (https://github.com/OpenVPN/openvpn-gui)

jodel9227:
Alright everything works fine now, thanks for your help !

jodel9227:
Do you by any chance know when this new command will be included in a release ?

selvanair:
Hopefully in the next release if we get a code review.

jodel9227:
Hello, just getting an update on this. We have customers waiting and would like to give them an ETA. 

Thanks !

selvanair:
@jodel9227 This has been merged and will be in the next release -- both in 2.4.9 and 2.5.0, but I don't have a date for either.

PR #320, opened 2019-11-18
Add '--command rescan' to rescan config folders

selvanair:
Add an new command 'rescan' that may be sent to a running
instance of the GUI to force it rescan the config folders.

Use case: with an instance of the is GUI running, one can
manually copy a config file to the config folder and start
it using "openvpn-gui --command rescan" followed by
"openvpn-gui --command connect foo"

v2: The calls to rebuild config file list and recreate
menus is refactored into a function.

See also: #319

### DIFF
--- a/main.c
+++ b/main.c
@@ -481,6 +481,10 @@ HandleCopyDataMessage(const COPYDATASTRUCT *copy_data)
     {
         ShowTrayBalloon(L"", copy_data->lpData);
     }
+    else if (copy_data->dwData == WM_OVPN_RESCAN)
+    {
+        OnNotifyTray(WM_OVPN_RESCAN);
+    }
     else
     {
         MsgToEventLog(EVENTLOG_ERROR_TYPE,
--- a/tray.c
+++ b/tray.c
@@ -208,6 +208,14 @@ DestroyPopupMenus()
     hMenu = NULL;
 }
 
+/* Rescan config folders and recreate popup menus */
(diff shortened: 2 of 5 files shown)


main.c
```diff
@@ -481,6 +481,10 @@ HandleCopyDataMessage(const COPYDATASTRUCT *copy_data)
     {
         ShowTrayBalloon(L"", copy_data->lpData);
     }
+    else if (copy_data->dwData == WM_OVPN_RESCAN)
+    {
+	OnNotifyTray(WM_OVPN_RESCAN);
```
lstipakov: Indent it a bit off

tray.c
```diff
@@ -264,6 +264,13 @@ OnNotifyTray(LPARAM lParam)
             }
         }
         break;
+
+    case WM_OVPN_RESCAN:
+        /* Rescan config folders and recreate popup menus */
+        DestroyPopupMenus();
```
lstipakov: We already have those 3 functions called in 2 places (WM_RBUTTONUP and WM_LBUTTONDBLCLK) and now we add one more. Can we factor them out into RescanAndRecreate() function?

selvanair: Good point. Will force push with this change as well as fix for the tab vs space error above.

selvanair:
@lstipakov does this look okay now?

lstipakov:
LGTM

PR #321, opened 2019-11-21
msvc: preliminary support

lstipakov:
This adds msvc project files and tiny code changes,
required to build this project with msvc.

Microsoft resource compiler, unlike mingw's windres,
doesn't fully support macros and multi-line strings,
so I had to create a separate resource file for msvc.

### DIFF
--- a/main.c
+++ b/main.c
@@ -31,7 +31,6 @@
 #include <shlwapi.h>
 #include <wtsapi32.h>
 #include <prsht.h>
-#include <pbt.h>
 #include <commdlg.h>
 
 #include "tray.h"
--- a/res/openvpn-gui-res.rc
+++ b/res/openvpn-gui-res.rc
@@ -38,6 +38,9 @@ ID_ICO_CONNECTED     ICON  DISCARDABLE  "connected.ico"
 ID_ICO_CONNECTING    ICON  DISCARDABLE  "connecting.ico"
 ID_ICO_DISCONNECTED  ICON  DISCARDABLE  "disconnected.ico"
 
+#ifdef _MSC_VER
+#include "openvpn-gui-res-en-msvc.rc"
+#else
 #include "openvpn-gui-res-cs.rc"
 #include "openvpn-gui-res-de.rc"
 #include "openvpn-gui-res-dk.rc"
@@ -58,6 +61,7 @@ ID_ICO_DISCONNECTED  ICON  DISCARDABLE  "disconnected.ico"
 #include "openvpn-gui-res-ua.rc"
 #include "openvpn-gui-res-zh-hans.rc"
 #include "openvpn-gui-res-zh-hant.rc"
+#endif
 
 /* Version information and such */
 VS_VERSION_INFO VERSIONINFO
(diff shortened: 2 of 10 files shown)


lstipakov:
ping @selvanair @mattock

res/openvpn-gui-res.rc
```diff
@@ -58,6 +59,9 @@ ID_ICO_DISCONNECTED  ICON  DISCARDABLE  "disconnected.ico"
 #include "openvpn-gui-res-ua.rc"
 #include "openvpn-gui-res-zh-hans.rc"
 #include "openvpn-gui-res-zh-hant.rc"
+#else
+#include "openvpn-gui-res-en-msvc.rc"
+#endif
```
selvanair: I think this would be cleaner if used as
```
#ifdef _MSC_VER
#include "openvpn-gui-res-en-msvc.rc"
#else 
<existing includes>
#endif 
```

main.c
```diff
@@ -31,7 +31,9 @@
 #include <shlwapi.h>
 #include <wtsapi32.h>
 #include <prsht.h>
+#ifndef _MSC_VER
 #include <pbt.h>
```
selvanair: pbt.h is an obsolete header  -- I think we can just remove it. May be as a separate commit?

selvanair:
@lstipakov This should be enough for a quick test build using MSVC. But, if we want to support MSVC builds by users (or for releases) we'll need a script that converts all resource files. Manual conversion is error-prone and hard to review. We cant possibly keep two sets of rc files in sync. 

With that caveat, I'm okay with this as a "preliminary" support. I haven't tested it, but if it works for you, good enough.

res/openvpn-gui-res.rc
lstipakov: agreed

main.c
lstipakov: moved into separate commit

lstipakov:
> @lstipakov This should be enough for a quick test build using MSVC. But, if we want to support MSVC builds by users (or for releases) we'll need a script that converts all resource files. Manual conversion is error-prone and hard to review. We cant possibly keep two sets of rc files in sync.

Well, it should be enough to develop with VS, except anything related to resource files - you can build, run, debug, edit code etc.

I've made a script which removes newlines and joins multiline LTEXT into single line. But VS is also not happy about using macros in LTEXT values, so there is still some work to do.

> With that caveat, I'm okay with this as a "preliminary" support.

Thanks, I have force pushed 2 commits with your suggestions. 

> I haven't tested it, but if it works for you, good enough.

I have also added appveyor script which builds VS project and saves artifact (openvpn-gui.exe).

mattock:
@selvanair does "but if it works for you, good enough" mean we can merge this?

main.c
```diff
@@ -31,7 +31,6 @@
 #include <shlwapi.h>
 #include <wtsapi32.h>
 #include <prsht.h>
-#include <pbt.h>
```
selvanair: Actually PBT macros are now in winuser.h which is included through windows.h. So even if we use them later, we dont need this obsolete header.

PR #327, opened 2020-01-07
Support for OpenVPN3

lstipakov:
This adds optional support for using OpenVPN3 client as an alternative to openvpn2.

Just replacing one client with another will not work:

     - OpenVPN3 doesn't use interactive service, it uses
    "agent" service with completely different protocol. OpenVPN GUI
    needs to talk to agent using HTTP and JSON.

     - OpenVPN3 management interface realtime notifications must be
    explicitly turned on in order for GUI to work.

To enable using openvpn3, one needs to build openvpn-gui with
./configure --enable-ovpn3 option. This also requires json-c library.

To switch betweet openvpn2 and openvpn3, see "OpenVPN Engine"
radiobutton group in Settings -> Advanced dialog.

Management interface - enabled OpenVPN3 client will be added soon
to openvpn3 repo (along with existing "cli" test client). Also agent
service will be opensourced in near future.


mattock:
Finally _open source_ OpenVPN 3 is moving towards becoming a real alternative to OpenVPN 2. With this PR merged there are not that many blockers in front of having an OpenVPN 3-based open source Windows installer. :+1:

cron2:
Hi,

On Wed, Jan 08, 2020 at 11:44:40PM -0800, Samuli Seppänen wrote:
> Finally _open source_ OpenVPN 3 is moving towards becoming a real alternative to OpenVPN 2. With this PR merged there are not that many blockers in front of having an OpenVPN 3-based open source Windows installer. :+1: 

While this sounds very nice, I'm somewhat wondering what the advantage
would be...  on Windows, with wintun merged, OpenVPN 2 should be able
to do whatever OpenVPN 3 can (and with the same speed) - and more, so
"investing quite a bit of work here" to get "less functionality" sounds
like a non-optimal tradeoff.

gert
-- 

mattock:
@cron2 technically you're correct. But getting more people to actually use a fully open source version of OpenVPN 3 is a good thing.

In my opinion the "make OpenVPN 3 a real open source project" part took way too long. Based on GitHub discussions (openvpn3, openvpn3-linux) it is finally gaining some traction. Maybe at some point a non-OpenVPN Inc. developer community will also form around it.

lstipakov:
> on Windows, with wintun merged, OpenVPN 2 should be able to do whatever OpenVPN 3 can (and with the same speed)

The keywords here are **should** and **with the same speed**.

Unfortunately this is not the case at the moment. When testing against openvpn3 backend with kernel acceleration, I got up to 1.8Gbit/s with openvpn3 client and 1.2Gbit/s with openvpn2 client (both wintun). As you see, the difference is significant.

I haven't looked deeper into that - it should be indeed possible to make openvpn2 up to speed with openvpn3. There must be something what ASIO does better than openvpn2 event loop. 

Besides performance, having openvpn3 as an option in community client will give the openvpn3 project larger exposure, which I think it needs.

mattock:
@selvanair any thoughts on the actual changes?

selvanair:
> @selvanair any thoughts on the actual changes?

I haven't got a chance to look at this carefully. Will do.

mattock:
@selvanair if you can do this soonish we could get this out in 2.5.0. Optional OpenVPN 3.x support in the "official" OpenVPN GUI version would be nice imho.

selvanair:
> @selvanair if you can do this soonish we could get this out in 2.5.0. Optional OpenVPN 3.x support in the "official" OpenVPN GUI version would be nice imho.

Everyone is putting me on notice these days, I must be slacking..
I don't have a setup to test this with OpenVPN3 so I can only check that the changes wont break anything. Will do that but I'm not convinced this is ready for merging into master

(i) We won't be distributing the GUI with this enabled in 2.5.0, right?
(ii) Where can one find this "ovpnagent" service referred to here? 

Unless we can provide instructions on how to install OpenVPN3 and the agent, I think we could merge this into a development branch and move to master only when ready for community use.

facboy:
this seems to have atrophied somewhat.  it merges on to master relatively easily (some things have been added to the _Advanced_ dialog so it needs moving around), but it doesn't actually work against openvpn3 3.8-master (ie built off github master).  `ovpnagent` logs a `PAGE NOT FOUND` error.

openvpn.c
```diff
@@ -2123,35 +2184,60 @@ StartOpenVPN(connection_t *c)
...
-        const wchar_t *extra_options = L" --pull-filter ignore route-method";
-        size += wcslen(extra_options);
+            c->hProcess = NULL;
+            c->manage.password[sizeof(c->manage.password) - 1] = '\n';
+
+            /* Ignore pushed route-method when service is in use */
+            const wchar_t* extra_options = L" --pull-filter ignore route-method";
+            size += wcslen(extra_options);
 
         _sntprintf_0(startup_info, L"%ls%lc%ls%ls%lc%.*hs", c->config_dir, L'\0',
```
facboy, line 2217: the indentation on these lines is wrong now.

facboy (commented):
apart from the indentation lgtm.  i suspect there are some line-ending differences with my branch but i'm not sure if that's because there are incorrect line-endings in my branch, or on this branch.

facboy:
lgtm

facboy:
lgtm.  been running your patched ovpn3 and this for the last few days.

selvanair (approved):
I have not tested using OpenVPN3, but the changes look sane, and use of OpenVPN 2.0 is not affected as far as I can see.

lstipakov:
Looks like this is ready for merge (/me looks at @selvanair, @mattock and @cron2).

PR #346, opened 2020-04-24
Support per-monitor DPI scaling.

Wouterdek:
People that combine regular monitors with high DPI or small screens often increase the DPI scaling of only one monitor in Windows. Applications that support this properly get notified by Windows when the window switches screen and rerender using the correct scale. For applications that do not support this, Windows stretches the bitmap of the UI to the correct size. This, however, results in a blurry/pixelated UI.

OpenVPN GUI already supported system DPI awareness, but not per-monitor DPI scaling.
This pull request fixes that. The GUI now scales exactly as you would expect, without any weird layout glitches or blurry UI.

The following changes were made:
- Mark application as PerMonitorV2 aware in manifest file.
- Replace old SHBrowseForFolder folder picker with newer IFileOpenDialog API, which scales better
- Handle WM_DPICHANGED messages to update options.dpi_scale. This is then used to re-layout the status window. For this, dpi_setscale() was moved to options.h so it could be accessed by openvpn.c


Wouterdek:
> Are the COM interfaces used here exposed by mingw headers? If not we'll have to declare them.

Where can I find an example of how to do this exactly?

> No asserts please. 

Fixed.

selvanair:
> > Are the COM interfaces used here exposed by mingw headers? If not we'll have to declare them.
> 
> Where can I find an example of how to do this exactly?

I don't have any examples, but, unless it works out of the box the whole interface will need to be declared. Not worth it unless this adds something substantial. That said, looks like mingw does have the required declarations and the appveyor error may be due to some missing includes.

I suggest to break this into two PRs (i) handle WM_DPICHANGED for per monitor DPI support and (ii) change file open dialogs to use more modern API.  That would make it easier to test and merge this.

For (i) your edited dpi_set_scale now gets options as a parameter for no reason (you are not using it). Actually I would suggest to move it to misc.c/h (not options.c/h) and then the options parameter may have to be passed in. And, isn't necessary to handle WM_DPICHANGED in other dialogs as well, not just for the status window?

For (ii), please test the patch using mingw which is our official way of building the releases to find why the appveyor build is failing.

selvanair:
Please do a rebase to current master (not a merge commit) and force push. It cant be accepted in the current form.

Also as mentioned in my review, include only the per-monitor dpi setting support in this PR as a single squashed commit -- the unrelated folder picking change does not belong here. You can submit the latter as a separate PR if you wish.

Wouterdek:
> For (ii), please test the patch using mingw to find why the appveyor build is failing. Our release process uses mingw builds.

I've fixed the mingw build.

> For (i) your edited dpi_set_scale now gets options as a parameter for no reason (you are not using it). Actually I would suggest to move it to misc.c/h (not options.c/h) and then the options parameter may have to be passed in.

I've moved the function to misc.h/misc.c

> And, isn't it necessary to handle WM_DPICHANGED in other dialogs as well, not just for the status window?

No, Windows provides "Automatic DPI scaling for dialogs" (as quoted from [here](https://docs.microsoft.com/en-us/windows/win32/hidpi/high-dpi-desktop-application-development-on-windows#per-monitor-dpi-scaling-support-by-ui-framework--technology)). Only the parts that aren't scaled automatically need it to work properly. I've tested this and the settings, login, status and file/folder browsers scale properly with the proposed code.

Wouterdek:
Changing the DPI scaling breaks SHBrowseForFolder, which is why I included it.

selvanair:
> Changing the DPI scaling breaks SHBrowseForFolder, which is why I included it.

Okay, then separate into two clean commits (one for DPI support, one for SHBrowseForFolder) and rebase with a clean commit history. PRs that have "merge commits" in your branch don't work with our merge strategy.

Wouterdek:
I've opened a new PR (PR #365) with two clean commits. I'll close this one.

Issue #351, opened 2020-06-09
escaped password even with static challenge. invalid password sent.


macskas:
Hello,

I think there is an issue with the password escaping using static challenge.
I understand that escaping needed on the management interface if you are sending raw passwords.
So for example: auth "Password" "%s"
`testpas"sword` needs to be escaped as `testpas\"sword`

But if you are using challenges your input is already base64 encoded so it cannot contain `"` or `\`. So escaping password inside the base64 would just break the input password.

Please consider fixing unnecessary escapes. But if this is the correct behaviour please also respond so I could make different password validation for windows clients with openvpn-gui.

## case 1 - static challenge enabled (incorrect password)
**relevant options:** [ 'auth-user-pass', 'static-challenge "Enter PIN" 1' ]
**input:**
> username: **testuser**
> password: **testpas"sword**
> static challenge: **123456**

**Server received (from environment variables):**

> **_client: openvpn windows version (for example 2.4.9, 'IV_GUI_VER'='OpenVPN_GUI_11'):_**
> username: `testuser`
> password: `SCRV1:dGVzdHBhc1wic3dvcmQK:MTIzNDU2Cg==` (decoded: `SCRV1:testpas\"sword:123456`)

> **_any other client (linux, iphone, android, macos - tunnelblick):_**
> username: `testuser`
> password: `SCRV1:dGVzdHBhcyJzd29yZAo=:MTIzNDU2Cg==` (decoded: `SCRV1:testpas"sword:123456`)

## case 2 - static challenge disabled (correct password)
**relevant options:** [ 'auth-user-pass' ]
**input:**
> username: **testuser**
> password: **testpas"sword**

**Server received (form environment variables):**

> **_client: openvpn windows version (for example 2.4.9, 'IV_GUI_VER'='OpenVPN_GUI_11'):_**
> username: `testuser`
> password: `testpas"sword`

> **_any other client (linux, iphone, android, macos - tunnelblick):_**
> username: `testuser`
> password: `testpas"sword`

selvanair:
Thanks for reporting this. Looks like a bug in OpenVPN-GUI. Will fix.

selvanair:
Fixed: should be in the next release.

PR #352, opened 2020-06-22
Do not do escape processing on static-challenge password and response

selvanair:
Strings passed to the management interface should escape characters
such as " and \ that have special meaning for the parser.
But, static-challenge password and response are base64 encoded
before passing to the management interface and get literally
transported to the server in that form. Escape processing of
these strings could result in altering the password and/or response.

Reported by: macskas https://github.com/OpenVPN/openvpn-gui/issues/351


cron2:
Looks good to me.  Ripping out chunks of unneccessary and potentially harmful code is always good :-)

So, ACK.

PR #267, opened 2018-05-30
Update openvpn-gui-res-ua.rc

toxi22:

PR #360, opened 2020-09-01
Copy missing resource strings to all language files

selvanair:
- All missing strings are copied from the English version
  which is the default.
- Description of "--command cmd" is also copied from the
  English version where missing.
- Some white-space changes for consistent formatting of all files

  No user-visible changes except for the --help output
  which will now include a description of the "--command cmd"
  option in English when a translation is not available.


ValdikSS:
@zhovner will try to translate untranslated Ukrainian strings first, then adapt more applicable translation from #267.

selvanair:
Rebased after merging 362 and force pushed.

selvanair:
If there are no objections I want to merge this. Without this translators don't see all strings and keeping this in sync as new translations come in is painful as well.

mattock:
@selvanair this look good afaics.

PR #363, opened 2020-09-06
Increase the buffer size for help message

selvanair:
The help message is truncated in some languages with
recent increases in textual information in the
usage string. Eg., see "openvpn-gui --help" with
Dutch selected as the language where the text
is about 2300 characters while the buffer is only 2048.

The window still fits on typical screen sizes, so
just increase the buffer size to 3000.


cron2:
Looks good to me :-)

PR #365, opened 2020-09-08
Support per-monitor DPI scaling and modern filedialogs

Wouterdek:
This pull request replaces PR #346. Provides per-monitor DPI scaling and replaces the old SHBrowseForFolder dialogs with modern IFileOpenDialog alternatives.

### DIFF
--- a/misc.c
+++ b/misc.c
@@ -341,6 +341,20 @@ ForceForegroundWindow(HWND hWnd)
 }
 
 /*
+ * Set scale factor of windows in pixels. Scale = 100% for dpi = 96
+ */
+void
+DpiSetScale(options_t* options, UINT dpix)
+{
+    /* scale factor in percentage compared to the reference dpi of 96 */
+    if (dpix != 0)
+        options->dpi_scale = MulDiv(dpix, 100, 96);
+    else
+        options->dpi_scale = 100;
+    PrintDebug(L"DPI scale set to %u", options->dpi_scale);
+}
+
+/*
  * Check user has admin rights
  * Taken from https://msdn.microsoft.com/en-us/library/windows/desktop/aa376389(v=vs.85).aspx
  * Returns true if the calling process token has the local Administrators group enabled
--- a/options.c
+++ b/options.c
@@ -27,6 +27,7 @@
 #endif
 
 #include <windows.h>
+#include <initguid.h>
 #include <windowsx.h>
 #include <prsht.h>
 #include <stdlib.h>
@@ -34,6 +35,7 @@
 #include <memory.h>
 #include <shlobj.h>
 #include <shlwapi.h>
+#include <combaseapi.h>
 
 #include "options.h"
 #include "main.h"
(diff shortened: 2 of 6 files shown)


options.c
```diff
@@ -441,37 +443,81 @@ GetConnByName(const WCHAR *name)
...
+            if (SUCCEEDED(result))
+            {
+                wcsncpy_s(selected_path, MAX_PATH, path, MAX_PATH);
+                selected_path[MAX_PATH - 1] = 0;
+
+                CoTaskMemFree(path);
 
-    PIDLIST_ABSOLUTE idlist = SHBrowseForFolder (&bi);
+                psi->lpVtbl->Release(psi);
+            }else
```
selvanair, line 505: A minor style thing -- we prefer else to start on a line

options.c
```diff
@@ -441,37 +443,81 @@ GetConnByName(const WCHAR *name)
...
+
+            result = pfd->lpVtbl->GetResult(pfd, &psi);
+            if(SUCCEEDED(result))
+            {
+                result = psi->lpVtbl->GetDisplayName(psi, SIGDN_FILESYSPATH, &path);
+            }
+
+            if (SUCCEEDED(result))
+            {
+                wcsncpy_s(selected_path, MAX_PATH, path, MAX_PATH);
```
selvanair, line 499: Assuming selected path has MAX_PATH space is not safe. Please add a third arg to BrowseFolder and pass _countof(initial_path) from the caller and use it here.

options.c
```diff
@@ -441,37 +443,81 @@ GetConnByName(const WCHAR *name)
...
+        // Set initial path
+        {
+            IShellItem* psi;
+            result = SHCreateItemFromParsingName(initial_path, NULL, &IID_IShellItem, &psi);
+            if (SUCCEEDED(result))
+            {
+                pfd->lpVtbl->SetDefaultFolder(pfd, psi);
+                psi->lpVtbl->Release(psi);
+            }
+        }
```
selvanair, line 482: Same here.

options.c
```diff
@@ -441,37 +443,81 @@ GetConnByName(const WCHAR *name)
...
-    bi.ulFlags = BIF_RETURNONLYFSDIRS | BIF_NEWDIALOGSTYLE;
-    bi.lpfn    = BrowseCallback;
-    bi.lParam  = (LPARAM) initial_path;
+        // Set initial path
+        {
+            IShellItem* psi;
+            result = SHCreateItemFromParsingName(initial_path, NULL, &IID_IShellItem, &psi);
+            if (SUCCEEDED(result))
+            {
+                pfd->lpVtbl->SetDefaultFolder(pfd, psi);
```
selvanair, line 479: SetDefaultFolder is not the right one to use here as it does not force the initial path. Use SetFolder. Otherwise, the dialog will always start with the previous folder which is not what we want as we use this for both config and log locations.

misc.c
```diff
@@ -340,6 +340,20 @@ ForceForegroundWindow(HWND hWnd)
...
+ */
+void
+DpiSetScale(options_t* options, UINT dpix)
+{
+    /* scale factor in percentage compared to the reference dpi of 96 */
+    if (dpix != 0)
+        options->dpi_scale = MulDiv(dpix, 100, 96);
+    else
+        options->dpi_scale = 100;
+    PrintDebug(L"DPI scale set to %u", o.dpi_scale);
```
selvanair, line 354: o.dpi_scale should be options->dpi_scale

options.c
```diff
@@ -441,37 +443,81 @@ GetConnByName(const WCHAR *name)
...
+    {
+        return false;
+    }
 
-static BOOL
-BrowseFolder (const WCHAR *initial_path, WCHAR *selected_path)
-{
-    BOOL ret = false;
-    BROWSEINFO bi;
+    result = CoCreateInstance(&CLSID_FileOpenDialog, NULL, CLSCTX_ALL, &IID_IFileOpenDialog, &pfd);
```
selvanair, line 459: (void **)&pfd is better to avoid compiler warning

options.c
```diff
@@ -441,37 +443,81 @@ GetConnByName(const WCHAR *name)
...
 
-    CLEAR(bi);
-    bi.lpszTitle = L"Select folder...";
-    bi.ulFlags = BIF_RETURNONLYFSDIRS | BIF_NEWDIALOGSTYLE;
-    bi.lpfn    = BrowseCallback;
-    bi.lParam  = (LPARAM) initial_path;
+        // Set initial path
+        {
+            IShellItem* psi;
+            result = SHCreateItemFromParsingName(initial_path, NULL, &IID_IShellItem, &psi);
```
selvanair, line 476: Similarly, (void **)&psi here.

selvanair (changes requested):
Finally I got to build and run-test this. I have only checked that it doesn't break anything, not multiple dpi monitors. Looks good apart from the issues mentioned. Could be merged once those are fixed. Thanks.

lstipakov:
@selvanair can this be merged? /me tested it too. Would be nice to get it into beta4.

selvanair:
This is good now, but now we have two commit with errors and one that corrects them (apart from style changes.) @Wouterdek please squash the changes to their commits and force-push.

misc.c
```diff
@@ -340,6 +340,20 @@ ForceForegroundWindow(HWND hWnd)
...
+ */
+void
+DpiSetScale(options_t* options, UINT dpix)
+{
+    /* scale factor in percentage compared to the reference dpi of 96 */
+    if (dpix != 0)
+        options->dpi_scale = MulDiv(dpix, 100, 96);
+    else
+        options->dpi_scale = 100;
+    PrintDebug(L"DPI scale set to %u", o->dpi_scale);
```
selvanair, line 354: Sorry, his is still not correct -- options->dpi_scale, not o->dpi_scale.

selvanair:
Thanks.

selvanair:
Oh, did I merge it before you got to push the version update? Should we revert, update version and then merge again?

mattock:
@selvanair I can personally live with the fact that the version of the OpenVPN GUI executable and version change in the repo do not match 100%. We'll be pushing out a new release in a week or two anyways, so the problem will solve itself.

Issue #364, opened 2020-09-08
Bug in menu level depth


YouveGotMeowxy:
I am using the latest version:

![image](https://user-images.githubusercontent.com/39491297/92427566-673ec280-f152-11ea-8d7a-80486e872d52.png)

 I have added a bunch of config files under the config folder, and have been organizing them in a coherent way. It appears that if you add too many levels, they just disappear from the menu entirely.

 I have attached my entire folder structure to demonstrate; it just needs to be extracted to the config folder, and then when running openvpn GUI, you can see most of the items don't show in the menu.

[PrivadoVPN.zip](https://github.com/OpenVPN/openvpn-gui/files/5185769/PrivadoVPN.zip)

selvanair:
I had arbitrarily limited the max depth at 4, but most of your configs are at one level deeper than that. I don't want to make it unlimited because of recursive calls, but its easy to make the max depth configurable from in the settings menu.

At some point you will also hit the silly MAX_PATH = 260 characters limit.

YouveGotMeowxy:
> but its easy to make the max depth configurable from in the settings menu.

That would be perfect. :)

selvanair:
I tested by scanning largish folder structures -- performance penalty of recursion is negligible and is naturally limited by the depth of directory trees. The dir tree of my whole C-drive has a max depth of 20. 

Even 10 deep levels of nesting makes the menu very clumsy to navigate, so I will just increase the max depth to 20 or so and leave it at that. Much easier than making this configurable.

PR #366, opened 2020-09-08
Increase the depth of directories scanned for config files

selvanair:
Some users want to use more deeply nested folders than currently
supported. Increase the max depth to 20 which should be plenty for
all.

See also issue: #364


cron2:
Easy enough :)

PR #367, opened 2020-09-14
Use iservice even when gui is running under admin

lstipakov:
We didn't use interactive service when gui was running
under admin because of some privilege escalation vulnerability in Vista.

Apparently this issue doesn't exist on Win7 and newer versions so
it is safe to use iservice on those systems.

Recent versions of mingw (like on Ubuntu18.04 which we use for
building installer) include versionhelpers.h so no need for compat layer.

### DIFF
--- a/main.c
+++ b/main.c
@@ -28,6 +28,7 @@
 #endif
 
 #include <windows.h>
+#include <versionhelpers.h>
 #include <shlwapi.h>
 #include <wtsapi32.h>
 #include <prsht.h>
@@ -263,7 +264,7 @@ int WINAPI _tWinMain (HINSTANCE hThisInstance,
     exit(1);
   }
 
-  if (!IsUserAdmin() && strtod(o.ovpn_version, NULL) > 2.3 && !o.silent_connection)
+  if ((!IsUserAdmin() || IsWindows7OrGreater()) && strtod(o.ovpn_version, NULL) > 2.3 && !o.silent_connection)
     CheckIServiceStatus(TRUE);
 
   BuildFileList();
--- a/openvpn.c
+++ b/openvpn.c
@@ -27,6 +27,7 @@
 
 #include <windows.h>
 #include <windowsx.h>
+#include <versionhelpers.h>
 #include <tchar.h>
 #include <stdlib.h>
 #include <stdio.h>
@@ -1937,7 +1938,7 @@ StartOpenVPN(connection_t *c)
         (o.proxy_source != config ? _T("--management-query-proxy ") : _T("")));
 
     /* Try to open the service pipe */
-    if (!IsUserAdmin() && InitServiceIO (&c->iserv))
+    if ((!IsUserAdmin() || IsWindows7OrGreater()) && InitServiceIO (&c->iserv))
     {
         DWORD size = _tcslen(c->config_dir) + _tcslen(options) + sizeof(c->manage.password) + 3;
         TCHAR startup_info[1024];


cron2:
Hi,

On Mon, Sep 14, 2020 at 11:52:08AM -0700, Lev Stipakov wrote:
> We didn&#39;t use interactive service when gui was running
> under admin because of some privilege escalation vulnerability in Vista.
> 
> Apparently this issue doesn&#39;t exist on Win7 and newer versions so
> it is safe to use iservice on those systems.

I do agree with this in principle.

OTOH, I found it very useful in my win10/tap/netbits tests to be able
to run "with iservice" and "without iservice" at will...  so how would
I do this in future?  "Run the GUI as admin *and* stop the iservice"?

(That would be good enough)

gert


-- 

lstipakov:
> how would I do this in future? "Run the GUI as admin *and* stop the iservice"? (That would be good enough) 

Yeah. In this case GUI will prompt "Interactive service is not started, tasks requiring administrative access may not work". I wonder if we could make it a bit smarter and if we're running on Win7 or newer we prompt "wintun will not work" instead.

selvanair:
Hi,

I would suggest we add an option in settings to always use iservice and
enable it by default.

While I agree connecting to iservice always is fine on win7+, I don't like
running the GUI as admin without at least a warning. There are so many
places out there that mistakenly advise that the OpenVPN-GUI must be
started as admin (left-over from 2.3 days). This will only increase the
number users running it as admin for the wrong reasons. But if this is what
it takes to get users stop complaining, I will grudgingly accept this.

Selva

selvanair:
> > how would I do this in future? "Run the GUI as admin _and_ stop the iservice"? (That would be good enough)
> 
> Yeah. In this case GUI will prompt "Interactive service is not started, tasks requiring administrative access may not work".

No, it will not show that popup if the GUI is started as admin.

> I wonder if we could make it a bit smarter and if we're running on Win7 or newer we prompt "wintun will not work" instead.

We should change the logic of that popup to only check for the service and print that wintun will not work and if on top of that not admin also include the current wording. I think the corner case of vista may be ignored as its not officially supported any longer.

lstipakov:
> No, it will not show that popup if the GUI is started as admin.

It will, assuming this PR is merged:

```
  if ((!IsUserAdmin() || IsWindows7OrGreater()) && strtod(o.ovpn_version, NULL) > 2.3 && !o.silent_connection)
    CheckIServiceStatus(TRUE);
```

(me stares at that popup right now)
 
> We should change the logic of that popup to only check for the service and print that wintun will not work and if on top of that not admin also include the current wording.

Yeah, will do.

lstipakov:
Changes required:

 - instead of unconditionally use iservice on >= Win7, have an option "use iservice even when running as admin" which is enabled by default

 - if iservice is not running,
   * if running as admin, prompt about non-working wintun 
   * if running as non-admin, use existing prompt and also mention wintun

@selvanair correct?

selvanair:
I was referring to the current behaviour regarding the popup

> Changes required:
> 
> * instead of unconditionally use iservice on >= Win7, have an option "use iservice even when running as admin" which is enabled by default
> * if iservice is not running,
>   
>   * if running as admin, prompt about non-working wintun
>   * if running as non-admin, use existing prompt and also mention wintun
> 
 
That covers all except the corner case in vista and sounds good to me. Show only if silent-connection is not on as we do now.

FWIW, we could put the test of IsWindows7OrGreater first as IsUserAdmin() is likely more expensive.

lstipakov:
Superseded with #368

PR #375, opened 2020-10-30
Put the output of the connect script to a log file

ffes:
This PR puts the output of the connect script `connection_up.bat` to a logfile named `connection_up.log` next to the other logs.

For now I only implemented this in the connect script, but obviously this needs to be done for the preconnect and disconnect script. I will add that code once the code I pushed so far is reviewed.

### DIFF
--- a/scripts.c
+++ b/scripts.c
@@ -226,16 +226,31 @@ RunDisconnectScript(connection_t *c, int run_as_service)
     if (!run_as_service)
         SetDlgItemText(c->hwndStatus, ID_TXT_STATUS, LoadLocalizedString(IDS_NFO_STATE_DISCONN_SCRIPT));
 
+    // Create the filename of the logfile
+    TCHAR script_log_filename[MAX_PATH];
+    _sntprintf_0(script_log_filename, _T("%s\\%s_down.log"), o.log_dir, c->config_name);
+
+    // Create the log file
+    SECURITY_ATTRIBUTES sa;
+    CLEAR(sa);
+    sa.nLength = sizeof(SECURITY_ATTRIBUTES);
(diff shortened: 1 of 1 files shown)


selvanair (commented):
Looks useful to me. Right now to get the o/p of scripts one has to use `config_up.bat` as a wrapper that executes the actual script with `> output_file 2>&1`  

For completeness let's do the same for pre and down scripts too.

ffes:
> Right now to get the o/p of scripts one has to use `config_up.bat` as a wrapper that executes the actual script with `> output_file 2>&1`

That is exactly the reason why I implemented this. I needed to debug a faulty script a while ago and added the redirects to every simple line in my script.

ffes:
> And, use MAX_PATH instead of 256.

I changed it in my new added commits but not in the existing code in this file. Do you want me to change those 3 existing lines as well, now that we're on it?

```c
TCHAR cmdline[256];
```

selvanair:
> > Right now to get the o/p of scripts one has to use `config_up.bat` as a wrapper that executes the actual script with `> output_file 2>&1`
> 
> That is exactly the reason why I implemented this. I needed to debug a faulty script a while ago and added the redirects to every simple line in my script.

Instead of redirecting every line use config_up.bat as a wrapper with a single line: script.bat > script.log 2>&1 where the actual script content is moved away to script.bat. But, agreed, let's provide a built-in solution.

selvanair:
Thanks for the changes. Looks good to me. 
Please squash the first two commits and force-push -- we like to keep the history clean. Then we can merge this.

ffes:
> Please squash the first two commits and force-push -- we like to keep the history clean.

Just did that

PR #383, opened 2020-12-01
Only change to reconnecting when already connected

### LINKED ISSUE #177: connect script is not executed on reconnect
i'm not sure if this is a bug or as designed. if intentional, i'd like to make a RFE to execute the connect script on reconnect.

i think modifying the following should suffice:
openvpn.c line 198:
```
/* Run Connect Script */
if (c->state == connecting || c->state == resuming || c->state == reconnecting)
    RunConnectScript(c, false);
```

ffes:
As discussed in #177 a fix to only change the state to reconnecting when already connected

Fixes #177

### DIFF
--- a/openvpn.c
+++ b/openvpn.c
@@ -324,12 +324,19 @@ OnStateChange(connection_t *c, char *data)
                 c->failed_psw_attempts++;
         }
 
-        c->state = reconnecting;
-        CheckAndSetTrayIcon();
+        // We change the state to reconnecting only if there was a prior successful connection.
+        if (c->state == connected)
+        {
+            c->state = reconnecting;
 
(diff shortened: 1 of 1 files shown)


openvpn.c
```diff
@@ -324,12 +324,19 @@ OnStateChange(connection_t *c, char *data)
                 c->failed_psw_attempts++;
         }
 
-        c->state = reconnecting;
-        CheckAndSetTrayIcon();
+        // We cannot interpret the state change to RECONNECTING as an indication of a restart after a prior successful connection
```
selvanair, line 327: A minor remark: Its more useful to phrase comments as applying to the new code, not to the old code. Like: We change the state to reconnecting only if there was a prior successful connection.

But I won't insist -- the code change looks good to me.

ffes: I've update the comment, squashed and force pushed

Issue #372, opened 2020-10-23
Add a global "Disconnect" menu item to context menu?


YouveGotMeowxy:
![image](https://user-images.githubusercontent.com/39491297/96944347-374c4400-14a0-11eb-81e0-910027e29f8a.png)

After (thankfully :) [increasing the level depth](https://github.com/OpenVPN/openvpn-gui/issues/364), there's another issue; whenever I need to disconnect, or switch to a different vpn, I have to drill down 7 levels to click the Disconnect menu item for that specific vpn.

Would it be possible to just add a disconnect item to the main context menu, that when clicked, simply disconnects any connected vpn? It'd be sooooo much easier. 

Like, in the screenshot, below the separator bar add "Disconnect", then another sep bar below it, and then the rest.

thanks!
Kacey

selvanair:
Thanks for the suggestion.
If you have only one connected profile, just double left-click on the tray icon to open the status window from where you can disconnect. We could extend that behaviour to cover multiple connections (within a reasonable number). That would provide an easy way to review all connected tunnels or to disconnect/reconnect any. Could be more useful than a "disconnect all".

PR #388, opened 2020-12-30
Open all active status windows on left-double-click

selvanair:
Currently we pop up the status window on double click only
if one connection is active though there is no strong reason to
limit this behaviour. In fact, when multiple connections are
stuck in the connecting state, its very useful to have a quick
way to examine their progress instead of having to drill down
the menu. Especially so when nested menu is in use.

A random variation of up to 100 pixel is added to the initial
position of the status window to avoid all windows falling on
top of each other.

To prevent an explosion of new windows in the very unlikely event
of numerous active connections, restrict the maximum windows
shown to 10.

Also see discussion under #372

### DIFF
--- a/openvpn.c
+++ b/openvpn.c
@@ -1662,6 +1662,12 @@ StatusDialogFunc(HWND hwndDlg, UINT msg, WPARAM wParam, LPARAM lParam)
 
         /* Set size and position of controls */
         RECT rect;
+        GetWindowRect(hwndDlg, &rect);
+        /* Move the window by upto 100 random pixels to avoid all
+         * status windows fall on top of each other.
+         */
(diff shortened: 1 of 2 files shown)


cron2:
Hi,

On Tue, Dec 29, 2020 at 10:37:38PM -0800, Selva Nair wrote:
> Currently we pop up the status window on double click only
> if one connection is active though there is no strong reason to
> limit this behaviour. In fact, when multiple connections are
> stuck in the connecting state, its very useful to have a quick
> way to examine their progress instead of having to drill down
> the menu. Especially so when nested menu is in use.

We have a double-click option?  How nice :-)

This definitely gets a Feature-ACK.  Will have a look at the code
ASAP.

gert
-- 

lstipakov (approved):
Compiled and tested - works as expected.

PR #389, opened 2020-12-30
Parse and display echo messages

selvanair:
- Messages are added to the same window which is popped up when a new message arrives if minimized or hidden. The latest message appears at the top. This simplifies widow life-cycle management and allows the user to review older messages.

The following will be covered by another PR 
- Localization of message window title and show the new options (disable_popup_messages and popup_mute_interval) in help
- Add  popup mute interval and option to disable of messages to the settings menu
- Automatic resizing of message window to fit the latest message
~- Add profile name and timestamp to the message~ Done
- Periodically truncate the message window buffer


selvanair:
Here is a sample image of the echo message window

msg-window

chipitsine:
A lot of text make people frustrate. Can we put a little of text + big red button ?

Big red button is so called "call to action", so people know what we want them to do (I.e. " your certifacate is about to expire... Press me to renew it")

selvanair:
Agreed, short messages should be preferred. As messages are sent by the server, admins get to decide the verbosity. Sending a header only message is also supported.

chipitsine:
Can it render a button with http link?

selvanair:
> Can it render a button with http link?

Right now no, because the proposed echo-msg syntax has no facility for indicating one. But it can display URLs as clickable links. 

It just displays what is received from the server -- add syntax on how to indicate a button or an image and we can support it.

cron2:
Hi,

I've read through all the patches currently in the set.

Everything that is "not related to painting windows" looks very reasonable
(ACK).  The windows stuff does not look "dangerous" (like, memory leaks
or anything) though I can't say if the "painting" bits are correct -
not enough windows know-how.

I'd go forward with it, so it can get exposure and remaining bugs shaken
out.

On the clickable URL stuff - that's nice magic :-)

gert
-- 

PR #390, opened 2020-12-30
Show a mark against profiles that are in connecting state

selvanair:
Currently we show a check mark on connected profiles with no
indication on profiles that may be in the connecting state.

Change this by adding a mark against connecting/reconnecting
profiles. The yellow connecting state icon is used to generate
this mark although a custom designed check mark may look better. 

In case of nested configs, the parent menus are marked with a tick
mark and only the profile is marked with the connecting icon.

Eg., here is one connected and one connecting profile each.

checkmark

No change in behaviour for profiles that are connected or
disconnected.

Trac #1241

### DIFF
--- a/tray.c
+++ b/tray.c
@@ -45,11 +45,31 @@ HMENU hMenu;
 HMENU hMenuConn[MAX_CONFIGS];
 HMENU hMenuService;
 
+HBITMAP hbmpConnecting;
+
 NOTIFYICONDATA ni;
 extern options_t o;
 
 #define USE_NESTED_CONFIG_MENU ((o.config_menu_view == CONFIG_VIEW_AUTO && o.num_configs > 25)   \
                                  || (o.config_menu_view == CONFIG_VIEW_NESTED))
+
+
+/* Create menu check bitmaps */
+static void
+CreateBitmaps()
+{
+    if (hbmpConnecting)
+        return;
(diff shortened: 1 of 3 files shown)


tray.c
```diff
@@ -45,12 +45,41 @@ HMENU hMenu;
 HMENU hMenuConn[MAX_CONFIGS];
 HMENU hMenuService;
 
+HBITMAP hbmpConnected;
```
lstipakov, line 48: This is not used with `SetMenuItemBitmaps()`. We only assign it in `CreateBitmaps()`.

lstipakov (commented):
New global variable `hbmpConnected` is not used and indentation in new code is a bit off - see tray.c:500.

tray.c
selvanair: > This is not used with `SetMenuItemBitmaps()`. We only assign it in `CreateBitmaps()`.

Right, I had considered using that image in place of the default check mark when connected. Removed & fixed the spacing errors (tabs).
Thanks.

lstipakov (approved):
Next step would be to draw icon with alpha transparency :)

selvanair:
> Next step would be to draw icon with alpha transparency :)
It should be possible to combine the mask in the icon file with the image, but I took the easy way out. Let's see, someone may contribute a custom designed image.

Issue #387, opened 2020-12-29
Use ephemeral management port instead of fixed port + offset


leechristensen:
I have a third-party app installed that often opens a series of ports in the port range 25000-26000 range (weird behavior, I know). I noticed that when it was running, I was unable to connect to the VPN.  I narrowed it down and found that the culprit is that openvpn-gui does not use an ephemeral port when starting, [it instead uses an offset from the port 25340](https://github.com/OpenVPN/openvpn-gui/blob/3fb55cba60c306255eea06b15145ad0c43d4e4cc/openvpn_config.c#L107).

My request is that it change to use an ephemeral port rather than a fixed port + offset. Thanks!

selvanair:
This is harder than it looks. One option is to change the logic to make the GUI listen on a random port and ask OpenVPN core to connect to it. Such a mode is supported by OpenVPN  but would require considerable changes in the GUI logic.

Another option is to ask openvpn to use a dynamic port but then getting that port number conveyed to the GUI would require a side channel.

What we can do easily is to make the port offset (25340) customizable by the user.

cron2:
Hi,

On Wed, Dec 30, 2020 at 03:58:41PM -0800, Selva Nair wrote:
> What we can do easily is to make the port offset (25340) customizable by the user.  

I think this would be a suitable and "quickly done" workaround for those
rare occasions where it conflicts (I think I've seen two problem reports
relating to this, in about 10 years...)

gert

-- 
"If was one thing all people took for granted, was conviction that if you 
 feed honest figures into a computer, honest figures come out. Never doubted 
 it myself till I met a computer with a sense of humor."
                             Robert A. Heinlein, The Moon is a Harsh Mistress

Gert Doering - Munich, Germany                             gert@greenie.muc.de

leechristensen:
Sounds like a configuration option would be the quickest win, and that's fine by me.

> Another option is to ask openvpn to use a dynamic port but then getting that port number conveyed to the GUI would require a side channel.
Rather than having openvpn use a dynamic port, could the GUI application discover dynamic port on it own by temporarily opening a dynamic port, get the port number with `getsockname`, close the port, and then pass the open "discovered" dynamic port in when starting openvpn?

selvanair:
[...]

Issue #370, opened 2020-09-23
It is not easy to use when I have so many many profile


SprayHank:
It is not easy to use when I have so many many profile

![微信截图_20200923161215](https://user-images.githubusercontent.com/2662845/93985978-5760e880-fdb8-11ea-98a8-fdfe6f247016.png)
![微信截图_20200923161527](https://user-images.githubusercontent.com/2662845/93985995-5d56c980-fdb8-11ea-9776-39019b5e00ca.png)
![微信截图_20200923161546](https://user-images.githubusercontent.com/2662845/93985997-5e87f680-fdb8-11ea-9bae-08160aa6c8e2.png)
![微信截图_20200923161608](https://user-images.githubusercontent.com/2662845/93986010-62b41400-fdb8-11ea-9ab7-fbee92e8c39c.png)
![微信截图_20200923161627](https://user-images.githubusercontent.com/2662845/93986025-65166e00-fdb8-11ea-9091-b042a5232651.png)
![微信截图_20200923161642](https://user-images.githubusercontent.com/2662845/93986030-66e03180-fdb8-11ea-8d2f-90fd80470280.png)
![微信截图_20200923161705](https://user-images.githubusercontent.com/2662845/93986032-68115e80-fdb8-11ea-9ae8-f1edef1aebf9.png)

I can only use the down arrow button to scroll it slowly,
Can it display multi-column like below ?

![无题](https://user-images.githubusercontent.com/2662845/93987127-c1c65880-fdb9-11ea-8ca8-f21fb3fec87d.png)

selvanair:
You can organise the profiles into folders and sub-folders to get a nested view. That would help somewhat with so many profiles. 

To do this, create sub directories in the config directory like (for example)

Asia\Japan and add all jp-tok..... and similar profiles in this folder
Asia\Korea\kr-seo.. files etc..
Europe\UK\uk-lon... files  etc.
etc... 

That will generate a hierarchical menu.

LoZio:
I have somewhat the opposite problem.
I have 15/20 config files, they are put in separate directories to be organized under config.
Using GUI 2.4 they all appear as a single menu, so I right clicked the icon and then was ready to connect.
With GUI 2.5 I have to right click, system profile, directory, specific file, connect. Being in the tray bar each menu appears to the right or left depending on its size, ad missing the pixel space moving on the touchpad you have to restart.
[...]

PR #391, opened 2021-01-01
Make 'management port offset' and 'menu view' user-configurable

selvanair:
- Add an option in the advanced settings menu for
the management port offset. Allows any value in the
range 1 to 61000 which with upto ~4000 added as connection
id keeps it in range.

Default is the currently hard coded value of 25340.

As Windows has no concept of privileged ports and the ephemeral
range used varies from version to version, no attempt is made to
avoid conflicts with ports in use.

- Add an option to choose the config menu view from the
advanced settings with three options:

Auto:   Automatically switch to the nested view when
        number of configs exceed a limit (currently 25)
Flat:   Force the flat view irrespective of the number of
        configs
Nested: Force the nested view irrespective of the number
        of configs

Issues: #370 and #387 


selvanair:
PR #391  rebased to master to include UI changes into the recently added Farsi language file.

cron2:
I have skimmed the language files (look consistent to me) and the C code changes (look simple and straightforward enough).

So, ACK from me.

Issue #397, opened 2021-01-21
swprintf %s format for wide strings


OZ1:
I'm trying to build the project using MSYS2 following BUILD.rst setup instruction.
In the resulting binary I get some strings truncated after the first character. The cause is that for unicode strings "%s" formatting is used, which is perfectly fine in MSVC. However MinGW sprintf functions expect "%ls" for wide strings. How did it suppose to work? How do you compile the app to include MSVC-style swprintfs? What steps are missing from BUILD.rst?

OZ1:
Maybe _swnprintf should be used instead?

selvanair:
Our official builds are made using mingw-w64 cross compiler on Linux. The code should work as is without exhibiting any string truncations. What version of mingw-w64?

As mingw targets windows runtime, "%s" in swprintf gets interpreted as corresponding to wide string, and "%S" as narrow string. That's Windows convention that predates C99 which requires %ls and %hs. Read about it here. https://devblogs.microsoft.com/oldnewthing/20190830-00/?p=102823

That said, I don't mind changing %s to %ls and %S (if any) to %hs to be explicit as (I think) that also works on all modern versions of Windows. 

I don't build on Windows, but the instructions in BUILD.rst look right to me. Here is how to cross-compile on Linux using mingw-w64  --- showing the simpler case of building without OpenSSL by disabling the pkey password change feature.

```
autoreconf -iv
make clean 
./configure --host=x86_64-w64-mingw32 --build=x86_64-pc-linux-gnu --disable-password-change
 make
```

OZ1:
it seems that just changing the functions does the job

selvanair:
> it seems that just changing the functions does the job

As I said swprintf should work as is. Please report your compiler version and documentation that says swprintf behaves differently on that version.

OZ1:
I am building the app in MinGW on Windows.

http://www.cplusplus.com/reference/cwchar/swprintf:
Notice that all format specifiers have the same meaning as in printf; therefore, %lc shall be used to write a wide character (and not %c), as well as %ls shall be used for wide strings (and not %s).
Format specifiers are described here: http://www.cplusplus.com/reference/cstdio/printf/

To reuse a build system, I've inserted some lines in main.c:
`int WINAPI _tWinMain (HINSTANCE hThisInstance,
                    UNUSED HINSTANCE hPrevInstance,
                    UNUSED LPTSTR lpszArgument,
                    UNUSED int nCmdShow)
{
[...]

PR #398, opened 2021-01-25
swprintf > _snwprintf

OZ1:
#397 MinGW treats %s in swprintf as a narrow string specifier while in _snwprintf does not


selvanair:
As discussed in #397 

Please follow up the discussion in $397 if you are still facing issues.

> #397 MinGW treats %s in swprintf as a narrow string specifier while in _snwprintf does not

Can you quote a reference for that? We have been using swprintf without any issues and it does work as I explained in #397.

selvanair:
Building with newer mingw (>= version 8)  requires a more through fix than proposed here. See https://github.com/OpenVPN/openvpn-gui/issues/397#issuecomment-767720692

Until then build with -D __USE_MINGW_ANSI_STDIO=0.

Closing.

PR #402, opened 2021-02-24
Update openvpn-gui-res-pt.rc

igorruckert:
Translated untranslated strings e fixed typos.


lstipakov (approved):
My colleague understands Portuguese, looks good in his opinion.

PR #405, opened 2021-02-26
Improve the check for whether global and local config_dir are the same

selvanair:
Extend the string comparison of the two paths to include comparing their
file information structure.  (See Remarks under the MSDN docs for
GetFileInformationByHandle)

If the strings are identical we treat them as identical without checking
further whether the paths are valid. This matches the current
behaviour (except that we do a case-independent comparison). Otherwise, 
the two paths are treated as identical if both exist, are accessible and 
point to the same object in the file system.

Trac: #1359, #1376

### DIFF
--- a/openvpn_config.c
+++ b/openvpn_config.c
@@ -337,6 +337,65 @@ BuildFileList0(const TCHAR *config_dir, int recurse_depth, int group, int flags)
     FindClose(find_handle);
 }
 
+/*
+ * Open a path and get its file information structure.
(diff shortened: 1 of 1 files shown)


cron2:
Looks good to me.  MSDN docs confirm that dwVolumeSerialNumber + nFileIndex{Low,High} are what is needed to uniquely determine "is this the same file or not".

(Unix would do a fstat() and compare inode numbers :-) )

PR #406, opened 2021-03-01
Apply transparency mask to the connecting-state checkmark image

selvanair:
Ref: https://github.com/OpenVPN/openvpn-gui/pull/390#pullrequestreview-560539108
Lack of transparency of the image added in PR 390 kept bugging me too.. So this PR:

Combine the image and mask in the connecting-state icon to form a bitmap
with transparency for use as the checkmark.

MSDN docs on SetMenuItemBitmaps is unclear about the use of
color bitmaps for checkmarks, but this appears to display well.
(Tested on Windows 10 only).

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -108,7 +108,8 @@ openvpn_gui_LDADD = \
 	-lnetapi32 \
 	-lole32 \
 	-lshlwapi \
-	-lsecur32
+	-lsecur32 \
+	-lmsimg32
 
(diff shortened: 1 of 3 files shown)


lstipakov (approved):

selvanair:
> For me it is still not fully transparent (see while background), but much better comparison to what we've had (black background).
> 

Thanks for testing this. I have a fairly white menu background and couldn't see the difference.. Will do more tests to see whether I can improve this. Drawing with DC is not my forte..

PR #407, opened 2021-03-01
Echo msg improvements

selvanair:
4 commits
- document new registry keys
- purge old messages when message window buffer grows large (max ~1000 lines kept)
- add option to disable messages and change mute intervals in the UI (settings menu)
- copy changes to all language files


cron2:
This looks reasonable.  It has conflicts with more recent language file adjustments, it seems...

selvanair:
> This looks reasonable. It has conflicts with more recent language file adjustments, it seems.

Will take that as an approval.  
Resolved minor conflicts due to recent changes and removal of msvc-specific resource file.

PR #409, opened 2021-03-01
Remove MAX_CONFIGS limit using WM_MENUCOMMAND for menu messages

selvanair:
We currently use WM_COMMAND message which is delivered with the
ID of the menu item requiring a unique ID for every command
(connect, disconnect etc..) for each connection profile. Instead,
use WM_MENUCOMMAND so that the message delivers a handle to the
menu and the position index of the menu item.

Connection menu array is now dynamically allocated. Yet, there
is still a limitation on the number of configs as the config
index + mgmt_port_offset must be < 65536 to be usable as a port
number. The error message shown for "too many configs" is reworded.
(English language file only).

Note: The current way of selecting the management port based on the
index of the config file increases chances of port conflicts
when the number of configs is large. It could be useful to change
this logic but that is beyond the cope of this PR.


tray.c
```diff
@@ -71,6 +72,27 @@ CreateBitmaps()
     }
 }
 
+/*
+ * Reallocate hMenuConn as required. In case of not enough memory,
+ * o.num_config is reset to max space available value so that the
+ * program can continue.
+ */
+void
+allocate_connection_menu()
```
lstipakov, line 81: All other functions in this file use CamelCase.

lstipakov (approved):
Looks good, tested with MSVC. One small nitpick.

tray.c
selvanair: Thanks for the review. Will change the function name for consistency.

selvanair:
Renamed allocate_connection_menu() to AllocateConnectionMenu()

PR #413, opened 2021-04-03
Updated Polish translation

thebodzio:
Updated Polish translation based on the most recent English localization resource.


thebodzio:
Another pull request of the same nature already present in the queue. Closing.

PR #418, opened 2021-04-22
Retry on management timeout instead of aborting

selvanair:
In  some cases the service may take a while to startup openvpn.exe,
causing connection to the management interface to timeout. This could
leave  behind the OpenVPN process if/when it eventually starts up.
(Trac 905, 1050).

As errors in starting up the OpenVPN daemon are independently
handled, its better to keep retrying the management interface connection
until aborted due to errors or by the user.

- On timeout, log a message on the status window and retry the
  management interface connection

- Eliminate the timed-out state that is no longer used

- Call StopOpenVPN() before abort so that OpenVPN daemon
  is not left running in case it starts up later.

- In the unlikely event that OpenManagement() fails, show an error

- User can abort by pressing disconnect

A "retrying.." message is logged on to the status window every
15 seconds.

See Trac: #905, #1050

### DIFF
--- a/main.c
+++ b/main.c
@@ -188,6 +188,7 @@ int WINAPI _tWinMain (HINSTANCE hThisInstance,
       { echo_,     OnEcho },
       { bytecount_,OnByteCount },
       { infomsg_,  OnInfoMsg },
+      { timeout_,  OnTimeout },
       { 0,        NULL }
   };
   InitManagement(handler);
--- a/manage.c
+++ b/manage.c
@@ -209,10 +209,9 @@ OnManagement(SOCKET sk, LPARAM lParam)
             else
             {
                 /* Connection to MI timed out. */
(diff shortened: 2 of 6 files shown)


selvanair:
Rebased to master and force-pushed. Can  we get a review of this -- it should help us to close Trac 905: https://community.openvpn.net/openvpn/ticket/905

lstipakov (approved):
Tested locally (paused service process on breakpoint before starting openvpn.exe process) - works as expected:

```
Mon Jun 20 15:57:49 2022 GUI> Connecting to management interface failed.
View log file (C:\Users\lev\OpenVPN\log\corp.log) for more details.
Mon Jun 20 15:57:49 2022 GUI> Retrying. Press disconnect to abort
```

After resuming service process, connection establishment proceeded.

Issue #417, opened 2021-04-21
disallow empty passwords ?


chipitsine:
we observe the following pattern in user behaviour: when they setup new PC, sometimes they *forgot* to press TAB to switch to "password" prompt and type password right after the login, i.e. user<missing tab>password. 

thus, password itself is empty, and sent as empty.

I'm not sure whether empty password is legitimate, maybe we should change OpenVPN-GUI behaviour to disallow empty passwords ?

I'd suggest to make "password" field red if OK is pressed on empty password. Thus we even do not have to translate it.

@selvanair , what do you think ?

chipitsine:
I cannot upload screenshot. Something changed in github upload :(

selvanair:
In OpenVPN core, empty username is not allowed, but empty password is not an error. I guess passing only username has its use cases like SSO where the auth process is federated out, or done out of band by other means. In such cases one may not want to send the password to the server.  

Ideally, we should have a way of indicating in the prompt process what inputs are required so that the rest can be greyed out in the dialog. Then it would be appropriate to demand a non-empty entries for all enabled inputs including password. In v2.6, we have some newer methods to support out of band authentication, but I don't think the initial username/password prompting has been improved. 

We could stop making OK the default button as soon the username is filled-in so that hitting return would not submit the dialog if password is empty. Without disabling OK so that it could be clicked with empty password.

chipitsine:
may I suggest the following approach ?

1) default behaviour is changed, i.e. empty passwords are disallowed
2) OK button is not disabled (people even do not press it, but hit Enter instead), 
3) if user hit Enter, password field becomes red on background, dialog is still there, connection is not started
4) if empty password is supposed to be legitimate for some reason, it might be allowed, for example, in config file


or ....

we can keep current default behaviour, but introduce additional flag in config, if it is set, empty passwords become disallowed

TinCanTech:
Could the GUI present a user dialogue window asking for empty password confirmation ?

selvanair:
Let's not make it more user-unfriendly. Those who mistakenly submit empty passwords is no different from those who mistype passwords. 

At best, we can make none of the buttons the defualt when all fields are not filled in though.  That should take care of those hit enter without filling in the password and still allow to press either button. Its either that or leave everything s is.

TinCanTech:
Make the user select a button and not have `OK/Enter` selected by default. I see the logic now.

chipitsine:
[...]

PR #419, opened 2021-04-23
Do not allow empty password input

selvanair:
- In User-Auth dialog require non-empty password or PIN
- In generic password dialogs require non-empty inputs
- Fix handling of dynamic challenge when response is not required

Issue: #417


chipitsine:
I will test next holiday, thanks!

chipitsine (approved):

chipitsine:
I tested. Works as expected. Nice.

TinCanTech:
> I will test next holiday, thanks!

> I tested. Works as expected. Nice.

Happy holiday ! ;-)

PR #420, opened 2021-05-13
CMake, vcpkg manifest and github actions

lstipakov:


lstipakov:
You can see the effect of github actions here: https://github.com/lstipakov/openvpn-gui/actions/runs/838850412

chipitsine (approved):

main.c
```diff
@@ -177,7 +177,7 @@ int WINAPI _tWinMain (HINSTANCE hThisInstance,
   mgmt_rtmsg_handler handler[] = {
       { ready,    OnReady },
       { hold,     OnHold },
-      { log,      OnLogLine },
+      { log_,      OnLogLine },
```
selvanair, line 180: While at it shall we decorate all those enums with an underscore -- like echo_ and ready_? Or all caps?

lstipakov: Agreed, I'll add _ to the rest enum names for consistency.

selvanair: Looks good now. 
For other changes, you now better and @chipitsine has approved. Merging.

CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
...
+    proxy.c
+    registry.c
+    save_pass.c
+    scripts.c
+    service.c
+    tray.c
+    viewlog.c
+    res/openvpn-gui-res.rc)
+
+configure_file(config-msvc.h config.h COPYONLY)
```
selvanair, line 28: @lstipakov  Currently PACKAGE_VERSION_RESOURCE is hard coded in config-msvc.h. For MSVC build how to extract it from configure.ac during build time?

lstipakov: Good point. That's how: https://github.com/OpenVPN/openvpn-gui/pull/427

Issue #422, opened 2021-05-16
how to clear / remove the password from the certificates or p12 certificates


comsyspro:
in the openvpn-gui you can change or set a new password for the used certificate. however you can't clear / remove the password by typing an empty one because you have to set at least a 8 characters long password there. so could you add this functionality or is there a clear password function implemented yet and if yes where to find it?

selvanair:
This looks like a bug from the earliest versions of the GUI that we have record of -- the length of the password is checked before optionally allowing empty passwords.

That said, password change is a rarely used feature these days as it works only when the key is in a file (PEM or pkcs12) with write access for user running the GUI. If possible, store the key and cert in the Windows cert store with key as non-exportable and use cryptoapi-cert. That will provide some protection to the key without requiring a password.

comsyspro:
thanks for your answer. it's only a logical thought but if you can add or change a password for a certificate you should also be able to delete it.

PR #423, opened 2021-05-16
Allow clearing of key password in ChangePassphraseDialogFunc

selvanair:
Use an empty password to clear any existing password in the
private key file. If not empty, the requirement of minimum 8
characters is retained.

For PEM key file, an empty password will clear encryption on
the key. For pkcs12 files it will set an empty password.
As use of an empty password is the default first try in
OpenVPN.exe when reading pkcs12 file, this effectively leads
to the user not prompted for a private key password during
connection setup.

See #422


lstipakov (approved):
I haven't tested it, but code looks fine and compiles both locally and on github actions.

PR #424, opened 2021-05-19
Make all resource files MSVC compatible

selvanair:
Commit 1 is automated using an ugly sed script https://gist.github.com/selvanair/ae78c29869d7c1d15abcb909f04676c6

For many controls we use 0 as the ID which is okay as we have unique IDs for all controls that we send messages to. But MSVC issues duplicate ID warnings for them -- should fix some time.


lstipakov (approved):

lstipakov:
Thanks a lot @selvanair @d12fk .

lstipakov:
Works like a charm on ARM64: 
screen

PR #425, opened 2021-05-23
Have unique and non-zero IDs for all controls

selvanair:
Should fix duplicate control id warnings from resource
compiler.


lstipakov (approved):
Tested with VSMC, "duplicate control ID" warnings disappeared.

PR #426, opened 2021-05-23
Text truncated in spanish translation: fix by removing repeated phrase

selvanair:
This is rebased on top of PR #425 -- so ignore commit 1

Fixes: ..\openvpn-gui\res\openvpn-gui-res-es.rc(219): warning RC4206: title string too long; truncated at 256


lstipakov (approved):
¡se ve bien!

PR #427, opened 2021-05-24
CMake: read version from configure.ac

lstipakov:
Instead of using hardcoded major/minor version in config-msvc.h,
use the values from configure.ac.


selvanair:
Looks good and works.

PR #428, opened 2021-05-31
Apply transparency mask to the connecting-state checkmark image

selvanair:
Trying again to get a better connecting checkmark --- this is meant to replace PR #406 which was not up to the mark. Looks better except for some ant-aliasing due to scaling.

image

The bitmap is recreated every time the popup menus are made although its sufficient to recreate it when system colours change. 

In fact we should reconsider the current practice of scanning config directories and recreating menus multiple times as its a slow process when there are 100's of configs. Left for later.

### DIFF
--- a/tray.c
+++ b/tray.c
@@ -55,21 +55,75 @@ extern options_t o;
                                  || (o.config_menu_view == CONFIG_VIEW_NESTED))
 
 
-/* Create menu check bitmaps */
 static void
-CreateBitmaps()
+DeleteMenuBitmaps(void)
 {
-    if (hbmpConnecting)
-        return;
+    if (hbmpConnecting) {
+        DeleteObject(hbmpConnecting);
+        hbmpConnecting = NULL;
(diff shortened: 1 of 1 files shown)


lstipakov:
Stared at the code, could not find any GDI leaks. 

Icon looks better - no black background, except for the case when menu line is selected :)

lstipakov:
IMG_0485

lstipakov (approved):

selvanair:
> Stared at the code, could not find any GDI leaks.
> 
> Icon looks better - no black background, except for the case when menu line is selected :)

I know :) The background is now hard coded to that of the menu which is different from the bluish highlight color. In case of system checkmark it seems the they redraw it when selected. That said, when high contrast desktop is selected (black menu background), the behaviour between system checkmark and our connecting one is less different. Unless there is an API to select different images for selected and unselected states, I will leave it at that.

PR #429, opened 2021-05-31
Web-based extra authentication

lstipakov:
This adds support for web-based extra authentication, which may be
used by OpenVPN Cloud. When enabled and client sends IV_SSO=openurl,
server pushes Info command OPEN_URL:<url>. The client opens that URL and
user authenticates.

### DIFF
--- a/openvpn.c
+++ b/openvpn.c
@@ -1282,6 +1282,25 @@ void OnByteCount(connection_t *c, char *msg)
 }
 
 /*
+ * Handle INFOMSG from OpenVPN. At the moment in only handles
+ * "OPEN_URL:<url>" message used by web-based extra authentication.
+ */
+void OnInfoMsg(connection_t* c, char* msg)
+{
+    PrintDebug(L"OnInfoMsg with msg = %S", msg);
+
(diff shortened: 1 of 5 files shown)


openvpn.c
```diff
@@ -1281,6 +1281,25 @@ void OnByteCount(connection_t *c, char *msg)
...
+/*
+ * Handle INFOMSG from OpenVPN. At the moment in only handles
+ * "OPEN_URL:<url>" message used by web-based extra authentication.
+ */
+void OnInfoMsg(connection_t* c, char* msg)
+{
+    PrintDebug(L"OnInfoMsg with msg = %S", msg);
+
+    if (strstr(msg, "OPEN_URL:"))
+    {
```
selvanair, line 1293: I think we should be more precise here using strbegins(). The format of <EXTRA> is not adequately documented but there should be no leading spaces matching with other message types like >INFO:, >NEED-OK and >PASSWORD: Anyway, using msg+9 below assumes this.

Looks good otherwise -- only stare-tested :)

lstipakov: Agreed.

selvanair:
This may be closed now.

PR #430, opened 2021-06-01
Support for crtext

lstipakov:


lstipakov:
This PR also includes commit with openurl support, which has its own PR https://github.com/OpenVPN/openvpn-gui/pull/429

openvpn.c
```diff
@@ -64,13 +64,14 @@ TerminateOpenVPN(connection_t *c);
...
-#define FLAG_CR_TYPE_SCRV1 0x1    /* static challenege */
-#define FLAG_CR_TYPE_CRV1  0x2    /* dynamic challenege */
-#define FLAG_CR_ECHO       0x4    /* echo the response */
-#define FLAG_CR_RESPONSE   0x8    /* response needed */
-#define FLAG_PASS_TOKEN    0x10   /* PKCS11 token password needed */
-#define FLAG_STRING_PKCS11 0x20   /* PKCS11 id needed */
-#define FLAG_PASS_PKEY     0x40   /* Private key password needed */
+#define FLAG_CR_TYPE_SCRV1  0x1    /* static challenege */
+#define FLAG_CR_TYPE_CRV1   0x2    /* dynamic challenege */
+#define FLAG_CR_TYPE_CRTEXT 0x80   /* crtext */
```
selvanair, line 69: Please move this to the end of the list, below FLAG_PASS_KEY. Keeping the values in order makes it easier to find what the next unused value is.

selvanair (commented):
Looks good.

openvpn.c
```diff
@@ -1281,6 +1329,44 @@ void OnByteCount(connection_t *c, char *msg)
             LoadLocalizedString(IDS_NFO_BYTECOUNT, in, out));
 }
 
+/*
+ * Handle INFOMSG from OpenVPN. At the moment in only handles
+ * "OPEN_URL:<url>" message used by web-based extra authentication.
```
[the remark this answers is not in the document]
lstipakov: Indeed.

openvpn.c
lstipakov: I was unsure should it be ordered by name or value. Will move to the end.

selvanair (approved):

selvanair (commented):
Our handling of CR_TEXT when response is not required appears to be incorrect. In case of dynamic challenge (CRV1) even when the user response is empty, there is something to submit as the reply is formatted as "CRV1::ID::response". But in case of cr-response there is nothing to submit in this case.

I think we should just show the message to the user and end the dialog. Otherwise management interface will flag an error when "cr-response" is issued with no argument.

This "no response required" is unlikely use case,  but we have to handle it. If the above sounds right, I'll make a small patch.

schwabe:
To be honest, I have no idea what challenge/response without response should be. It is such an obscure corner case that I am fine with both solutions (implement empty response or don't respond at all) but we should document it

selvanair:
In that case let's be consistent with  CRV1 which documents that an empty response should be submitted. In this case that would be cr-response "" and the server side handler will get CR_RESPONSE,{CID},{KID},  with the response_base64 text empty.

PR #431, opened 2021-06-03
Handling of CR_TEXT when no response is required

selvanair:
As with CRV1, submit an empty string as the response.
Our base64-encode functiton can handle empty input to generate
an empty string as output.

Also make ensure the message box is shown in foreground,
and not dependent on the status window which may be hidden.

### DIFF
--- a/openvpn.c
+++ b/openvpn.c
@@ -650,7 +650,7 @@ GenericPassDialogFunc(HWND hwndDlg, UINT msg, WPARAM wParam, LPARAM lParam)
             {
                 wchar_t title[256];
                 GetWindowTextW(hwndDlg, title, _countof(title));
-                if (MessageBox(param->c->hwndStatus, wstr, title, MB_OKCANCEL) == IDOK)
+                if (MessageBoxExW(NULL, wstr, title, MB_OKCANCEL|MB_SETFOREGROUND|MB_TOPMOST, GetGUILanguage()) == IDOK)
                     SimulateButtonPress(hwndDlg, IDOK);
                 else
                     SimulateButtonPress(hwndDlg, IDCANCEL);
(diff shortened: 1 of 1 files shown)


lstipakov:
I am not allowed to submit an empty response. 
empty

Should we add something like

```
diff --git a/openvpn.c b/openvpn.c
index d29363c..8e1f028 100644
--- a/openvpn.c
+++ b/openvpn.c
@@ -679,7 +679,8 @@ GenericPassDialogFunc(HWND hwndDlg, UINT msg, WPARAM wParam, LPARAM lParam)
             SetForegroundWindow(hwndDlg);

         /* disable OK button by default - not disabled in resources */
-        EnableWindow(GetDlgItem(hwndDlg, IDOK), FALSE);
+        if (!(param->flags & FLAG_CR_TYPE_CRTEXT))
+            EnableWindow(GetDlgItem(hwndDlg, IDOK), FALSE);

         break;

@@ -693,7 +694,7 @@ GenericPassDialogFunc(HWND hwndDlg, UINT msg, WPARAM wParam, LPARAM lParam)
             if (HIWORD(wParam) == EN_UPDATE)
             {
                 /* enable OK if response is non-empty */
-                BOOL enableOK = GetWindowTextLength((HWND) lParam);
+                BOOL enableOK = GetWindowTextLength((HWND) lParam) || (param->flags & FLAG_CR_TYPE_CRTEXT);
                 EnableWindow(GetDlgItem(hwndDlg, IDOK), enableOK);
             }
             break;

```

selvanair:
We do  not show the dialog when flag R is not present -- instead we show a messagebox and autosubmit an empty response. The PR only adds "" as the response so that it can be submitted without error.

When flag R is present, the dialog is shown and we do not allow an empty response to be entered by the user -- that is deliberate.

lstipakov (approved):

PR #436, opened 2021-07-02
URL profile import

lstipakov:
This adds Access Server (https://openvpn.net/access-server/) profile import functionality. While it might sound scary, it just adds one menu item "Import from Access Server", which opens simple host/username/password dialog. When user clicks "OK", VPN profile gets downloaded from Access Server. The logic is described here: https://github.com/OpenVPN/openvpn3/blob/master/doc/webauth.md#basic-api

Missing pieces are (to be added later):
 - support for AS with self-signed certificates
 - support for 2FA


lstipakov:
@selvanair @cron2 I can provide you credentials to our test AS instance if you want to test this.

selvanair:
> @selvanair @cron2 I can provide you credentials to our test AS instance if you want to test this.

That would be nice -- I have an AS install lurking around somewhere but don't recall what state its in... I'll have time to test only after a next week though..

lstipakov:
@cron2 @selvanair have you guys had a chance to test this?

lstipakov:
I have re-implemented profile download using slightly different WinInet API, which enables to show system prompt (https://docs.microsoft.com/en-us/windows/win32/api/wininet/nf-wininet-interneterrordlg) in case of certificate verification error and gives user a chance to continue. 

This enables to support AS instances with self-signed certs.

cron2:
Hi,

On Mon, Jul 19, 2021 at 03:04:37AM -0700, Lev Stipakov wrote:
> @cron2 @selvanair have you guys had a chance to test this?

I haven't yet, but as I mentioned in the meeting two weeks ago, I'm not
sure if this is a good approach.

"Easy import from AS", yes, but "add an extra download method that
requires putting in some magic URL and foreign credentials to OpenVPN GUI",
maybe not.

I wonder if this can be solved by

 - having the web interface of AS (where you login with your AS credentials,
   and can verify "yes, this is the machine I want to talk to") produce
   the profile, as HTTP download, with a specific MIME-Type

   -> I assume that all this is in place already anway, but haven't checked

 - register said MIME-Type in Windows for "if such a file is downloaded,
   open OpenVPN GUI for import"


this would have less code impact on the GUI, and a more logical (for me)
flow of credentials and HTTPS access as well.

gert

-- 

lstipakov:
Hi,

You are right, AS has a web UI, from where one can download profile and then manually import in into openvpn-gui.

However, I believe that supporting profile import from AS would be a better user experience. The code uses WinHTTP (which means no external dependencies, contrary to the first implementation I had with libcurl) which takes care about HTTPS and prompts user via standard OS dialog when here is something wrong with certificate.

What comes to "magic URL" and "foreign credentials" - the URL is just a host name, and openvpn-gui already has a code to deal with credentials, including dynamic challenge support, which might be used for profile download. I reused a lot of existing code with minimal changes. The whole process of fetching profile via REST API is public and well documented: https://github.com/OpenVPN/openvpn3/blob/master/doc/webauth.md#rest-profile-download-api

Profile download is already supported by following clients:
 - OpenVPN Connect (Windows, macOS, Android, iOS)
 - openvpn3-linux (https://github.com/OpenVPN/openvpn3-linux/blob/master/src/python/openvpn3-as)
 - OpenVPN for Android (https://github.com/schwabe/ics-openvpn/blob/master/main/src/ui/java/de/blinkt/openvpn/fragments/ImportASConfig.kt)

So I think there is a missing part in openvpn-gui functionality, which this PR adds.

I haven't played with MIME types, but if something like this could be done, we could have it, too.

novaflash:
The magic URL is standard REST API call. This function has existed since approx. 12 years ago. Third-party tunxten client can do it. OpenVPN 3 Linux client can do it. OpenVPN GUI is considered the power user's tool, would be nice if OpenVPN GUI can do it too. 

Regarding mime-type - yes that is a good idea and that is a solution we're moving towards, but that's still in development, both on server and client side. So this is something that should be done too. But what lstipakov implemented is what exists now.

lstipakov:
> I tend to agree that we need not look at this as support for some opaque protocol in AS. The approach looks usable in non AS installations too and it could be called something more generic like "Import from URL".

I just checked OpenVPN Connect and this functionality is indeed named there "Import from URL". I'll rename it here to "Import from URL" too.

cron2:
Hi,

On Tue, Jul 20, 2021 at 07:30:18AM -0700, Selva Nair wrote:
> I tend to agree that we need not look at this as support for some
> opaque protocol in AS. The approach looks usable in non AS installations
> too and it could be called something more generic like "Import from URL".

OK.  I withdraw my objection, then.

I still think that "I log into AS, click on the profile, and my local
client asks me whether I want to import it" is a more logical user 
experience - but that might be because this is how I deal with my 
user certificates.  They have a web interface that generates a link
and a QR code (for "I do the typing on the PC but want the profile
on my mobile") and on clicking, the .ovpn file is downloaded and can
then be imported.

On iOS, "click -> download -> import" is all straightforward ("yes, yes,
yes, save"), other platforms are less automatic.

gert

-- 

novaflash:
Yeah the generic name is better.

lstipakov:
Ping @cron2 @selvanair 
Would you mind to take a look on it @d12fk  ?

selvanair (commented):
I have only skimmed through the code and here are some quick comments:

Things like OVPN_ACCESS_SERVER_PROFILE need to be replaced by more generic keywords like OVPN_PROFILE, OVPN_PROFILE_FRIENDLY_NAME etc. Supporting AS-keywords as an alternative on top of that is okay, but we do not want a feature that is intimately tied to AS.

The code itself needs some improvement: 

- strcpy instead of strncpy, strncpy used without checking the result will fit, etc. 
- _snwprinf needs NUL termination check. In fact we use swprintf instead of _snwprintf elsewhere (both have same signature and both do not guarantee NUL termination). It may be time to add a wrapper like openvpn_swprintf. 
- friendly_name is not assigned
- As we do everywhere else, treat all user input as wide strings -- url, username, password etc., and convert to utf8 if required. ANSI depends on the code page in effect.
- All message boxes show to the user should be localized
- What is the character code specification of the profile data -- if its utf8, convert to wide character after download. Or check how wide character wininet functions behave -- if they do automatic conversion during download, use of wide character versions may just work.
- In the dialog, do we have to require password to be filled in? I can think of situations where a password may be queried over a side channel.
- URL parsing could use UrlGetPartW ?

lstipakov:
Thanks for the review, @selvanair! I believe I addresses most of your concerns. A few things:

 - I switched to using `swprintf`. According to MSFT doc, it adds NULL terminator (https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/sprintf-sprintf-l-swprintf-swprintf-l-swprintf-l?view=msvc-160#remarks). Also I made a small test app and checked that this is indeed the case. So unless we have an evidence that modern GCC/MSVC doesn't add NULL terminator, I don't think we need to add it manually or have our own wrapper.

 - What is the benefit of converting downloaded profile into wchar? At the moment dynamic challenge string is ASCII, and the existing code in gui which parses that string is dealing with plain chars. I don't think we ever tried to have challenge string like "enter google authenticator code" with non-ACSII chars. Maybe this is something which could be implemented separately.

 - UrlGetPartW doesn't work if there no prefix in the input, which is most likely the case when user enters URL.

 - At the moment password is required.

selvanair:
Thanks for the changes.
I've to find some time to give this a  test run, sorry for taking so long.

> Thanks for the review, @selvanair! I believe I addresses most of your concerns. A few things:
> 
> * I switched to using `swprintf`. According to MSFT doc, it adds NULL terminator (https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/sprintf-sprintf-l-swprintf-swprintf-l-swprintf-l?view=msvc-160#remarks). Also I made a small test app and checked that this is indeed the case. So unless we have an evidence that modern GCC/MSVC doesn't add NULL terminator, I don't think we need to add it manually or have our own wrapper.

What happens if the call returns an error (-1)? I think the standard is not clear about whether a NUL is added in this case nor is the implementation docs. On top of that if someone builds this using MINGW_ANSI_STDIO, they get a different implementation of swprintf which may have its own issues. Its much safer to add NUL or have a wrapper.

> * What is the benefit of converting downloaded profile into wchar? At the moment dynamic challenge string is ASCII., and the existing code in gui which parses that string is dealing with plain chars. I don't think we ever tried to have challenge string like "enter google authenticator code" with non-ACSII chars. Maybe this is something which could be implemented separately.

I was not referring to anything in particular in the profile -- OpenVPN profile is allowed to contain non-ascii characters and we should not break that. Including the challenge string that you mention. As long as it correctly downloads UTF8 data and writes it to a file without mangling non-ascii characters, its okay. I thought the easiest way to do that would be to convert the data from UTF8 to wide characters. For example, if the profile name contains non-ascii characters, would it get interpreted correctly or would it get interpreted by the codepage in use which we do not want.

Also the general practice in the GUI code so far has been to treat almost all strings as wide char and convert to UTF8 when interacting with OpenVPN.

> * UrlGetPartW doesn't work if there no prefix in the input, which is most likely the case when user enters URL.
> * At the moment password is required.

That's easy to change. Juts enable OK button if username and url are filled in leaving password optional. If SSO is in use, we may not want the user to send password to the app, as authentication will be done independently as required.

lstipakov:
All concerns should be now addressed.

Took me longer than it should, but now URL profile import is UTF8-compatible. I tested with non-ASCII chars both in profile content and dynamic challenge request / response.



how r u doing
 
resp

lstipakov:
ping @selvanair

lstipakov:
ping @selvanair @cron2

lstipakov:
also @d12fk

selvanair:
Sorry for the long delay in getting back to this.

Thanks for the edits. Finally I got a chance to test this. Unlike what I thought, the URL that user input is just the hostname (plus optional :port) not a generic URL. If we are to provide this as a generally usable way of importing config from a web address, we should be taking any valid URL and downloading from it. Why not allow any url like https://example.com/foo/bar/x=y?etc...  As an admin I could just send the user a url and they should be able to cut and paste it into the dialog and get the profile imported.

Apart from that there are some minor issues like (i) in ParseURL, if the port is not specified, it defaults to 443 even if the scheme is explicitly given as http (ii) unused variables like username and password in the dialog function (iii) space for URL is too short if we allow a generic URL etc. I'll give it a more thorough review once we have consensus about what the feature is.

lstipakov:
Hi,

(i) and (ii) should be now fixed. 

About generic URLs - the endpoint address format is somewhat standardized here https://github.com/OpenVPN/openvpn3/blob/master/doc/webauth.md#basic-api

For example

> https://asdemo.openvpn.net/rest/GetUserlogin?tls-cryptv2=1&action=import

If we are going to support any valid URLs such as https://example.com/foo/bar?x=y instead of hostname only,  how do we implement it in accordance to abovementioned spec? Would it be enough to just add path to (/rest/GetUserlogin) to whatever URL we're given, stripping parameters?

selvanair:
Why not require the full URL even for AS? Like `https://asdemo.openvpn.net/rest/GetUserlogin?tls-cryptv2=1&action=import` or `https://asdemo.openvpn.net/rest/GetAutologin` or whaetever. That also eliminates the need for the Autologin tick mark as the meaning of that is not clear to a casual user.

Anhyway, the user has to be somehow informed of how to get the config and that could be via an email with a URL which could be cut-n-pasted to the dialog. All options like autologin and others are embedded in the URL requiring no further choices on part of the user.

lstipakov:
My understanding is that REST API details is not something user needs / should know. AS (or another server which implements that spec) is deployed somewhere and only thing user is interested in is a hostname.  I would also argue that from usability perspective requiring hostname is a better option comparison to the full URL - hostname is something which could be easily remembered, if needed.

Besides, all existing clients which implement profile fetching functionality (opensource openvpn3-linux ja openvpn for android, closed-source connect) work with hostnames.

novaflash:
For user experience it is better that only hostname + credentials leads to a successful import. But it has to be safe.

Allowing http:// without the benefit of HTTPS security is quite bad. Transporting private keys over an unsecured channel is not good. Doing it over https:// only at least forces some reasonable security. Hence the additional logic to check validity of SSL certificate. Also, if you do just HTTP get request without implementing a REST API, you can only implement HTTP basic auth (over HTTPS), which doesn't let you do multifactor authentication. This solution does allow that. Again, better for security.

selvanair:
I totally agree we should require https. But let's don't get lost in any pretext of "RESTfulness" -- we are not implementing a REST client in the GUI. Its just one URL for downloading a profile. Also we are not distributing a server manager that speaks AS's REST API, we expect administrators to implement their own server side processing. There is no reason to force them to use a particular URL format.

If we want to make use of REST, we need a smarter implementation. One which returns, say, JSON to direct the next step which could involve downloading of dependent files like scripts etc.

As for multifactor, it only requires a state transition at the web server, which any URL can can do with proper server-side processing. Sure, we have to standardize the error response that triggers the dynamic response. For that, using the proposed xml is fine by me.  Essentially, while we do need an agreement on the format of data or errors returned by the web server so that the GUI can parse it right, the URL could be anything.

At least users I support wont care whether they have to paste a URL or hostname or something else in to a dialog -- its all opaque to them anyway. In fact, that radio button "Autologin" will stump most users I know and lead to a service call. The server should decide whether the user has to be given a profile with auth-user-pass or not.

Finally, nothing is gained by hiding the details of the URL and quite some flexibility is lost by it. What if an admin wants to pass client-version=11.25. Each time a new option is added we change the implementation?

I think the way it's coded, it's "Import from AS or AS-clone server" in a very limited way and not "Import from a URL". That's of no interest to me, but some may find it useful.

lstipakov:
Selva,

I think I understand your point - you don't want certain commercial closed-source product support in open-source community maintained client. Nevertheless, I want to reiterate my points:

 - profile download API spec is open source and has existed for many years and hasn't changed much
 - to the best of my knowledge, the only product which has implemented that spec is Access Server
 - 3 other OpenVPN clients which support profile download (openvpn3-linux, openvpn for android, openvpn connect) have implemented profile download according that spec

Based on those points, I see no reason why openvpn-gui should implement things differently. That said, I want to find a compromise solution which you as this project's maintainer would accept.

What if we expand URL import dialog like this:

```
----------------------------------
|Import from URL                 |
----------------------------------
|                                |
| URL: <editbox>                 |
|                                |
| Import from Access Server: [X] |
| -----------------------        |
| | Username: <editbox> |        |
| | Password: <editbox> |        |
| | Autologin: [X]      |        |
| -----------------------        |
|                                |
| <OK button>  <Cancel button>   |
----------------------------------

```

We can discuss if "Import from AS" checkbox should be checked by default - I think as for now most of users will be AS users. When it is unchecked, the groupbox with controls is disabled. This way this feature might be valuable for your users - they just paste opaque URL (which might look like https://example.com/vpnserver?client-version=42), which client
uses as is to download profile. On the other hand AS users will check "AS" checkbox, enter `as.mycompany.com` as URL, username/password and check "autologin" if needed - they're likely aware of its meaning.

What do you think?

selvanair:
@lstipakov : I understand the push for an implementation that AS users will easy to use. Shall we just rename this to "Import from AS" or "Import from Server" (back to the original) and we can then add the feature to Import from a generic URL with little work, reusing much of what you have implemented in this PR.

Essentially same as what you wrote, but I was thinking of refactoring Import: by replacing "Import from file.." to just "Import" with sub menus like "Import from file...", "Import from AS" or "Import from URL". The approach you suggested is fine too, though less complex dialogs are easier on the user.

The URL input could be made somewhat longer for a better UX, say about 60 characters -- currently it start scrolling at 33 or 34 characters.

lstipakov:
I have pushed an update:

 - Added "Import" menu item, which now includes "Import from file" (same text as before) and "Import from Access Server". "Import" and "Import from file" strings are added to all resource files.
 - Dialog title is changed from "Import Profile from URL" to "Import Profile from Access Server"

as.c
```diff
@@ -32,11 +33,338 @@
...
+        if (!InternetReadFile(hRequest, buf + pos, READ_CHUNK_LEN, &bytesRead)) {
+            ShowWinInetError(hWnd);
+            return FALSE;
+        }
+        if (bytesRead == 0) {
+            size = pos;
+            break;
+        }
+
+        buf[pos + bytesRead] = '\0';
```
selvanair, line 183: This should be moved before the break above, isn't it?

as.c
```diff
@@ -32,11 +33,338 @@
...
+
+        if (pos + bytesRead >= size) {
+            size += READ_CHUNK_LEN;
+            char* ptr = realloc(buf, size + 1);
+            if (!ptr) {
+                MessageBoxW(hWnd, L"Out of memory", _T(PACKAGE_NAME), MB_OK);
+                return FALSE;
+            }
+            buf = ptr;
```
selvanair, line 192: Possible memory leak on failure, and *pbuf not reassigned on success. Suggested fix:
```
*pbuf = realloc(*pbuf, size + 1);
if (!*pbuf) {
     free(buf);
     MessageBoxW......
     return FALSE;
}
buf = *pbuf;
```
Then local variable `ptr` not required.

as.c
```diff
@@ -32,11 +33,338 @@
...
+
+    /* handle cert errors */
+    /* https://www.betaarchive.com/wiki/index.php/Microsoft_KB_Archive/182888 */
+    if (!HttpSendRequestW(hRequest, NULL, 0, NULL, 0)) {
+        DWORD err = GetLastError();
+        if ((err == ERROR_INTERNET_INVALID_CA) ||
+            (err == ERROR_INTERNET_SEC_CERT_CN_INVALID) ||
+            (err == ERROR_INTERNET_SEC_CERT_DATE_INVALID) ||
+            (err == ERROR_INTERNET_SEC_CERT_REV_FAILED)) {
+
```
selvanair, line 342: I think we should error out here and not allow unsafe behaviour.

as.c
```diff
@@ -32,11 +33,338 @@
...
+/**
+ * Download profile from AS and save it to a special-named temp file
+ *
+ * @param hWnd handle of window which initiated download
+ * @param host AS hostname, entered by user, might contain protocol and port
+ * @param username UTF-8 encoded username used for HTTP basic auth
+ * @param password UTF-8 encoded password used for HTTP basic auth
+ * @param autologin should autologin profile be used
+ * @param out_path full path to where profile is downloaded. Value assigned by this function.
```
selvanair, line 280: Caller must ensure out_path is at least MAX_PATH. Better MAX_PATH + 1 as even microsoft is unsure whether +1 is needed :) I would prefer passing in the size of out_path from caller. Also see below.

as.c
```diff
@@ -32,11 +33,338 @@
...
+    ExtractProfileName(wbuf, comps.host, name, MAX_PATH);
+    free(wbuf);
+
+    /* save profile content into tmp file */
+    DWORD res = GetTempPathW(MAX_PATH, out_path);
+    if (res == -1 || res > MAX_PATH) {
+        MessageBoxW(hWnd, L"Failed to get TMP path", _T(PACKAGE_NAME), MB_OK);
+        goto done;
+    }
+    wcscat(out_path, name);
```
selvanair, line 432: Use wcsncat or our sntprintf_0 to append the file name. Also safer to pass in the size of out_path or ensure at least room for MAX_PATH+1 is available.

as.c
```diff
@@ -304,18 +373,46 @@ DownloadProfile(HANDLE hWnd, const WCHAR *host, const char *username, const char
...
-    }
+    HttpQueryInfoW(hRequest, HTTP_QUERY_STATUS_CODE | HTTP_QUERY_FLAG_NUMBER, &status_code, &length, NULL);
 
-    /* download profile content */
     size_t size = 0;
-    if (!DownloadProfileContent(hWnd, hRequest, &buf, &size))
+
+    /* download profile content */
+    if ((status_code == 200) || (status_code == 401)) {
+        if (!DownloadProfileContent(hWnd, hRequest, &buf, &size))
```
selvanair, line 384: Would the user get any feedback if this happens? Show an error?

misc.c
```diff
@@ -632,3 +634,55 @@ open_url(const wchar_t *url)
...
+        ShowLocalizedMsg(IDS_ERR_IMPORT_SOURCE, source);
+        return;
+    }
+
+    _tcscat(fileName, ext);
+
+    TCHAR destination[MAX_PATH];
+    _sntprintf_0(destination, _T("%s\\%s"), o.config_dir, fileName);
+
+    destination[_tcslen(destination) - _tcslen(o.ext_string) - 1] = _T('\0');
```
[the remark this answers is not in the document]
selvanair: My review got submitted half-way through.. Also requires changing line 656 to
_sntprintf_0(destination, _T("%s\\%s.%s"), destination, fileName, o_ext_string);

as.c
```diff
@@ -32,11 +33,338 @@
...
+
+    WCHAR* strport = wcsstr(url + domain_off, L":");
+    if (strport) {
+        WCHAR* end;
+        wcsncpy(comps->host, url + domain_off, strport - url - domain_off);
+        comps->port = wcstol(strport + 1, &end, 10);
+    }
+    else {
+        if (comps->port == 0)
+            comps->port = 443;
```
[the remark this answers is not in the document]
selvanair: This makes no sense now -- not sure where related comments went. Will re-write it.

selvanair:
Two general comments that went missing:

(i) We should just reject downloading from links with unverifiable certificates. Server administrators have to either use publicly verifiable certs or pre-install trusted CA certs in user machines (as one would do in some environments). We have no way of showing the failing certificate to the user. Even otherwise this is not a place where a user can make a judgement about the failure.

(ii) I do not think we should send username/password over http. So it has to be either https enforced or no password send otherwise. The former is simpler. We can still parse "url" starting with http://, but set the secure flag nevertheless and default the port to 443 unless explicitly given.

as.c
```diff
@@ -32,11 +33,338 @@
...
+            wcsncpy(friendly_name, pch + wcslen(FRIENDLY_NAME_TOKEN), PROFILE_NAME_LEN - 1);
+            friendly_name[PROFILE_NAME_LEN - 1] = L'\0';
+        }
+
+        pch = wcstok(NULL, L"\r\n");
+    }
+
+    if (wcslen(friendly_name) > 0)
+        swprintf(out_name, out_name_length, L"%ls.ovpn", friendly_name);
+    else if (wcslen(profile_name) > 0)
```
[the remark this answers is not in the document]
lstipakov: I see, didn't know that extension is customizable. I applied change you've proposed above, so the "correct" extension will be applied during import.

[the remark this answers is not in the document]
lstipakov: I will add a comment explaining that.

as.c
```diff
@@ -32,11 +33,338 @@
...
+/**
+ * Extract profile name from profile content.
+ *
+ * Profile name is either (sorted in priority order):
+ * - value of OVPN_ACCESS_SERVER_FRIENDLY_NAME
+ * - value of OVPN_ACCESS_SERVER_PROFILE
+ * - URL
+ *
+ * @param profile profile content
+ * @param url profile URL, might be used for name
```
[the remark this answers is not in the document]
lstipakov: Agreed,

cron2:
Hi,

On Mon, Aug 23, 2021 at 01:32:43AM -0700, Lev Stipakov wrote:
> +    if (wcslen(friendly_name) > 0)
> +        swprintf(out_name, out_name_length, L"%ls.ovpn", friendly_name);
> +    else if (wcslen(profile_name) > 0)

Does our windows code not follow our coding style wrt "no conditionals
without {} brackets"?

gert

-- 

lstipakov:
> Does our windows code not follow our coding style wrt "no conditionals without {} brackets"? 

openvpn-gui style is somewhat mixed - I've seen examples with and without brackets.

dsommers:
Just adding my 2cents to this discussion.

> (i) We should just reject downloading from links with unverifiable certificates. Server administrators have to either use publicly verifiable certs or pre-install trusted CA certs in user machines (as one would do in some environments). We have no way of showing the failing certificate to the user. Even otherwise this is not a place where a user can make a judgement about the failure.

I agree.   There has been some concerns about outdated OSes not being able to fully verify some AS certificates as well.  But I do believe we need to ensure end-users can have full confidence in the AS server they download the profile from.  It would be a fairly simple MITM attack vector to collect usernames/passwords, and with that getting a copy of the profile as it is being downloaded - including username/password.  End users may not fully understand these potentials and the risk related to that.

And in regards to oudated OSes - when an OS vendor stops supporting a release, there should not be any need for us to continue supporting it.

> (ii) I do not think we should send username/password over http. So it has to be either https enforced or no password sent otherwise. The former is simpler. We can still parse "url" starting with http://, but set the secure flag nevertheless and default the port to 443 unless explicitly given.

Agreed.  No plain-text passing of credentials should be allowed ever.

as.c
```diff
@@ -32,11 +33,338 @@
...
+* Extracts protocol, port and hostname from URL
+*
+* @param url URL to parse, length must be less than URL_MAX
+*/
+void
+ParseUrl(const WCHAR *url, struct UrlComponents* comps)
+{
+    ZeroMemory(comps, sizeof(struct UrlComponents));
+
+    int domain_off = 0;
```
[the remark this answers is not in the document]
lstipakov: Done. Although this might confuse a user which tries to connect via http and gets an error - we don't indicate that we force https. Maybe add uneditable editbox (sounds good, no?) with "https://" just at the left side of editbox with url. OTOH that might look ugly.

as.c
lstipakov: In my tests `InternetReadFile` added `NULL` terminator to a buffer when `bytesRead` was `0` (and yes I checked that this is the case for `realloc`-allocated memory, not `calloc`). Since this behavior is not mentioned in docs, let's indeed add `NULL` terminator no matter what `bytesRead` value is.

as.c
lstipakov: Indeed, especially `*pbuf not reassigned after realloc` was quite bad. I introduced `char* buf` to make code look nicer (comparison to using `*pbuf` in many places) and got bitten by it. Fixed.

as.c
lstipakov: Changed to `swprintf` and NULL termination.

as.c
lstipakov: Added `out_path_size` parameter.

as.c
lstipakov: `DownloadProfileContent` displays an error message in case of failure. I added this to doxygen comment.

lstipakov:
I've implemented all suggestions except unverifiable certificates. 

- "We have no way of showing the failing certificate to the user." - we use `InternetErrorDlg()` which shows localized OS prompt about exact problem with certificate and gives user a chance to see the certificate and proceed.
Sieppaa

 - There are many AS instances without valid certificates

I could mention that OpenVPN Connect allows to proceed with incorrect certificate (after displaying a warning), and moreover so do other apps, like browsers. Why should we behave differently?

selvanair:
Thanks for all the edits.

>> "We have no way of showing the failing certificate to the user." - we use InternetErrorDlg() which shows localized OS prompt about exact problem with certificate and gives user a chance to see the certificate and proceed.

I should have worded it differently. My point was not about showing the certificate --  an unverifiable certificate may be okay in some limited situations, but not in this case.

> I could mention that OpenVPN Connect allows to proceed with incorrect certificate (after displaying a warning), and moreover so do other apps, like browsers. Why should we behave differently?

We are not writing a browser (or Connect:). I do not think we want such behaviour in OpenVPN-GUI.

dsommers:
> Why should we behave differently?

Because we can do better ;-)

lstipakov:
I made it so that only in debug builds one could proceed with incorrect certificate.

as.c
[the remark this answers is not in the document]
selvanair: It'll behave like sites that automatically redirect to https. Looks fine to me as is. 

We could prefill with https:// and make that part not editable. There are other possible improvements including stripping leading whitespace, but let's get this merged and see how it fares in real use.

as.c
```diff
@@ -32,11 +33,341 @@
...
+    if (!wbuf) {
+        MessageBoxW(hWnd, L"Failed to convert profile content to wchar", _T(PACKAGE_NAME), MB_OK);
+        goto done;
+    }
+    ExtractProfileName(wbuf, comps.host, name, MAX_PATH);
+    free(wbuf);
+
+    /* save profile content into tmp file */
+    DWORD res = GetTempPathW(out_path_size, out_path);
+    if (res == -1 || res > MAX_PATH) {
```
selvanair, line 431: I thought I mentioned this before, but cant find it in my previous comments.. Anyway:

-1, is not a documented return value, while 0 or > buffer-size are: `(res == 0 || res > out_path_size)` ?
If checking for -1 is also required, suggest to write it as (DWORD) -1. 

All look good otherwise.

lstipakov: Not sure where `-1` came from, replaced with `(res == 0 || res > out_path_size)`.

selvanair (approved):

cron2:
Hi,

On Mon, Aug 23, 2021 at 01:42:46AM -0700, Lev Stipakov wrote:
> > Does our windows code not follow our coding style wrt "no conditionals without {} brackets"? 
> 
> openvpn-gui style is somewhat mixed - I've seen examples with and without brackets.

I consider Selva the current final authority on OpenVPN gui coding
style, so it's not my decision.

If I *were* to decide, I'd ask for all new code to be written according
to "main openvpn" coding style (which especially means "always {} brackets"),
possibly amended by "the way windows is coded" (// comments or such).

But this is strictly not a topic for this PR, but for a separate discussion.

gert

-- 

PR #437, opened 2021-07-14
Update Japanese translation

yamatarobe:
Translated based on the newest version.


res/openvpn-gui-res-jp.rc
```diff
@@ -452,20 +454,20 @@ Use openvpn-gui --help for more info."
...
-    IDS_ERR_INSTALL_ISERVICE     """OpenVPNServiceInteractive"" is not installed.\n\
-Tasks requiring administrative access may not work."
-    IDS_ERR_NOTSTARTED_ISERVICE  """OpenVPNServiceInteractive"" is not started.\n\
-Tasks requiring administrative access may not work."
-    IDS_ERR_NOTSTARTED_ISERVICE_ADM  """OpenVPNServiceInteractive"" is not started.\n\
-Wintun driver will not work."
+    IDS_ERR_ACCESS_SERVICE_PIPE "サービスパイプへのアクセスに失敗しました。"
+    IDS_ERR_WRITE_SERVICE_PIPE "サービスパイプへの書き込みに失敗しました。"
+    IDS_ERR_INSTALL_ISERVICE     """OpenVPNServiceInteractive"" がインストールされていません。\n\
+管理者権限で実行すると動作する可能性があります。"
```
selvanair, line 460: This wording is suggesting to run as admin when iservice is not installed. I thing we should say the connection may fail without the service. I want to encourage the use of iservice :)
Same below.

res/openvpn-gui-res-jp.rc
```diff
@@ -478,26 +480,26 @@ OpenVPNがインストールされていない可能性があります。"
...
-%s\n\nMake sure you have the right permissions."
-    IDS_NFO_IMPORT_SUCCESS "File imported successfully."
-    IDS_NFO_IMPORT_OVERWRITE "A config named ""%s"" already exists. Do you want to replace it?"
-    IDS_ERR_IMPORT_SOURCE "Cannot import file <%s> as it is already in the global or local config directory"
+    IDS_ERR_IMPORT_EXISTS "設定 ""%s"" は既に存在しています。"
+    IDS_ERR_IMPORT_FAILED "ファイルがインポートできません。以下のパスは作成されませんでした。\n\n\
+%s\n\n適切な権限があることを確認してください。"
+    IDS_NFO_IMPORT_SUCCESS "ファイルが正しくインポートされました。"
+    IDS_NFO_IMPORT_OVERWRITE "設定 ""%s"" は既に存在しています。上書きしますか？"
+    IDS_ERR_IMPORT_SOURCE "<%s> はグローバル/ローカル設定ディレクトリにあるためインポートできません。"
```
selvanair, line 488: ディレクトリ -> フォルダ to be consistent with the changes above in the settings menu?

selvanair (commented):
Thanks for your contribution. A very thorough job. My Japanese is poor and rusty, but see some comments below.

res/openvpn-gui-res-jp.rc
yamatarobe: OK, I understand that the use of this service is strongly recommended.
I'll update this message soon.

res/openvpn-gui-res-jp.rc
yamatarobe: I think that "フォルダ" is more commonly used in these days.
I'll update the translation.

yamatarobe:
Thank you for your feedback.
You understand the delicate nuance in Japanese !

I'll update the translation as soon as possible.

selvanair (commented):
Meiryo UI is nicer but not installed by default in all regions. Its in the supplemental fonts pack and may have to be manually installed. If the font is not present, Windows substitutes some other and badly messes up rendering without any warnings.

Is there any way we could fallback to a sane default if the font is not found? Its beyond the scope of this PR, but would be a useful addition. 

For now, if you think most Japanese users would have the font installed, I'm okay with this.

yamatarobe:
OK, I'll change the font to "Microsoft Sans Serif".
As a side note, Meiryo UI is installed on Windows (Japanese).

selvanair:
Okay, we can revisit this and add Meiryo UI as preferred when we have a process for a fallback font.

If you don't mind I'll squash all these commits into one before merging. Keeps the history clean.

selvanair:
Hmm.. I thought I chose to squash and merge, but it ended up just rebasing.

Thanks for your contribution.

Issue #440, opened 2021-08-11
CR_TEXT dialog should be closed if server receives client-auth, client-deny or client-kill


ro0p:
We use the server's management interface to authenticate users and handling 2FA. Our users can use OTP or push notification on their mobile device for 2FA. OTP works fine this way, but when push notification is accepted or rejected on mobile device, we send client-auth or client-deny to the server and the client is connects or disconnects as expedted, but CR_TEXT dialog stays and waits for user action.

 It would be very useful if this dialog would be closed automatically if the state of the connection is changed from "AUTH_PENDING" to anything else while the dialog waits for user input.

Thank you

selvanair:
I suppose you are sending CR_TEXT with no-response required (i.e. flag R not present) but the dialog still needs the user to press OK. When it was coded it was felt that if a message is to be shown to the user, some form of acknowledgement from the user is required before it can be closed. 

Your proposal to close it when authenticated makes sense. Or should we add a timeout?

ro0p:
Flag R is present, because the input field is required for typing the OTP if push notification is not working for some reason (network issue, buggy device etc).
I think using a timeout is not a good idea, because users can be... dumb :)

Closing the dialog on state change  will be great. If this behavior breaks the original design, it could be controlled by another flag.

Thank you

selvanair:
If R is present we interpret it as response is required and have to consider all use cases before auto closing the dialog. IIRC, there is an optional timeout value in AUTH_PENDING message. We could close the dialog at its expiry as the connection will anyway restart at that point. 

That said, looking for a state change that makes any pending auth dialog obsolete and closing them sounds good at first blush. Let's give it some more thought.

ro0p:
Sounds great.
Thank you

selvanair:
@ro0p see PR #441. Would appreciate it if you could test it.

ro0p:
Works like charm!
Thank you for the very quick solution!

selvanair:
Fixed by #441

PR #441, opened 2021-08-14
Close cr-text challenge dialog on state change

selvanair:
- All windows are notified when OpenVPN state changes -- cr-text dialog is made to close on receiving this (see issue #440)
- Change the way "challenges" that require no response are processed -- required to handle state-change notification
- Fix a bug in the parsing of the challenge text (we were truncating strings that contain colons)

### DIFF
--- a/openvpn.c
+++ b/openvpn.c
@@ -249,6 +249,20 @@ parse_assigned_ip(connection_t *c, const char *msg)
 }
 
 /*
+ * Send a custom message to Window hwnd when state changes
+ * hwnd : handle of the window to which the message is sent
+ * lParam : pointer to the state string (char *) received from
+ *          OpenVPN daemon (e., "CONNECTED")
+ * The function signature matches callback for EnumThreadWindows.
+ */
+BOOL
+NotifyStateChange(HWND hwnd, LPARAM lParam)
+{
+    SendMessage(hwnd, WM_OVPN_STATE, 0, lParam);
+    return TRUE;
+}
+
+/*
  * Handle a state change notification from the OpenVPN management interface
  * Format <TIMESTAMP>,<STATE>,[<MESSAGE>],[<LOCAL_IP>][,<REMOTE_IP>]
  */
@@ -273,6 +287,9 @@ OnStateChange(connection_t *c, char *data)
     if (pos == NULL)
         return;
     *pos = '\0';
+
+    /* notify the all windows in the thread of state change */
+    EnumThreadWindows(GetCurrentThreadId(), NotifyStateChange, (LPARAM) state);
 
     if (strcmp(state, "CONNECTED") == 0)
     {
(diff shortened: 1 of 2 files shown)


lstipakov:
Hi,

After this change CR dialog disappears quickly after popping up. See the difference below (second gif shows current behavior before this patch).

dialog_disappears
dialog_ok

selvanair:
Thanks for testing.
I was under the impression that when auth is pending no state change message arrives unless it times out or auth somehow completes out-of-band. But it looks like state change to GET_CONFIG happens when the auth is still pending. Will have to make this conditional on certain state changes.

Why is the client moving to GET_CONFIG and start sending PUSH_REQUEST when auth is pending? Is that a feature or a bug in OpenVPN?

schwabe:
Feature, kind of. PUSH_REQUEST is basically the keep alive in this state.

TinCanTech:
> Why is the client moving to GET_CONFIG and start sending PUSH_REQUEST when auth is pending? Is that a feature or a bug in OpenVPN?

_sending PUSH_REQUEST when auth is pending_ is a new OpenVPN 2.5 feature.  More commonly known as deferred-auth.

selvanair:
Thank you both. 

I suppose I can assume that the state will change to RECONNECTING or EXITING if auth doesn't complete or to CONNECTED if it succeeds. If so closing the dialog in those cases should be enough. 

This is essentially to handle cases where the challenge says, e.g.,  "Approve the push notification on your phone or input the code from the authenticator app". When the user opts do the former, he is not to be bothered to click OK on this dialog as well.

schwabe:
yes the change will change to something else when the auth doesn't complete.

selvanair:
force pushed with following changes
(i) The bug fix commit removed -- already merged
(ii) The dialog is now auto-closed only if the state changes to something other than GET_CONFIG.

lstipakov:
I haven't tested auto-close, but tested that dialog is displayed, doesn't disappear on GET_CONFIG and, when response is not required (tested by modifying flags in debugger), response-related fields are hidden.

openvpn.c
```diff
@@ -764,6 +778,20 @@ GenericPassDialogFunc(HWND hwndDlg, UINT msg, WPARAM wParam, LPARAM lParam)
...
+    case WM_OVPN_STATE: /* state change message is received from OpenVPN daemon */
+        /*
+         * AUTH_PENDING immediately transitions to GET_CONFIG until
+         * auth succeeds or connection restarts/aborts.
+         * Close the CR_TEXT dialog if state changes to anything
+         * other than GET_CONFIG. The state is in lParam.
+         */
+        param = (auth_param_t *) GetProp(hwndDlg, cfgProp);
+        if ((param->flags & FLAG_CR_TYPE_CRTEXT)
+            && strcmp((const char*) lParam, "GET_CONFIG"))
```
lstipakov, line 790: space between `char` and `*`

lstipakov (approved):

openvpn.c
selvanair: fixed

PR #442, opened 2021-08-15
Bug fix for challenge string parsing

selvanair:
This is the same as commit 3 of #441, but issuing separately as this is a bugfix, has no Windows quirks, and could get a quick review .. wink-wink.

Fixes parsing of the challenge text that could contain the
delimiter ':'


lstipakov (approved):
Tested both cases (dynamic CR as part of URL profile import), works as indented. Also thanks for parse_crtext() simplification - it has only one colon delimiter, no need for strtok loop.

PR #443, opened 2021-08-19
Add missing calling convention

lstipakov:


mattock:
I ran into this problem which make @lstipakov look into it. I tested this fix and it works.

PR #446, opened 2021-08-28
Import command

selvanair:
Add a command-line option
`--command import full-path-of-config` 
(alternatively `--import full-path` also works)

Could be called from a second instance or first instance.

Rationale: `openvpn-gui.exe --command import path` could be used to import a downloaded config. Setting this as the default command for `.ovpn` files would make easy import by double click, right click and choose default action or even automatically with some browser co-operation. 

TODO: set this command as the default shell action during installation.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -175,6 +175,9 @@ exit
 rescan
      Rescan the config folders for changes
 
+import ``path``
+     Import the config file pointed to by ``path``.
+
 If no running instance of the GUI is found, these commands do nothing
 except for *--command connect config-name* which gets interpreted
 as *--connect config-name*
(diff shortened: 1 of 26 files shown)


selvanair:
Any thoughts on this feature and about setting the default shell action to --command import "%1"?

Also, anyone knows how to make browsers automatically open a downloaded file based on file associations and MIME database? Chrome on Windows does it after a one time user intervention (always open files of this type), but not out of the box.

lstipakov:
Code looks good. Compiled and tested - works as intended.

lstipakov (approved):

PR #448, opened 2021-08-31
Import from a generic url

selvanair:
Extend the "import from AS" functionality to general URLs.

Notes:
Scanning the profile name from tags in the data, and dynamic
response support are same as in the implementation for AS.
These features are usable outside of AS although we may want
to support more generic tag names (not done).

To avoid input errors leading to download and import of arbitrary 
web content, the content-type is required to be 
`application/x-openvpn-profile`


as.c
```diff
@@ -417,13 +446,27 @@ DownloadProfile(HANDLE hWnd, const WCHAR *host, const char *username, const char
         goto done;
     }
 
+    /* check content-type if specified */
+    if (strlen(comps->content_type) > 0)
+    {
+        char tmp[256];
+        DWORD len = sizeof(tmp);
+        BOOL res = HttpQueryInfoA(hRequest, HTTP_QUERY_CONTENT_TYPE, tmp, &len, NULL);
```
lstipakov, line 454: Any reasons not to use W variant?

lstipakov (approved):
I haven't fully tested "Import from URL" functionality, since I got stopped by "HTTP content-type mismatch" error, because AS REST API returns "text/plain" as content type. AS profile import still works.

as.c
selvanair: Just not to give the impression that we support arbitrary unicode characters in this header string. I think MIME-types have to be us-ascii.

selvanair:
@lstipakov Thanks for the review.

After downloading a profile with correct user and password, doing it again with an empty password succeeds. 

It seems InternetSetOption for username or password with an empty string does nothing, and somehow any previously set value is retained even in a newly opened handle. After the first call, InternetOpen() seems to just duplicate some internal structure with any previously set values.

Do we want to keep this "feature", or should we somehow wipe the credentials?

lstipakov:
Nice catch. One way to fix it would be to make password mandatory, like it used to be :) But yeah, I think we should wipe the credentials.

selvanair:
Empty password has some use cases -- like authentication is federated so you don't want the user to reveal the password to the server. It requires something like open_url to properly support it, so moot in this case...

In any case its a good practice to wipe passwords from memory. Will make a patch.

PR #461, opened 2021-10-27
Fix typo in openvpn-gui-res-ua.rc

playday3008:


lstipakov (approved):
I checked with the native Ukrainian speaker and it looks good.

Issue #463, opened 2021-11-10
Ver 2.5.4-gui on Win 10: after connection GUI dialog immediately minimizes to icon taskbar tray-connected status icon will disappear when clicked


mapper14:
Installed OpenVPN-2.5.4-I604-amd64.msi on Win 10 Build 19043.
I'm using config scripts supplied by a paid VPN service.
After successful connection to server, GUI dialog immediately minimizes to an OpenVPN solid shaded icon indication active connection.
Clicking on this icon to reopen GUI dialog causes icon to disappear with no GUI dialog reopening.
After icon disappears connection persists but user has lost control to disconnect or perform any action on current connection.
So far, my only solutions to closing the connection are:
1 - Kill the OpenVPN Daemon process in Win Taskmanager
2 - Reboot
Previous builds of OpenVPN I've installed and used, 2.4.0, 2.4.2, on Windows 7 and 10 did not immediately minimize, used Hide button to minimize,  and clicking the minimized icon would reopen the GUI dialog to manually disconnect.

selvanair:
Check task manager to see whether the GUI is dead or still running.  If its running, the icon may be in the "hidden icons" list of the tray.

If GUI is really dead, check event log for any errors logged. Also post the tail end of the log file in %USERPROFILE%/OpenVPN/log/<config-name>.log -- it will show at what point the GUI (management client) disconnected.

mapper14:
When the icon disappears, my Win 10 Taskmanager always shows the task "OpenVPN Daemon" is still running, and I'm remain connected to the VPN server and am obtaining a VPN Proxy IP that I expect - by doing a quick check on WhatsMyIP
The Icon is definitely not hidden in 'hidden task' popup in the taskbar. To make absolute certain I've set Windows Tasksettings options to display ALL minimized task, dialogs or app in my Taskbar.  It's just simply NOT there and I have no access back to the OpenVPN-gui dialog.

Here's the last log file but understand that the event at 2021-11-10 02:39:12 is a manual END of the OpenVPN Daemon.  I was still connected to that Montreal server up to that point.

[Canada-Montreal_2_TCP444_SMART.log](https://github.com/OpenVPN/openvpn-gui/files/7599955/Canada-Montreal_2_TCP444_SMART.log)

Was this fixed in #464?

selvanair:
> Was this fixed in #464?

[...]

PR #464, opened 2021-11-12
Fix crash when clicking on tray icon after importing the first profile

lstipakov:
When clicking on tray icon, menu items are deleted and then recreated.
Deletion uses o.num_config:

    for (i = 0; i < o.num_configs; i++)
        DestroyMenu(hMenuConn[i]);

Commit 8e4183f9 ("Add '--command import' command line option")
added BuildFileList() call which modifies o.num_configs
but doesn't touch menus. When clicking on tray icon after import,
abovementioned code attemps to access invalid item in hMenuConn array
and crashes when this is the first imported profile and hMenuConn is NULL.
In other DestryMenu is called with invalid argument.

Fix by recreating popup menus instead of just rescan file list -
this will first delete menus with correct o.num_config value.

### DIFF
--- a/misc.c
+++ b/misc.c
@@ -704,6 +704,6 @@ ImportConfigFile(const TCHAR* source)
     }
 
     ShowTrayBalloon(LoadLocalizedString(IDS_NFO_IMPORT_SUCCESS), fileName);
-    /* rescan file list after import */
-    BuildFileList();
+    /* destroy popup menus, based on existing num_configs, rescan file list and recreate menus */
+    RecreatePopupMenus();
 }
--- a/tray.c
+++ b/tray.c
@@ -332,7 +332,7 @@ DestroyPopupMenus()
 }
 
 /* Rescan config folders and recreate popup menus */
-static void
+void
 RecreatePopupMenus(void)
 {
     DestroyPopupMenus();
--- a/tray.h
+++ b/tray.h
@@ -45,6 +45,7 @@
 #define IDM_CLEARPASSMENU       (1 + IDM_PASSPHRASEMENU)
 #define IDM_RECONNECTMENU       (1 + IDM_CLEARPASSMENU)
 
+void RecreatePopupMenus(void);
 void CreatePopupMenus();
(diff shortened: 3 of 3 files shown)


selvanair (approved):
The decision to only rebuild the file list and not call RecreatePopupMenus() was a deliberate "optimization" that turned into a bug. Thanks for the analysis and the fix -- looks good.

lstipakov:
@mattock If it doesn't take much time, I think we should make I604. This bug is not critical, but not crashing after importing the first profile would be a good thing.

cron2:
Hi,

On Fri, Nov 12, 2021 at 07:17:38AM -0800, Lev Stipakov wrote:
> @mattock If it doesn't take much time, I think we should make I604. This bug is not critical, but not crashing after importing the first profile would be a good thing.

Or we just go to 2.5.5, with the DHCP/.0 fix plus all the previous
windows fixes.

gert

-- 

selvanair:
> Hi,
> On Fri, Nov 12, 2021 at 07:17:38AM -0800, Lev Stipakov wrote: @mattock If it doesn't take much time, I think we should make I604. This bug is not critical, but not crashing after importing the first profile would be a good thing.
> Or we just go to 2.5.5, with the DHCP/.0 fix plus all the previous windows fixes. gert

And "import" in context menu of .ovpn (right-click import): the PR in openvpn-build has an "approval" but not merged..

mapper14:
I don't see any further status on PR#464.
Selvanair says changes are approved.
I reinstalled 2.5.4 1604 and icon crash behavior is still there.
Is a fix now in a future build, perhaps 2.5.5 ?

selvanair:
As mentioned at the end of this comment: https://github.com/OpenVPN/openvpn-gui/issues/463#issuecomment-978819625
this fix be available in the next release.

That said, this only fixes the crash soon after a profile is imported -- it does not cause a crash every time a connection is made as you reported in #463, which may be specific to your local installation.

mapper14:
Thx for the clarification.

PR #465, opened 2021-11-17
Show a prompt during profile import using --import

selvanair:
The user is prompted with a message showing the config
name that will be imported. The user can accept or cancel
the operation.

If the user was already prompted for over-write permission
because a config with the same name exists, no further dialog
is shown.

Import using the menu (Import File...) is not affected.

Rationale:
We want to set "Import" as the default verb for the context
menu of .ovpn files. This will allow import of configs by
double-click. Also when .ovpn file is downloaded using a browser,
setting the default browser action to "open" will result in an import.
In such cases a silent import action could be surprising, and a
prompt showing what is being imported could provide a better UX.

On the flip-side, the prompt/dialog will also be shown when import
is done from the context menu of .ovpn by "right click and
choose import" or when "openvpn-gui.exe --import foo"
or "openvpn-gui.exe --command import foo" is executed. As import
is an action that does not result in an immediately visible result
(unlike, say, edit or print), a prompt requiring user action is of
some value even in these cases. At worst it's a minor annoyance.

See also: OpenVPN/openvpn-build#227 and discussions there-in

### DIFF
--- a/as.c
+++ b/as.c
@@ -682,7 +682,7 @@ ImportProfileFromURLDialogFunc(HWND hwndDlg, UINT msg, WPARAM wParam, LPARAM lPa
             if (downloaded) {
                 EndDialog(hwndDlg, LOWORD(wParam));
 
-                ImportConfigFile(path);
+                ImportConfigFile(path, false); /* do not prompt user */
                 _wunlink(path);
             }
             return TRUE;
--- a/main.c
+++ b/main.c
@@ -471,7 +471,7 @@ HandleCopyDataMessage(const COPYDATASTRUCT *copy_data)
     }
     else if(copy_data->dwData == WM_OVPN_IMPORT && str)
     {
-        ImportConfigFile(str);
+        ImportConfigFile(str, true); /* prompt user */
     }
     else if (copy_data->dwData == WM_OVPN_NOTIFY)
     {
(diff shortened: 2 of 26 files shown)


lstipakov (approved):
Tested, works as expected.

The only thing I would change is prompt - I think if we ask user for decision, it should be presented as question, like "Do you want to import profile <clie.ovpn>"? yes/no. 

At the moment it looks a bit odd to my taste (Importing profile/ok-cancel). But feel free to disagree.

selvanair:
> Tested, works as expected.
> 
> The only thing I would change is prompt - I think if we ask user for decision, it should be presented as question, like "Do you want to import profile <client.ovpn>"? yes/no.
> 
> At the moment it looks a bit odd to my taste (Importing profile/ok-cancel). But feel free to disagree.

I was thinking of it as a compromise between progress information and request for user consent, as one could land here after various actions, some with explicit intent to import, some less so .  

I'm okay to change the text to the question you suggested. Should it be "profile" or "config"? -- we have not been consistent in the past: we have "config" in many strings and "profile" in a few.

cron2:
Hi,

On Thu, Nov 18, 2021 at 07:58:46AM -0800, Selva Nair wrote:
> I'm okay to change the text to the question you suggested. Should it be "profile" or "config"? -- we have not been consistent in the past: we have "config" in many strings and "profile" in a few.

The rest of the OpenVPN "client" ecosystem (Android etc) seems to talk
about "profiles" more than "configs"...

gert
-- 

selvanair:
Changed the prompt to a question with YES/NO. Fixup commits to be squashed before merge.

lstipakov (approved):

Issue #466, opened 2021-11-18
Ver 2.5.4-gui on Win 10: unable to change password of PKCS#12 certificate


BlackBeretEH:
Hi, 

When I've tried to change certificat password through OpenVPN GUI, it failed with error below (see screenshot).
I've reinstalled previous version and it's works, so it seems to be related to 2.5.4 version. 

![Clipboard - 17 novembre 2021 10_34](https://user-images.githubusercontent.com/86951328/142409758-cc7df74c-44af-4aa4-93b0-1e8f6aece7c3.png)

Have a nice day !

lstipakov:
Thanks for reporting this. The fix will be in the next release.

Here is the binary with the fix you might want to try: https://github.com/OpenVPN/openvpn-gui/suites/4395661066/artifacts/115917818

BlackBeretEH:
> Thanks for reporting this. The fix will be in the next release.
> 
> Here is the binary with the fix you might want to try: https://github.com/OpenVPN/openvpn-gui/suites/4395661066/artifacts/115917818

Thanks for the support !

PR #467, opened 2021-11-18
Fix broken "change password" functionality

lstipakov:
When we link with natively-built OpenSSL .DLLs
(not cross compiled with MinGW), we are expected to include
applink.c, which provides glue between OpenSSL BIO layer
and compiler run-time. This doesn't apply to ARM64.

Failure to do that results in "no OPENSSL_Applink" fatal error
during password change.

See the corresponding fix in openvpn2:

https://sourceforge.net/p/openvpn/mailman/message/37361982/

### DIFF
--- a/passphrase.c
+++ b/passphrase.c
@@ -28,6 +28,11 @@
 #include <openssl/pem.h>
 #include <openssl/err.h>
 #include <openssl/pkcs12.h>
+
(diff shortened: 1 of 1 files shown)


lstipakov:
This fixes https://github.com/OpenVPN/openvpn-gui/issues/466

selvanair (approved):

Issue #468, opened 2021-11-27
Popup challenge response window size


ziedrv:
Hello,
I hope this can be fixed,. A long challenge text must be displayed at the end use but the window size doesn't fit to show all the message. The popup size is not dynamic.

**Challenge text:** To sign in, use a web browser to open the page https://microsoft.com/devicelogin and enter the code XXXXXXXXX to authenticate.

![141252](https://user-images.githubusercontent.com/24189414/143683234-840da4ed-e665-45bb-93c0-45bc2c3a9604.PNG)

selvanair:
That particular dialog (dynamic challenge response) is already resizable. 

The problem may be that your text is getting wrapped and  there is no vertical space for a second line. We can increase the line length and allow space for a second line. Will test and see.

That said, giving a URL in the challenge text will provide a poor UX as the user cannot even cut and paste the URL.  See whether the new client-pending-auth management command that can ask the client to open a URL works better for this.

ziedrv:
As mentioned in management note, client-pending-auth is a good idea. I am trying to implement it in python code it but it will took time because there is a little information and no examples how to use this option (It's a new feature in version 2.5).

I confirm that I can resize the popup window manually but text still fixed.
![412552](https://user-images.githubusercontent.com/24189414/143721882-741af013-dc0d-4c0b-a0ab-63f0b940dae3.PNG)

selvanair:
The maximum width of the line is currently limited to ~60 characters, and vertical space assigned is just enough for one line. The text wraps at that point as you can see from some text fragments (black pixels) showing up on line 2. This will be fixed in #469. 

For client-pending-auth we currently support openurl in the GUI. For your use case it should be usable if the device code can be submitted as GET.

schwabe:
@ziedrv see for example this script: https://github.com/schwabe/openvpn/commit/0976079d8bbd078fd1a1dc902ad1a3b03173b6b0

selvanair:
@ziedrv You can test the fixed version using CI-build executables here: https://github.com/OpenVPN/openvpn-gui/actions/runs/1513839435#artifacts

ziedrv:
Did I miss something ?

![1000212](https://user-images.githubusercontent.com/24189414/143926633-8de1c3c7-00c1-44f2-9753-496e815d1364.PNG)

ziedrv:
OK it's the vcruntime version. I this it should be the same as the one installed vcruntime140.dll

[...]

PR #469, opened 2021-11-27
Provide more space for challenge dialog text

selvanair:
We do use a re-sizable dialog box for dynamic challenge-response
to cater for potentially long lines of challenge text. But the
space specified for the widget is enough for only a single short line
(~60 characters) of text.

Increase the horizontal and vertical space to allow for up to
two lines of ~120 characters per line.

The default size of the Window is not changed. But it is
automatically resized if the space required for the text
is longer than the window width minus some margin. The max
horizontal size of the window is capped at 640 nominal pixels
as longer text will be wrapped in to two lines.

Github issue #468

Auto-resized example:
cr3


lstipakov:
Tested locally that auto-resize works.

lstipakov (approved):

PR #471, opened 2022-01-24
Add openssl3 support for msvc build

lstipakov:
Add openssl3 support for msvc build

 - remove vcpkg manifest and use whatever openssl
 version is installed. To build with openssl3, one could
 use openssl3 port from openvpn/contrib/vcpkg-ports.

 - build with openssl1.1.1 and openssl3 in GitHub Actions

### DIFF
--- a/.github/workflows/msbuild.yml
+++ b/.github/workflows/msbuild.yml
@@ -7,13 +7,22 @@ jobs:
     strategy:
       matrix:
         arch: [x86, amd64, amd64_arm64]
+        ossl: [openssl, openssl3]
+        include:
+        - arch: amd64_arm64
+          triplet: arm64
+        - arch: x86
+          triplet: x86
+        - arch: amd64
+          triplet: x64
 
     env:
       # Indicates the location of the vcpkg as a Git submodule of the project repository.
       VCPKG_ROOT: ${{ github.workspace }}/vcpkg
       SOLUTION_FILE_PATH: .
       BUILD_CONFIGURATION: Release
-      buildDir: '${{ github.workspace }}/build'
+      VCPKG_OVERLAY_PORTS: '${{ github.workspace }}/openvpn/contrib/vcpkg-ports'
+      buildDir: '${{ github.workspace }}/build/'
 
     runs-on: windows-latest
     steps:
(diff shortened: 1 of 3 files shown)


selvanair (commented):
I'm not familiar with vcpkg or its ports, but can we avoid adding files with hard-coded library versions (e.g., OpenSSl 3.0.1) ? Building with a newer version of OpenSSL would require changes hare and them matching changes in OpenVPnn core too, isn't it? That doesn't sound the right approach to me.

Can we have a common vcpkg port that could be used here as well as in the core. Or add this to openvpn-build?

lstipakov:
Ideally we should not host any ports ourselves and use openssl3 port provided by vcpkg official repo, which is already long overdue. https://github.com/microsoft/vcpkg/issues/20031

We could use vcpkg port from openvpn repo, but that would require dependency to openvpn when building openvpn-gui with openssl3. This should not be a problem for release building since we checkout all repos. For GitHub Actions we could have a bit ugly solution of cloning openvpn before building openvpn-gui. I think we can live with it until openssl3 is added into vcpkg.

I am also thinking that we need to support building openvpn-gui with both openssl1.1.1 and openssl3. Right now we use vcpkg manifest mode (the recommended way), where we explicitly list dependencies which are automatically installed, and we list openssl3. Maybe we should drop it and use whatever openssl library is installed.

lstipakov:
Also let GitHub Actions build with both flavors - openssl1.1.1 and openssl3.

cron2:
Hi,

On Mon, Jan 24, 2022 at 11:33:16PM -0800, Lev Stipakov wrote:
> I am also thinking that we need to support building openvpn-gui with both openssl1.1.1 and openssl3. Right now we use vcpkg manifest mode (the recommended way), where we explicitly list dependencies which are automatically installed, and we list openssl3. Maybe we should drop it and use whatever openssl library is installed.

Is *anything* "installed" on a MSVC build environment by default?

MinGW builds have no "default" OpenSSL (as it needs to be cross-built)

gert


-- 

lstipakov:
No, nothing is installed. But that's not an issue - `vcpkg install openssl` and `vcpkg install openssl3` do the thing. And with caching it doesn't take time.

cron2:
Hi,

On Tue, Jan 25, 2022 at 12:11:03AM -0800, Lev Stipakov wrote:
> No, nothing is installed. But that's not an issue - `vcpkg install openssl` and `vcpkg install openssl3` do the thing. And with caching it doesn't take time.

You write

  "Maybe we should drop it and use whatever openssl library is installed."

so, if "nothing is installed", that confuses me.

gert

-- 

lstipakov:
Ah, I mean whatever is "somehow" installed before building gui - the same thing we do with openvpn, where one needs to run `vcpkg install openssl` or `vcpkg install openssl3` before building openvpn. 

For gui it was different - we listed dependencies (openssl) which got automatically installed as part of gui build process. But since we need to build both flavors (with openssl and openssl3), this approach doesn't work anymore.

lstipakov:
Force-pushed a new version.

lstipakov:
Gentle ping @selvanair. Note that there are no changes (anymore) to the code.

.github/workflows/msbuild.yml
```diff
@@ -23,23 +32,32 @@ jobs:
...
-      uses: lukka/run-vcpkg@v6
+    - name: Clone openvpn repo
+      uses: actions/checkout@v2
       with:
-        setupOnly: true
-        vcpkgGitCommitId: 'a267ab118c09f56f3dae96c9a4b3410820ad2f0b'
-        additionalCachedPaths: ${{ env.buildDir }}/vcpkg_installed
-        appendedCacheKey: ${{ matrix.arch }}-${{ hashFiles( '**/vcpkg.json' ) }}
+        # TODO: replace with openvpn/openvpn
+        repository: lstipakov/openvpn
```
selvanair, line 39: I guess this is because openssl3 port is not merged to openvpn/openvpn yet.

selvanair (commented):
Looks good except for the openvpn repo name. Shall we wait for the openssl3 port to get merged to OpenVPN?

By the way, we could add a few lines in BUILD.rst about build using vcpkg. Need not be in this PR, though.

.github/workflows/msbuild.yml
lstipakov: Yep, https://patchwork.openvpn.net/patch/2254/

Someone needs to ack it (wink wink).

lstipakov: Acked, merged. Repo name fixed.

selvanair (approved):
Looks good now.

PR #472, opened 2022-02-02
wcstok: use C standard version

lstipakov:
Replace old wcstok signature with C standard version,
which keeps state between invocations in "context" parameter
instead of internal per-thread context.

This allows to get rid of _CRT_NON_CONFORMING_WCSTOK
define in CMakeLists.txt

Reported-by: Kai Schtrom


selvanair (commented):
The three arg version of wcstok is C-standard, but Windows had a wrong function signature (two args only) until recently. As a consequence, the run-time targetted by mingw does not/cannot support it. We can't change this without breaking mingw build.

Anyway, the two arg version is thread-safe and that's all what we need. Just keep  _CRT_NON_CONFORMING_WCSTOK in msvc build.

lstipakov:
I changed to `wcstok_s`, now both mingw and msvc are happy.

selvanair (approved):
LGTM

Issue #474, opened 2022-02-02
Support for password change feature and OpenSSL 3.0


selvanair:
In the GUI we use OpenSSL only for changing password of privtae keys (PEM and pkcs12). With OpenSSL 3.0, this is broken for legacy key encryption algorithms (eg., 40bitRC2-CBC). We have a couple of options:
(i) load legacy & default providers at startup
(ii) add a configurable option to enable legacy provider 
My preference would be do (i) if fips is not enabled, do nothing otherwise. Easier to implement and no user action required.

cron2:
Hi,

On Wed, Feb 02, 2022 at 02:40:17PM -0800, Selva Nair wrote:
> In the GUI we use OpenSSL only for changing password of privtae keys (PEM and pkcs12). With OpenSSL 3.0, this is broken for legacy key encryption algorithms (eg., 40bitRC2-CBC). We have a couple of options:
> (i) load legacy & default providers at startup

Sounds like the easiest way forward.

gert
-- 
"If was one thing all people took for granted, was conviction that if you 
 feed honest figures into a computer, honest figures come out. Never doubted 
 it myself till I met a computer with a sense of humor."
                             Robert A. Heinlein, The Moon is a Harsh Mistress

Gert Doering - Munich, Germany                             ***@***.***

PR #476, opened 2022-02-09
Load legacy provider when required

selvanair:
Issue #474

### DIFF
--- a/main.c
+++ b/main.c
@@ -53,8 +53,7 @@
 #include "as.h"
 
 #ifndef DISABLE_CHANGE_PASSWORD
-#include <openssl/evp.h>
-#include <openssl/err.h>
+#include <openssl/crypto.h>
 #endif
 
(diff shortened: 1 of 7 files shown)


selvanair:
It seems appveyor build is pulling includes from OpenSSL 1.0.2 and failing to resolve OPENSSL_init_crypto. As both 1.1.1 and 1.0.2 seems to get installed, do we need to specify OPENSSL_CRYPTO_LIBS and OPENSSL_CRYPTO_CFLAGS to ensure the latter gets used during build? Or remove 1.0.2?

lstipakov (approved):
Compile-tested and checked that OSSL env variables are indeed set. Haven't tested password change functionality.

Looks good.

lstipakov:
I suggest to kick appveyor and add GitHub Actions with MinGW. I can do that.

Issue #482, opened 2022-03-10
User name passed to management interface as part of CRV1 response is not escaped


selvanair:
Reported by: Jakob Curdes <jc@info-systems.de>

Relevant log provided by Jakob Curdes <jc@info-systems.de>

After the user/pass dialog

`2022-03-10 19:51:15 us=937000 MANAGEMENT: CMD 'username "Auth" "authpoint\\JakobC"'`

User name with embedded backslash correctly escaped during first User/Auth exchange

```
2022-03-10 19:51:15 us=953000 MANAGEMENT: CMD 'password [...]'

(...)

2022-03-10 19:51:16 us=218000 Control Channel: TLSv1.2, cipher TLSv1.2 ECDHE-RSA-CHACHA20-POLY1305, peer certificate: 2048 bit RSA, signature: RSA-SHA256
2022-03-10 19:51:16 us=218000 [Fireware SSLVPN Server] Peer Connection Initiated with [AF_INET][85.220.177.78:443](http://85.220.177.78:443/)
2022-03-10 19:51:17 us=406000 MANAGEMENT: >STATE:1646938277,GET_CONFIG,,,,,,
2022-03-10 19:51:17 us=406000 SENT CONTROL [Fireware SSLVPN Server]: 'PUSH_REQUEST' (status=1)
2022-03-10 19:51:17 us=453000 AUTH: Received control message: AUTH_FAILED,CRV1:R,E:1285:YXV0aHBvaW50XEpha29iQw==:Type "p" to receive a push notification or type your one-time password


2022-03-10 19:51:17 us=453000 TCP/UDP: Closing socket
2022-03-10 19:51:17 us=453000 SIGUSR1[soft,auth-failure] received, process restarting
2022-03-10 19:51:17 us=453000 MANAGEMENT: >STATE:1646938277,RECONNECTING,auth-failure,,,,,
2022-03-10 19:51:17 us=453000 Restart pause, 5 second(s)
2022-03-10 19:51:24 us=812000 MANAGEMENT: CMD 'username "Auth" "authpoint\JakobC"'
```
User name not escaped during the CRV1 exchange.

selvanair:
Fixed by #483

PR #483, opened 2022-03-11
Fix passing username for CRV1 response

selvanair:
Escape the username string before passing to management
interface. For other dialogs this is already done.

Move string-escape to a function and process the username
through it.
Also escape space, single quote in addition to double quote
and backslash.

Reported by: Jakob Curdes <jc@info-systems.de>

Issue: #482

### DIFF
--- a/misc.c
+++ b/misc.c
@@ -161,6 +161,46 @@ out:
     return retval;
 }
 
+/**
+ * Escape \ space ' and " in a string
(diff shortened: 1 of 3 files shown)


cron2:
Looks good to me.  Not tested, just stared at code (and compared previous implementation to new function).  ACK.

jc-ol:
I confirm the original issue is solved with this change. Jakob Curdes

PR #484, opened 2022-03-11
Check pointer is not NULL before passing to strlen

selvanair:
Fixing my own mistake in PR #483 


cron2:
Uh, yes, indeed.  Unlikely to be triggered, but still wrong.  ACK

Issue #487, opened 2022-03-30
error if quote (') in password


j2s2:
[openvpn-gui v11.27 / openvpn v2.5.6]

if we type a quote in the password field, we have a error in log : 
    ERROR: Options warning: Bad backslash ('\') usage in TCP:0...

Apparently it is the only character which leads to the error.

Sincerly

selvanair:
Thanks for reporting this --- I can reproduce it.
We sent the password in single quotes and also escape any single quote, double quote, backslash and space in the password. But it seems, OpenVPN core does not support escaping in single quote delimited phrases -- that means its  not possible to have embedded single quotes in suc strings. Someone familiar with escape processing in the core (options.c), please confirm. We recently fixed escape processing in username but we use double quotes there which works.

selvanair:
Much of what I wrote above turns out to be wrong -- so trying again.
We do send password in double quotes but the problem seems to be escaping single quotes. The core wants only double-quote, backspace, and space escaped, but I added escaping of single quote in a recent patch.

cron2:
Hi,

On Thu, Mar 31, 2022 at 07:28:51AM -0700, HeadOnPascal wrote:
> Had the same problem and downgraded to OpenVPN GUI 2.5.5. Here it's working. 

Yes, this is fallout of a bugfix (and code cleanup...) related to
character escaping in usernames.  Seems the "cleanup" bit had unexpected
consequences.

gert
-- 
"If was one thing all people took for granted, was conviction that if you 
 feed honest figures into a computer, honest figures come out. Never doubted 
 it myself till I met a computer with a sense of humor."
                             Robert A. Heinlein, The Moon is a Harsh Mistress

Gert Doering - Munich, Germany                             ***@***.***

selvanair:
The "Command Parsing" section in management-notes has some description of what needs to be escaped and lists only ", backslash and space.

Supporting escaped single quotes in the parser would be nice, but fixing this here may be easier/quicker/safer ?

cron2:
Hi,

On Thu, Mar 31, 2022 at 08:19:44AM -0700, Selva Nair wrote:
> The "Command Parsing" section in management-notes has some description of what needs to be escaped and lists only ", backslash and space.

So let's stick to this... "it's documented", after all :-)

gert
-- 
"If was one thing all people took for granted, was conviction that if you 
 feed honest figures into a computer, honest figures come out. Never doubted 
 it myself till I met a computer with a sense of humor."
[...]

PR #489, opened 2022-04-04
Do not escape single quote character

selvanair:
Commit 6271d2f67 added escaping of single quote in
quoted strings which is wrong. Fix it.

Ref: issue #487


cron2:
As discussed in the PR - makes sense, matches the documentation on what the core expects.  ACK.

PR #497, opened 2022-05-09
Use vcpkg manifest 

lstipakov:


mattock:
For what it's worth, I was able to build openvpn-gui using the new procedure on my "msibuilder" VM. I only tested one preset.

Code-vise I can't really give this a proper review. Nothing really obvious caught my eye.

lstipakov:
@selvanair @cron2 I know you guys are fans of MSVC build system, what do you think?

I think commit messages are reasonable enough and explain what had been done and why.

selvanair:
> @selvanair @cron2 I know you guys are fans of MSVC build system, what do you think?

You got that right -- lol
I can't say I understand the changes, but the PR looks sane, useful and documented. +1 from me.

selvanair (approved):

PR #501, opened 2022-05-26
CMakePresets: fix arm64 debug/release config

lstipakov:


mattock:
Prior to this PR the arm64 version of openvpn-gui ended always being a debug build that depended on DLLs that were not present in the installer or on a typical Windows installation. This PR fixes that problem so ACK.

PR #504, opened 2022-05-30
cmake: replace Ninja generator with MSVC

lstipakov:
Ninja makes build slightly faster but requires running
"x64" developer command prompt, not the "default" one.
Without that, cmake silenly produces x86 binaries and it might
take a while to find out the reason. To avoid confusion, switch back
to MSVC generator.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -5,6 +5,7 @@
             "name": "base",
             "hidden": true,
             "binaryDir": "${sourceDir}/out/build/${presetName}",
+            "generator": "Visual Studio 16 2019",
             "cacheVariables": {
                 "CMAKE_TOOLCHAIN_FILE": {
                     "value": "$env{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake",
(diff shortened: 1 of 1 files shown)


selvanair:
I'm not qualified to comment on this, yet, the dev command prompt sets env vars such as PATH and INCLUDE (runs varsall.bat xx?) which Ninja may not be doing automatically.  
That said the gha builds look right and 64 bit artifacts do work even before this patch. And, with this patch the gha built executables acquire a dependence on VCRUNTIME140D.dll which is not nice.

lstipakov:
Yeah my bad, with VS generator one needs to specify release/debug at build preset.

lstipakov:
Looks like fixed now.

mattock:
I build all the openvpn-gui architectures manually and checked that the resulting openvpn-gui was

1. Was of the correct architecture (32-bit or 64-bit)
2. Was a release build (included VCREDIST140.DLL, _not_ VCREDIST140D.DLL)

I'll wait approving this until I've built the new OpenVPN 2.5.7 installers and have tested them properly, even though I'm fairly certain that the problem is gone now.

mattock:
I tested 2.5.7 installers with openvpn-gui that used this new build process. All worked fine on x64 and arm64, no traces of the issues encountered in the Ninja times. ACK.

PR #506, opened 2022-06-06
Update of PR #327 (OpenVPN3 support)

facboy:
@lstipakov i merged master back into your PR #327 and addressed some of the comments from @selvanair.

is there interest in getting this merged?


selvanair:
IIRC, when the original PR was submitted, ovpnagent was not available publicly. Has that changed?

lstipakov:
ovpnagent is now available in source (as @facboy) mentioned and also binary form (GitHub Actions artifacts, for example https://github.com/OpenVPN/openvpn3/suites/6374869439/artifacts/231032114).

The interest in merging this is a bit less than before, since with (coming) dco-win most of functionality will be moved from openvpn2/3 into kernel. That said, I think we could review and merge it - assuming it is useful for community and doesn't complicate code much.

lstipakov:
Since we have switched to CMake:

 - do we really need those travis changes? I don't think we use travis anymore.

 - @facboy could you add required changed to CMakePresets.json? For example, add something like "ovpn3-x64-release-ossl3" - this way we ensure that openvpn3-related code is part of our build automation

 - @selvanair I don't have write access to this repo and cannot give an approval to run workflow (aka github actions build automation). Could you grant this approval (without merging yet :) ?

facboy:
> could you add required changed to CMakePresets.json? For example, add something like "ovpn3-x64-release-ossl3" - this way we ensure that openvpn3-related code is part of our build automation

done

lstipakov:
Yay, it works :) I was able to connect to corp vpn with dco-win driver using openvpn-gui and openvpn3 ovpnagent/omicliagent.

Prerequisites:

 - add `omi_exe_path` key at `HKLM\SOFTWARE\OpenVPN` of type `REG_SZ` pointing to the location of `omicliagent.exe` - the latter one could be downloaded from GHA.
 - install `ovpnagent.exe` as a service (which can also be downloaded from GHA):
`c:\Users\lev\Projects\ovpn3-build\ovpn3\core\out\build\x64-Debug-dco\openvpn\ovpnagent\win>ovpnagent.exe install`
Note that running `ovpnagent.exe` in a foreground will not work - `CreateProcessAsUser()` will fail.

One small thing I noticed is that settings dialog height has grown by 40px even if ENABLE_OVPN3 is not set. Can we not do it (and resize programmatically instead?) for existing configurations?

lstipakov:
Yay, it works :) I was able to connect to corp vpn with dco-win driver using openvpn-gui and openvpn3 ovpnagent/omicliagent.

Prerequisites:

 - add `omi_exe_path` key at `HKLM\SOFTWARE\OpenVPN` of type `REG_SZ` pointing to the location of `omicliagent.exe` - the latter one could be downloaded from GHA.
 - install `ovpnagent.exe` as a service (which can also be downloaded from GHA):
`c:\Users\lev\Projects\ovpn3-build\ovpn3\core\out\build\x64-Debug-dco\openvpn\ovpnagent\win>ovpnagent.exe install`
Note that running `ovpnagent.exe` in a foreground will not work - `CreateProcessAsUser()` will fail.

One small thing I noticed is that settings dialog height has grown by 40px even if ENABLE_OVPN3 is not set. Can we not do it (and resize programmatically instead?) for existing configurations?

facboy:
hmm...i suspect that is beyond me!

i believe if you run `ovpnagent` as admin it works - that's how i observe the logs when debugging.

facboy:
i think this does it.

lstipakov:
I am revamping #327 based on @facboy's work - will rebase on top of master, remove merge commits and squash fixes.

This also works:

```
#ifdef ENABLE_OVPN3
#define ADVANCED_DIALOG_HEIGHT 320
#else
#define ADVANCED_DIALOG_HEIGHT 280
#endif
```

and then 

```
/* Advanced Dialog */
ID_DLG_ADVANCED DIALOGEX 6, 18, 252, ADVANCED_DIALOG_HEIGHT
```

facboy:
kk, closed.

Issue #158, opened 2017-04-17
PKCS#11: Selecting smartcards dynamically via dialog


EugeneKin:
Hi folks!
Are there any plans to implement "pkcs11-id-management" feature?

With latest 11.5 version i get "GUI> Error: Received NEED-STR message -- not implemented"

selvanair:
Yes, its in my TODO list but not a high priority as no one seems to have missed it so far :)

EugeneKin:
Got it, thanks.
OpenVPN Manager has this feature since "stone age", but is slightly outdated now, last release was 4 years ago.
It does not support new "OpenVPN Interactive Service". Works with Openvpn 2.4 in appmode (openvpn.exe) with admin rights only.

tefod-zz:
After a day of research, I have now also stumbled across this.
Issue;
Want to use yubikey hardware tokens with imported certificates in the PIV storage. Works great together with installed OpenSC as PKCS11 driver:
```
pkcs11-providers 'c:\Program Files\OpenSC Project\PKCS11-Spy\pkcs11-spy.dll'
pkcs11-id 'pkcs11:model=PKCS%2315%20emulated;token=acme;manufacturer=piv_II;serial=91fca8502536985d;id=%04'
```
But: to get the hardware PIV compliant the serial (CHUID) from each token has to be unique - this is done by the YubiKey Manager or with the CLI Tools automatically!
That means on every change in the PIV storage (e.g. adding a new certificate) this number changes. Even if you delete and add the same certificate! 
This would lead in a company environment with dozens of users, that you have to prepare and deploy for each employee a own OVPN - this is unfortuantely not maintainable nor user friendly.

Its a pitty that the `pkcs11-id-management` has not been added or that we can't select the key per slot or with placeholders etc.

Did someone found a workaround for this? I can't believe, that the OpenVPN Gui is not used with hardware tokens in a company environment as foreseen.

selvanair:
I distribute user-specific profiles, and have no idea how to avoid that. The GUI now supports importing ovpn via a URL, which makes it easier than before.
That said, have you tried using this token with `--cryptoapcert ...` instead? If an approproate driver is installed, Windows should automatically load certificates in the token to the store. I find that easier to administer.

tefod-zz:
@selvanair that's really awesome. Just tried it right now - works flawlessly. This has some great benefits; i dont need to deploy OpenSC. Would work even without the YubiKey Minidriver (is been detected by Windows). 
[...]

PR #508, opened 2022-06-28
Support pkcs11-id-management

selvanair:
Features
- shows a list of pkcs11 certificates for user to select from
- double click on any entry to see more details of the certificate
- suggest to insert a token if no certificates found
- retry button to reload certificate list

Requires --pkcs11-id-management in the config file.

Issue #158

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,6 +50,7 @@ target_link_libraries(${PROJECT_NAME} PRIVATE
     Gdi32.lib
     Comdlg32.lib
     Ole32.lib
+    Cryptui.lib
     Wininet.lib)
 
 target_include_directories(${PROJECT_NAME} PUBLIC ${CMAKE_CURRENT_BINARY_DIR})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,6 +28,7 @@ add_executable(${PROJECT_NAME} WIN32
     tray.c
     viewlog.c
     as.c
+    pkcs11.c
     res/openvpn-gui-res.rc)
 
 find_package(OpenSSL REQUIRED)
--- a/Makefile.am
+++ b/Makefile.am
@@ -95,6 +95,7 @@ openvpn_gui_SOURCES = \
 	env_set.c env_set.h \
 	echo.c echo.h \
 	as.c as.h \
+	pkcs11.c pkcs11.h \
 	openvpn-gui-res.h
 
 openvpn_gui_LDFLAGS = -mwindows
--- a/configure.ac
+++ b/configure.ac
@@ -76,6 +76,8 @@ case "$host" in
 		;;
 esac
 
+AC_CHECK_LIB(cryptui, CryptUIDlgViewContext)
+
 AC_PROG_CPP
 AC_PROG_INSTALL
 AC_PROG_LN_S
--- a/localization.c
+++ b/localization.c
@@ -101,6 +101,57 @@ SetGUILanguage(LANGID langId)
     gui_language = langId;
 }
 
+static int
+LocalizedSystemTime(const SYSTEMTIME *st, wchar_t *buf, size_t size)
+{
+    int date_size = 0, time_size = 0;
+    LCID locale = MAKELCID(GetGUILanguage(), SORT_DEFAULT);
+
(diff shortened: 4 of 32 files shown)


lstipakov:
I am not too much into pkcs11, is there any way I could test this?

Also shouldn't we update all language resources with the new dialog?

selvanair:
> I am not too much into pkcs11, is there any way I could test this?
> 
> Also shouldn't we update all language resources with the new dialog?

I thought of getting  some feedback before copying resources -- in case the UI needs change/improvement. 

Once could use softhsm to make a software token and test, but I have never tried this on Windows. The following is based on the steps I follow on Linux. Softhsm2 and pkcs11-tool (from opensc) are available for windows, so should work there as well

```
pkcs11-tool.exe --module path-to-lib\softhsm2-x64.dll --init-token --label SoftToken  
```

load existing client.key and client.cert files to the token

```
pkcs11-tool.exe --module path-to-lib\softhsm2-x64.dll --login -w client.key --type privkey --id 1
pkcs11-tool.exe --module path-to-lib\softhsm2-x64.dll --login -w client.cert --type cert  --id 1 
```
use same id for matching key and cert to match. Add more key-pairs each pair with a unique id.

Test as
`openvpn--show-pkcs11-ids path-to-lib\softhsm2-x64.dll`

Add to openvpn config
`pkcs11-providers path-to-lib\softhsm2-x64.dll`

selvanair:
Alternatively, the following appears to work:

Install only softhsm2 [note]. Then unzip the following to its `var\softhsm2\tokens\`  folder.

https://drive.google.com/file/d/1-DJRGXEyZBhEgGC8vba9oWI9-Jfqcy6D/view?usp=sharing

This zip file contains a "soft token" I created with sample certificates and keys in OpenVPN repo.  The so and user PINs are both set to 1234.

`openvpn --show-pkcs11-ids <softhsm-path>\lib\softhsm2-x64.dll` should then display certificates available.

But, to connect using this, the server has to obviously accept keys signed by the sample ca.

[note]
If using the portable zip version of softhsm2, set SOFTHSM2_CONF to `<softhsm-path>\etc\softhsm2.conf` and edit that file to point directories.tokendir to `<softhsm-path>\var\softhsm2\tokens`). No other setup needed.

lstipakov:
Sorry for the delay. I followed @selvanair's instructions and got dialog with certificates in GUI. However, when I selected certificate, openvpn process crashed.
xkey_bug

Not sure if crash is related to this change, though.

selvanair:
As the crash is in openvpn.exe and seems to shwo the sign() callback address in xkey_provider is 0xFFFFFFFFFF, this has to be unrelated. But this should not happen -- what version of openvpn.exe is this?
Can you try with pkcs11-id-management replaced by 'pkcs11-id <ID>' in the config file where <ID> is the one reported by `openvpn.exe --show-pkcs11-ids <provider-dll>` ? Or copy the ID from the log file. If that shows the same crash, we know its unrelated to the GUI.

lstipakov:
Yes I got the same crash. I use locally-built openvpn master.

selvanair:
I will try to reproduce -- did we break xkey-provider + pkcs11-helper !?

selvanair:
@lstipakov  I'm unable to reproduce this using mingw built master + OpenSSL 3.0.1. The OpenSSL lib for Windows I use is an slightly old build  though (From Dec 2021). I suppose plain --cert and --key options do work with your build. Does cryptoapicert work? That also uses xkey-provider and the sign_op function pointer. 

The value of 0xFFFFFFFF you see in debugger  is probably -1, but  that's bizarre as its a function pointer passed through OpenSSL params (as octet pointer).

lstipakov:
@selvanair I also managed to reproduce it with GHa binaries from master (https://github.com/OpenVPN/openvpn/actions/runs/2579097466).

lstipakov:
(I'll be on vacation till 17.7, I can have a look into xkey-provider thing if we haven't figure it out by then)

selvanair:
Patch for sent to the devel list should hopefully fix it.

lstipakov:
I finally managed to test it against the latest master, including abovementioned patch, works as expected!

lstipakov (approved):

PR #509, opened 2022-06-30
Automatically choose management interface port number

selvanair:
Bind a socket to a dynamic port and then close to identify
a free port and use it when starting openvpn.exe.

Trac: 1051

### DIFF
--- a/misc.c
+++ b/misc.c
@@ -773,3 +773,54 @@ set_openssl_env_vars()
         }
     }
 }
+
+/*
+ * Find a free port to bind and return it in addr.sin_port
+ */
+BOOL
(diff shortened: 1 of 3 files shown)


selvanair:
Updated to retain the current behavioir when possible. Try port = user-defined-offset + config-index first, falling back to a free ephemeral port if the former fails.

misc.c
```diff
@@ -773,3 +773,54 @@ set_openssl_env_vars()
         }
     }
 }
+
+/*
+ * Find a free port to bind and return it in addr.sin_port
+ */
+BOOL
```
lstipakov, line 780: We do not seem to use return value.

lstipakov (approved):

PR #512, opened 2022-07-08
Support controlling persistent (pre-started) connections

selvanair:
This PR implements connecting the GUI to instances started by automatic service or by any other means.

The GUI scans config files in the `config-auto` directory and attaches to its management i/f if available.

-- requirements: have `management ip port [pw_file]` in the config 
-- have the automatic service running and  starting the connections in config-auto
(For now, use a numeric v4 IP for management i/f -- better stick to 127.0.0.1. Further, pw_file == stdin not supported).
-- use nested config menu view in the GUI (advanced settings) -- not required, but highly recommended if you have more than a few configs. It will keep persistent connections in a distinct menu hierarchy.

The GUI will not launch or terminate these processes, but will display the status, allow user to disconnect (i.e,, put on management-hold) and connect the tunnel, respond to password dialogs etc. (needs management-query-password for that to work) 

On exiting the GUI, such connections are left at their current state. Behaviour of configs in config_dir and global_config_dir is unchanged. Start the GUI as limited user, as before.

To test, use an executable from GHA build artifacts, have atleast one  config in the config-auto folder with OpenVPNService running.

Notes:
(1) `config-auto` directory location is set by the installer and could be customized in `HKLM\OpenVPN\`
(2) If you want to disallow users from controlling such auto-started connections, just leave out the `management ..` option or read-protect the management password file.


cron2:
Hi,

On Thu, Jul 07, 2022 at 05:40:48PM -0700, Selva Nair wrote:
> This PR implements connecting the GUI to instances started by automatic service or by any other means.

Feature-ACK, this sounds useful, and the accompanying "start openvpn before
logon" is a great future to come :-)

The PR has a number of commits right now - do you want to squash them, or 
shall I go review them as they are?

gert
-- 

selvanair:
> The PR has a number of commits right now - do you want to squash them, or shall I go review them as they are?

Thanks for your interest... As no one has started a review, I can cleanup a bit, squash some fixups and push. Will do this weekend.

selvanair:
@cron2 @lstipakov squashed fixups, combined related commits and re-pushed.

selvanair:
Rebased to the new tip (including the pkcs11 PR) and resolved two minor conflicts.

openvpn.c
```diff
@@ -2177,18 +2160,53 @@ StartOpenVPN(connection_t *c)
...
     {
         ShowLocalizedMsg(IDS_ERR_CREATE_THREAD_STATUS);
-        goto out;
+        return false;
+    }
+
+    /* Launch openvpn.exe using the service or directly */
+    if (!LaunchOpenVPN(c))
+    {
+        TerminateThread(hThread, 1);
```
lstipakov, line 2318: is this new `TerminalThread()` call really needed?

lstipakov (approved):

openvpn.c
lstipakov, line 2318: Is this new `TerminateThread()` call really needed?

lstipakov:
Hm I thought I approved just one commit but looks like here you could approve the whole PR only, unlike in bitbucked. Let me check rest of commits.

openvpn.c
```diff
@@ -2431,10 +2558,18 @@ SuspendOpenVPN(int config)
...
 RestartOpenVPN(connection_t *c)
 {
-    if (c->hwndStatus)
+    if (c->state == onhold)
+    {
+        ReleaseOpenVPN(c);
+    }
+    else if (c->hwndStatus)
+    {
         PostMessage(c->hwndStatus, WM_OVPN_RESTART, 0, 0);
```
lstipakov, line 2592: For the sake of consistency, one might add `RestartOpenVPN()` here.

main.c
```diff
@@ -277,6 +277,7 @@ int WINAPI _tWinMain (HINSTANCE hThisInstance,
   if (use_iservice && strtod(o.ovpn_version, NULL) > 2.3 && !o.silent_connection)
     CheckIServiceStatus(TRUE);
 
+  CheckServiceStatus();	/* Check if automatic service is running or not */
```
lstipakov, line 280: When automatic service is not installed (which is by default), this triggers "Failed to open OpenVPNService" message box.

lstipakov (commented):
Shouldn't we remove `--service-only` from usage in resources?

openvpn.c
selvanair: If we fail here, won't the thread created above hang around in a suspended state and leak memory?

main.c
selvanair: Oh, right, we have to remove that warning -- will fix.
BTW, though this is not the place to discuss it, why don't we install automatic service by default (but not set it to start unless required)? It's a pain to install it later as we do not support modify in the MSI.

options.c
```diff
@@ -75,6 +75,8 @@ ExpandOptions (void)
 {
     ExpandString (o.exe_path, _countof(o.exe_path));
     ExpandString (o.config_dir, _countof(o.config_dir));
+    ExpandString (o.global_config_dir, _countof(o.global_config_dir));
+    ExpandString (o.config_auto_dir, _countof(o.config_auto_dir));
```
selvanair, line 79: Probably its wrong to expand these strings as these are restricted locations and a limited user should not be allowed to override them using env vars. Same with install_path below (not touched by this patch).. Right?

main.c
```diff
@@ -550,6 +582,9 @@ LRESULT CALLBACK WindowProcedure (HWND hwnd, UINT message, WPARAM wParam, LPARAM
         SendMessage(hwnd, WM_CLOSE, 0, 0);
         break;
       }
+      /* A timer to periodically tend to persistent connections */
+      SetTimer(hwnd, 0, 10000, ManagePersistent);
```
lstipakov, line 620: Why do we wait 10 seconds before the first attempt to attach to daemons?

lstipakov: Ah I know why. We don't wait 10 seconds - on startup we start all connections with `auto_connect` flag, this also handles persistent connections. This code gives us a chance to reattach.

lstipakov:
About config parser - do we really need all that complexity only to get something like `management 127.0.0.1 12345`?

main.c
selvanair: Yes, as you figured. But, on further thought, logically it may be better to do even the first start here in ManagePersistent but for now this works.

selvanair:
A simpler approach may be possible, but not much simpler: we still need to handle escapes, single vs double quotes, embedded spaces which may or may not be escaped, and changes to current directory (for password file). Think of 
cd "C:\\My\ S'crets\"
...
management "127.0.0.1" '2500' "pw file"
which cannot be easily tokenized by strtok or such.

Also, we already have a half-baked config parser in passphrase.c and inability to parse the config has been felt as a limitation sometimes. This appears to be a chance to get it done and reuse it in future.

selvanair:
Added a few commits on top: fixups and resources and an improvement. Fixup commits to be squashed later.

lstipakov:
Works here, but one thing I've found - when you manually detach from daemon process, for example by pressing "detach" button on status window, and then right click on tray icon, in several second you will be attached again.

lstipakov:
Another thing - I noticed high CPU usage of gui process and looks like we got a busy loop in `ThreadOpenVPNStatus`:

```
/* Run the message loop for the status window */
    while (WM_QUIT != msg.message)
    {
        DWORD res;
        if (!PeekMessage(&msg, NULL, 0, 0, PM_REMOVE))
        {
            if ((res = MsgWaitForMultipleObjectsEx (1, &wait_event, INFINITE, QS_ALLINPUT,
                                         MWMO_ALERTABLE)) == WAIT_OBJECT_0)
            {
                if (wait_event == c->hProcess)
                    OnProcess (c, NULL);
                else if (wait_event == c->iserv.hEvent)
                    OnService (c, NULL);
            }
            continue;
        }
```
 - `PeekEvent()` returns `FALSE` when there are no messages 
 - `MsgWaitForMultipleObjectsEx()` is called with `-1` (pseudo handle returned by `GetCurrentProcess`) as `wait_event` and returns `WAIT_FAILED` with `ERROR_INVALID_HANDLE`  
 - `continue` is being called and we're back to `PeekMessage()` call.

if we don't expect to read anything from process or service pipe, shouldn't we use blocking `GetMessage()` for persistent connections?

selvanair:
> if we don't expect to read anything from process or service pipe, shouldn't we use blocking `GetMessage()` for persistent connections?

Yes, that would be better.. I was trying to reuse the current code with minimal change, but looks like using the pseudo handle for wait  is not good. Will fix next week. Also the detach/auto-attach issue -- I had seen it an earlier iteration. auto_connect = false on detach should take care of it, but something else could be missing. Will test again.

selvanair:
> Works here, but one thing I've found - when you manually detach from daemon process, for example by pressing "detach" button on status window, and then right click on tray icon, in several second you will be attached again.

I couldn't reproduce this -- is it with the current version of the PR?

lstipakov:
Yeah, let me double-check it. (might go on Monday)

lstipakov:
Also I don't have management-hold in my config - not sure if it affects.

lstipakov:
> I couldn't reproduce this -- is it with the current version of the PR?

Yes. Did you click on tray icon after detach from status window? The click on the tray icon sets `auto_connect` flag to `true`:

```
if (o.enable_persistent == 2)
{
    c->auto_connect = true;
}
```

```
>	openvpn-gui.exe!AddConfigFileToList(int config, const wchar_t * filename, const wchar_t * config_dir) Line 120	C
 	openvpn-gui.exe!BuildFileList0(const wchar_t * config_dir, int recurse_depth, int group, int flags) Line 315	C
 	openvpn-gui.exe!BuildFileList(...) Line 459	C
 	openvpn-gui.exe!RecreatePopupMenus() Line 316	C
 	openvpn-gui.exe!OnNotifyTray(__int64 lParam) Line 331	C
```

Next time `ManagePersistent` timer callback is called, we do attach:

```
if (o.conn[i].flags & FLAG_DAEMON_PERSISTENT
    && o.conn[i].auto_connect
    && (o.conn[i].state == disconnected || o.conn[i].state == detached))
{
    /* disable auto-connect to avoid repeated re-connect
     * after unrecoverable errors. Re-enabled on successful
     * connect.
     */
    o.conn[i].auto_connect = false;
    o.conn[i].state = detached; /* this is required to retain management-hold on re-attach */
    StartOpenVPN(&o.conn[i]); /* attach to the management i/f */
}
```

selvanair:
> > I couldn't reproduce this -- is it with the current version of the PR?
> 
> Yes. Did you click on tray icon after detach from status window? The click on the tray icon sets `auto_connect` flag to `true`:

Yes, I get it now. I always have atleast one connection active in which case the profile list is not recreated from scratch (only new config files, if any, are enumerated). In that case this doesn't happen -- so I missed it. Will fix.

Some more details: profile list is recreated on tray icon click only if all connections are in state == disconnected. Until the last 4 "fixup/improvement" commits I was setting all detached connections as state = detached, which suppresses profile list recreation. Will find a fix that keeps the feature of "preserving managment-hold on auto-connect but not on manual connect" without having this auto-reattach mis-feature.

selvanair:
Fixups for (i) auto re-attach after rescan and (ii) high CPU usage pushed. For (i) I chose an easy way out of not re-creating the profile list when enable_persistent is 2 (auto-attach mode). Newly added config files are still scanned for, so same behaviour as when a connection is active.

The whole re-scanning feature could benefit from some reworking: like rescan only when a new profile is imported etc..  But its beyond the scope of this patch-set.

lstipakov (approved):
Auto-attach and busy-loop issues are fixed. Looks good!

config_parser.c
```diff
@@ -0,0 +1,180 @@
...
+    p = ce->line;
+    s = ce->sline;
+    unsigned int i = 0;
+    int status;
+
+    for ( ; *p != L'\0';  p++, s++)
+    {
+        if (*p == L' ' || *p == L'\t') continue;
+
+        if (_countof(ce->tokens) <= i)
```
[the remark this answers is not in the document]
selvanair: That occurred to me as well. I did not want to ignore comments, but we can use a single token per comment (whole line or trailing part). Will do.
That said, the present code will just truncate the comment content, so no harm done as we are not (yet) using the parsed results to re-write the ovpn file or anything like that. So this could be done as a followup?

openvpn.c
selvanair: Yes, its a bug-fix as well.

selvanair:
Added a commit to not tokenize comments and some fixes for config_parser.c.

selvanair:
Squashed all fixups and force-pushed (no code changes). Ready for merge.

lstipakov:
Tested latest changes (not tokenizing comments etc). LGTM.

lstipakov (approved):

PR #513, opened 2022-07-22
adjust build options to harden binaries

chipitsine:
enable hardware-enforced stack protection on
compatible hardware/software (/CETCOMPAT linker option)

enable QSpectre protection

enable Control Flow Guard

issues were found by Microsoft BinSkim


lstipakov (approved):
Isn't `DYNAMICBASE` set by default?

Looks good, and we do the same thing for openvpn builds.

shipitsin-ilia:
dynamicbase seems to be redundant indeed.

I'll keep an eye on it when planning next PR

PR #524, opened 2022-09-01
Disable persistent state check for OpenVPN3

facboy:
`openvpn3` does not support the `state` management command as a query, it requires either an `on` or an `off` parameter to be supplied.  This commit disables the `state` check when using OpenVPN3.

Signed-off-by: Christopher Ng <facboy@gmail.com>

### DIFF
--- a/openvpn.c
+++ b/openvpn.c
@@ -122,8 +122,11 @@ OnReady(connection_t *c, UNUSED char *msg)
     ManagementCommand(c, "echo on all", OnEcho, combined);
     ManagementCommand(c, "bytecount 5", NULL, regular);
 
-    /* ask for the current state, especially useful when the daemon was prestarted */
-    ManagementCommand(c, "state", OnStateChange, regular);
+    if (o.ovpn_engine != OPENVPN_ENGINE_OVPN3) { // openvpn3 doesn't support this and will respond with an error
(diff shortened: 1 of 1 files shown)


selvanair:
This should be fixed in OpenVPN3, not here.

lstipakov:
OpenVPN3 OMI already reports state as a response to `state on`, so it should be easy to add `state` support ("give me current state"). I could do it soonish.

facboy:
yeah was having a look, i see it's a bug in openvpn3 OMI.

lstipakov:
@facboy I pushed the fix to openvpn3 (https://github.com/lstipakov/openvpn3/actions/runs/2978552284). let me know if that works for you.

facboy:
> @facboy I pushed the fix to openvpn3 (https://github.com/lstipakov/openvpn3/actions/runs/2978552284). let me know if that works for you.

oh i made my own PR lol: https://github.com/OpenVPN/openvpn3/pull/239.  i'll close it.

only thing i would suggest is perhaps use

```
const std::string arg1 = o.get_default(1, 16, "1")
```

instead of 

```
const std::string arg1 = o.size() > 1 ? o.get(1, 16) : "1";
```

i've built it locally with the above change and it works fine now.  i'm sure either will work.

lstipakov:
I agree, changed to `o.get_default()`.

PR #525, opened 2022-09-02
Fix crash on empty state response

lstipakov:
OpenVPN3 doesn't yet support "state"
management command without parameters.

While this has to be fixed on OpenVPN3
side, it doesn't mean that gui could simply crash.

### DIFF
--- a/openvpn.c
+++ b/openvpn.c
@@ -286,6 +286,9 @@ OnStateChange(connection_t *c, char *data)
 {
     char *pos, *state, *message;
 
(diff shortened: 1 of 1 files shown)


lstipakov:
I have pushed to [fix](https://github.com/lstipakov/openvpn3/commit/6fe54a349e082472347d7fa30747f630d6da311f) to OpenVPN3, but regardless gui should not crash.

selvanair (approved):

PR #529, opened 2022-10-18
PLAP: Add an option to register the COM dll

selvanair:
And some small improvements:
 - try to start the service if not running but PLAP is enabled
 - on PLAP screen do not list profiles with no --management option (they just fail)

### DIFF
--- a/misc.c
+++ b/misc.c
@@ -969,3 +969,80 @@ dpi_initialize(options_t *options)
 
     DpiSetScale(options, dpix);
 }
+
+#define PLAP_CLASSID "{4fbb8b67-cf02-4982-a7a8-3dd06a2c2ebd}"
+int
+GetPLAPRegistrationStatus(void)
+{
+    int res = 0;
+    HKEY regkey;
+    wchar_t dllPath[MAX_PATH];
+
+    _sntprintf_0(dllPath, L"%ls%ls", o.install_path, L"bin\\libopenvpn_plap.dll");
+    if (!CheckFileAccess(dllPath, GENERIC_READ)) {
+        res = -1;
+    }
+    else if (RegOpenKeyExW(HKEY_CLASSES_ROOT, L"CLSID\\"PLAP_CLASSID, 0, KEY_READ, &regkey)
+              == ERROR_SUCCESS) {
+        res = 1;
+        RegCloseKey(regkey);
+    }
+
+    return res;
+}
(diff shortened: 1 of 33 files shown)


misc.c
```diff
@@ -969,3 +969,80 @@ dpi_initialize(options_t *options)
...
+{
+    DWORD res = -1;
+    const wchar_t *cmd = L"C:\\windows\\system32\\reg.exe";
+    wchar_t params[MAX_PATH];
+
+    /* Run only if the state has changed */
+    int plap_status = GetPLAPRegistrationStatus();
+    if (plap_status > 0 && (BOOL) plap_status == value)  return 0;
+
+    if (value) {
```
lstipakov, line 1005: There is an assumption here that if plap dll is present, then .reg files are present too. This is probably fine since we expect installer to copy those 3 files together.

lstipakov (approved):
Stared at the code and tested.

lstipakov:
At the second thought, the text "Enable Start Before Logon" is a bit confusing - it could mean either
 - enable/disable automatic service, which implements SBL
 - register PLAP, which enables user to provide VPN credentials before logging in

Since we clearly mean the second option, how about just "Enable Pre-Logon Access Provider"? Those who need this functionality know what does it mean.

misc.c
selvanair: That is the assumption.
Same with checking the registry entries to determine already registered -- not all entries required for COM registration are checked.

selvanair:
Text changed to "Enable Pre-Logon Access Provider".

lstipakov (approved):

lstipakov:
> But, right, this option is for advanced users who should know what it is.

Maybe later we could add a hyperlink [what is that?](https://github.com/OpenVPN/openvpn-gui/tree/master/plap) right after that checkbox, pointing to a `readme.rst` in github repo. which explains this feature for curious-but-not-that-nerdy users.

selvanair:
> > But, right, this option is for advanced users who should know what it is.
> 
> Maybe later we could add a hyperlink [what is that?](https://github.com/OpenVPN/openvpn-gui/tree/master/plap) right after that checkbox, pointing to a `readme.rst` in github repo. which explains this feature for curious-but-not-that-nerdy users.

Sounds like a plan :) There are several options that could benefit from help links.

PR #530, opened 2022-10-18
service: change CheckServiceStatus() return type to VOID

lstipakov:
CheckServiceStatus() return value is never used - the status is set to global options_t struct.

While on it, remove unneccessary "false" argument
and reformat the code.

### DIFF
--- a/main.c
+++ b/main.c
@@ -487,7 +487,7 @@ HandleCopyDataMessage(const COPYDATASTRUCT *copy_data)
 static void CALLBACK
 ManagePersistent(HWND hwnd, UINT UNUSED msg, UINT_PTR id, DWORD UNUSED now)
 {
-    CheckServiceStatus(false);
+    CheckServiceStatus();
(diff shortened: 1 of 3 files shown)


main.c
```diff
@@ -51,6 +51,7 @@
 #include "save_pass.h"
 #include "echo.h"
 #include "as.h"
+#include "service.h"
```
selvanair, line 54: This is already included above (at line 41).
Looks good otherwise.

lstipakov: Indeed. Removed and force-pushed.

selvanair:
Hmm.. I tried to resolve the conflict (after merging 529) but that added a merge commit to your branch, but we prefer rebase and merge. I'm still clueless sometimes as to how github works. Please rebase to master and push again.

lstipakov:
Rebased and pushed.

selvanair (approved):
Thanks.

PR #531, opened 2022-10-30
forget-passwords on disconnect and README edits

selvanair:
- Forget passwords while stopping persistent connections
- Update README
- Document persistent and PLAP in README


lstipakov (approved):
Let's see if it `forget-passwords` breaks openvpn3 integration :)

selvanair:
Its sent only to persistent connections on "disconnect", so OpenVPN3 should not be affected -- in theory ;)

PR #536, opened 2022-11-04
Update copyright year and add OpenVPN version to About

selvanair:


cron2:
The changes look reasonable, so "ACK".  What irritates me a bit is all these msvc warnings on the rc files - is this something new as a consequence of this patch?  or "old warnings"?

selvanair:
Oops, I missed that string length of this text exceeds 256 characters in some rc files. Will fix.

selvanair:
Fixed the second commit which was truncating strings in some cases.

As static texts (LTEXT/RTEXT) in rc files cannot exceed 256 chars (MSVC limit?), and we are already at that limit in some languages, the version string is now added by cutting and insetting the text at run time. No rc files are changed. 

This may look like kludge, but the alternative is to move the whole "About" text to a resource string and render it at run time. LTEXT/RTEXT are limited to 256 characters while resource strings can hold more (up to 4096 characters?). We have been working around this by splitting the "About" text into shorter segments but that doesn't work with long sentences that exceeds the limit.

cron2:
Hi,

On Sat, Nov 05, 2022 at 01:20:52PM -0700, Selva Nair wrote:
> Fixed the second commit which was truncating strings in some cases.
> 
> As static texts (LTEXT/RTEXT) in rc files cannot exceed 256 chars (MSVC limit?), and we are already at that limit in some languages, the version string is now added by cutting and insetting the text at run time. No rc files are changed. 

Thanks for the explanation.  With the way these warnings are rendered
in the GH view I did not see that the string continues "under the warnings",
so it looked "short".

> This may look like kludge, but the alternative is to move the whole "About" text to a resource string and render it at run time. LTEXT/RTEXT are limited to 256 characters while resource strings can hold more (up to 4096 characters?). We have been working around this by splitting the "About" text into shorter segments but that doesn't work with long sentences that exceeds the limit.

Looks good to me.  So, ACK.

gert

-- 

PR #540, opened 2022-11-10
Bump version to 11.31.0.0

flichtenheld:
Signed-off-by: Frank Lichtenheld <frank@lichtenheld.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -20,7 +20,7 @@ dnl  59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
 
 AC_PREREQ(2.59)
 define([_GUI_VERSION_MAJOR], [11])
-define([_GUI_VERSION_MINOR], [30])
+define([_GUI_VERSION_MINOR], [31])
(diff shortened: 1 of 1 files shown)


selvanair (approved):

selvanair:
Use "Rebase and Merge" (or squash and merge, if appropriate) instead of "Create merge commit". It keeps the history cleaner, especially for version bumps.

flichtenheld:
> @flichtenheld Use "Rebase and Merge" (or squash and merge, if appropriate) instead of "Create merge commit". It keeps the history cleaner, especially for version bumps.

That destroys the tags. For the pure version-increase commit I will just not do a PR in the future, that way I can do a real Fast-Forward which doesn't seem to be allowed by GH. Just need to remember to coordinate the CHANGES.rst update. In case you would like to take over the version increase and tagging (like @cron2 does for OpenVPN/openvpn), then let me know. We can certainly change the process.

flichtenheld:
Sure, there is no law that we need to link to the CHANGES.rst of the actual tag. Still, will try to remember to give you a warning about a planned release build so you have at least a chance to review it.

PR #541, opened 2022-11-11

justwho:


lstipakov (approved):
Looks good according to my Chinese-speaking colleague,

PR #546, opened 2022-11-27
WIP: Use RTL layout for Farsi

selvanair:
This requires feedback from Persian language users


lstipakov:
Not too much into Farsi, but looks indeed RTL.

Näyttökuva 2022-11-28 100000

lstipakov (approved):
Let's wait for Farsi speaker for confirmation.

selvanair:
Added a commit to fix some bidirectional text in "About"
Before: 
about-b

Incorrect character ordering of trailing ">" and "/"  leads to "<Copyright ..." and "/https://..."

After:
about-a

The fix uses explicit LRE markers. If there is a better way, please suggest.

selvanair:
"WIP" removed from commit message and force-pushed. No code changes.

PR #547, opened 2022-11-29
Make popup message boxes RTL language aware

selvanair:
This also bumps lowest supported Windows version to Windows 7 in mingw builds. I'm not sure what we target in MSVC.

### DIFF
--- a/echo.c
+++ b/echo.c
@@ -404,6 +404,13 @@ static void
 AddMessageBoxText(HWND hwnd, const wchar_t *text, const wchar_t *title, const wchar_t *from)
 {
     HWND hmsg = GetDlgItem(hwnd, ID_TXT_MESSAGE);
+    DWORD align[2] = {PFA_LEFT, PFA_RIGHT}; /* default alignments for text and title */
+
+    if (LangFlowDirection() == 1)
+    {
+        align[0] = PFA_RIGHT;
+        align[1] = PFA_LEFT;
+    }
 
     /* Start adding new message at the top */
     SendMessage(hmsg, EM_SETSEL, 0, 0);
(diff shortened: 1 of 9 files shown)


localization.c
```diff
@@ -69,6 +69,22 @@ FindResourceLang(PTSTR resType, PTSTR resId, LANGID langId)
     return FindResource(o.hInstance, resId, resType);
 }
 
+/*
+ * Return value: 0 for LTR, 1 for RTL, 2 or 3 for vertical
+ */
+int
+LangFlowDirection(void)
```
lstipakov, line 76: Just a nit-pick. It looks like that this function is always used to get the value of `rtl` flag:

    DWORD rtl = (LangFlowDirection() == 1) ? MB_RIGHT|MB_RTLREADING : 0;

Wouldn't it be better to rename it to `GetRTLFlags()` and make it return `MB_RIGHT|MB_RTLREADING` or `0` ?

lstipakov (approved):

localization.c
selvanair: Sure, for use in message boxes that would  be expedient. But we may need flow direction to handle RTL in other Windows like the PLAP task dialog or echo-message. I'll check those and simplify this if there is no other immediate need for a general function like this.

selvanair:
Changes:
- Defined MBOX_RTL_FLAGS macro instead of repeated `rtl = (LangFlowDirection() ==1 ) ? MB_RIGHT | MB_RTLREADING : 0`
- Retained LangFlowDirection() for use in next commit
- Added commits to handle RTL in echo-msg window and PLAP task dialog

echo.c
```diff
@@ -404,6 +404,13 @@ static void
 AddMessageBoxText(HWND hwnd, const wchar_t *text, const wchar_t *title, const wchar_t *from)
 {
     HWND hmsg = GetDlgItem(hwnd, ID_TXT_MESSAGE);
+    DWORD align[2] = {PFA_LEFT, PFA_RIGHT}; /* default alignments for text and title */
+
+    if (LangFlowDirection() == 1)
```
lstipakov, line 409: I wonder why Windows hasn't provided macros for reading layout [values](https://learn.microsoft.com/en-us/windows/win32/intl/locale-ireadinglayout).

lstipakov (approved):

PR #560, opened 2022-12-12

justwho:


lstipakov (approved):
Consulted a colleague, native Chinese speaker, looks good!

Issue #260, opened 2018-05-03
Suggestion: Add Autostart "this" config at startup


TinCanTech:
On the General Tab:
Startup:
[X] Launch at startup
[Drop down box] Start this profile

This does not have to be a drop down box, it could be a manually filled in field etc..

TinCanTech:
@Crypt-ix .. you can connect to multiple servers at the same time using the GUI.

While I understand how to do this with shortcut properties, my idea is for making that step even easier by building that functionality into the GUI itself.

Anyway, it is only a suggestion not a bug per se.

yegors:
Just install the Windscribe client, it has custom config support and the feature you seek: https://cdn-images-1.medium.com/max/1200/1*HXJzS1RJmM3foewUpEx12Q.png

cron2:
Hi

On Tue, Jun 18, 2019 at 08:35:48AM -0700, Yegor S wrote:
> Just install the Windscribe client, it has custom config support and the feature you seek: https://cdn-images-1.medium.com/max/1200/1*HXJzS1RJmM3foewUpEx12Q.png

Please stop spamming OpenVPN issues with advertisement for Windscribe.

There's tons of nice solutions out there, some free and some commercial,
but we do not need advertisement for any of them in an OpenVPN issue.

gert

-- 
"If was one thing all people took for granted, was conviction that if you 
 feed honest figures into a computer, honest figures come out. Never doubted 
 it myself till I met a computer with a sense of humor."
                             Robert A. Heinlein, The Moon is a Harsh Mistress

Gert Doering - Munich, Germany                             gert@greenie.muc.de

PR #565, opened 2022-12-20
Active connections on exit/logout are auto-started in the next session

### LINKED ISSUE #215: 1Tap Connect
I suggest add this feature:
Connect to last server by one click on system tray icon

selvanair:
Closes #215 #260

To test: start one or more connections then exit the GUI without explicitly disconnecting these connection(s). As before, the GUI will warn that active connections will be closed -- click OK. Start the GUI again and it will auto-start the connections that were active at exit. If autostart is not desired, just stop the connection before exit. 

Same behaviour if user logs out with active connections. On re-login previously active connection(s) will get restarted. 

The feature may be disabled in the settings menu.


main.c
```diff
@@ -949,3 +959,126 @@ ErrorExit(int exit_code, const wchar_t *msg)
...
+
+    for (int i = 0; i < o.num_configs; i++)
+    {
+        if (o.conn[i].state == disconnected
+            || o.conn[i].flags & FLAG_DAEMON_PERSISTENT)
+        {
+            continue;
+        }
+        /* accumulate space needed for list of active connections */
+        if (active_conns)
```
lstipakov, line 1000: We already check this condition above, on line 986, and return if it is false.

main.c
```diff
@@ -949,3 +959,126 @@ ErrorExit(int exit_code, const wchar_t *msg)
...
+        free(active_conns);
+        MsgToEventLog(EVENTLOG_ERROR_TYPE, L"Out of memory while persisting state in registry");
+        return;
+    }
+
+    wchar_t *p = list;
+    for (int i = 0; i < nactive; i++)
+    {
+        connection_t *c = &o.conn[active_conns[i]];
+        if (c->state == disconnected || c->flags & FLAG_DAEMON_PERSISTENT)
```
lstipakov, line 1023: We exclude disconnected and persistent connections in the loop above. Why do we need to check it here again?

main.c
```diff
@@ -949,3 +959,126 @@ ErrorExit(int exit_code, const wchar_t *msg)
...
+    DWORD len = 0;
+
+    LSTATUS status = RegGetValueW(HKEY_CURRENT_USER, GUI_REGKEY_HKCU, L"auto_restart_list",
+                                RRF_RT_REG_MULTI_SZ, NULL, NULL, &len);
+    if (status != ERROR_SUCCESS || len == 0)
+    {
+        return;
+    }
+
+    list = malloc(sizeof(len));
```
lstipakov, line 1059: Shouldn't it be `malloc(len)` ?

selvanair: That should be sizeof(wchar_t)*len
I did test this, couldn't have worked with such silly bugs. Let me check what I pushed.

lstipakov: I managed to break it to heap corruption by putting long enough string into registry value. With short strings it just "works" - although I expected VS compiler to be smarter and detect those things.

main.c
selvanair: Right, no need to check persistent flag again. As for the state, it could have changed in the mean time in case the connection was in the process of disconnecting. But, on further thought, the second check is only marginally useful and its not critical to get this absolutely right. Will remove.

lstipakov:
Looks good. I quickly tested it and it works as expected.

lstipakov (approved):

Issue #567, opened 2022-12-21
Silent failure of PLAP on adding management password


selvanair:
@selvanair I've been testing the PLAP without management interface password and all worked fine.
Now that I would like to test it on our corporate environment, I have added a password file to the management interface like this:

`management 127.0.0.1 8888 passwd.txt`

The passwd.txt file has one line with the management interface password. I've copied the file to the same path as the ovpn (C:\Program Files\OpenVPN\config-auto), but after this, I can see the PLAP on Windows Logon screen and when I choose the ovpn profile it doesn't ask any username/password (or any OpenVPN window) and immediately shows that the connection failed.
If I try to start the connection with the OpenVPN GUI then I see a quick window opening and closing again. Log file does not log anything.
If I remove the passwd.txt parameter from the management configuration, it works like expected.
Am I missing something or doing something wrong?
Thanks again for all you work for making this PLAP a reality! :)

_Originally posted by @gitnewcomer in https://github.com/OpenVPN/openvpn-gui/issues/77#issuecomment-1361414392_

selvanair:
Moved to a new issue --- #77 was about a feature request that has been completed and closed. Better to open new issues for reporting errors like this or to request improvements.

@gitnewcomer 

Based on what you describe, my guess would be: you added the password, restarted the service, then edited the password file and did not restart the service again. That would lead to a management password mismatch and the only error would be in the event log. We should probably show an error popup for this ---  iirc, there was some reason not to, but I'll reconsider. At least we've to return a relevant reason for failure to PLAP screen.

> Log file does not log anything.

OpenVPN daemon for persistent connections is started by OpenVPNService and it must log something to the log file --- even if the startup failed for some reason. I guess you mean that the log file does not show anything related to the failure of PLAP or GUI. That would be the case if the connection to management interface failed due to management password mismatch -- in that case GUI/PLAP logs a warning to Windows EventLog.

Restart OpenVPNService and check the log file in `C:\Program Files\OpenVPN\log\<config-name>.log` to ensure the process has started up. Then try GUI and PLAP again. You can also directly test the management password by connecting to management interface using telnet (exit the GUI before trying this): `telnet 127.0.0.1 8888` it will prompt for management password. If that works, quit telnet before starting the GUI or PLAP again.

BTW, unlike user configs, you have to always restart OpenVPNService after editing a config file in `config-auto` folder for changes to take effect. This is true for all persistent connections (those started at boot up) whether PLAP is in use or not.

[...]

PR #568, opened 2022-12-22
Increase max size of management password buffer

selvanair:
As we now allow users to set a management password (for persistent connections), the max size of password should match what openvpn.exe can handle (128 or 4096 bytes depending on build options).

Increase the buffer size to 4096 though such large passwords may not work in practice. 127 bytes + NUL, may be a safe upper limit.

For the random password used for connections spawned by the GUI, the current size of 15 bytes + NUL is retained.

Fixes: #567

### DIFF
--- a/openvpn.c
+++ b/openvpn.c
@@ -2390,6 +2390,12 @@ LaunchOpenVPN(connection_t *c)
     HANDLE hNul = NULL;
     DWORD written;
     BOOL retval = FALSE;
+    DWORD passwd_len = 16; /* incuding NUL */
+
+    if (passwd_len > sizeof(c->manage.password))
+    {
+        passwd_len = sizeof(c->manage.password);
+    }
 
     RunPreconnectScript(c);
 
@@ -2403,7 +2409,7 @@ LaunchOpenVPN(connection_t *c)
     }
 
     /* Create a management interface password */
-    GetRandomPassword(c->manage.password, sizeof(c->manage.password) - 1);
+    GetRandomPassword(c->manage.password, passwd_len - 1);
 
     find_free_tcp_port(&c->manage.skaddr);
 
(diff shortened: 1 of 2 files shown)


lstipakov (approved):

gitnewcomer:
@selvanair how can I test the fix? I have tried the files from artifacts `(openvpn-gui_x64_ossl3.zip)` and got an error that the function `OPENSSL_strcasecmp` was not found on `libssl-3-x64.dll`. Prior I have installed OpenVPN 2.6 Beta2.
Is there a new MSI with the fix I could test?
Thanks!

lstipakov:
Looks like artifacts use OpenSSL 3.0.2 and 2.6beta uses 3.0.7. Let me fix the artifacts building,

lstipakov:
@gitnewcomer Could you try artifacts from https://github.com/lstipakov/openvpn-gui/actions/runs/3764681539 ? This is based on master + openssl version bump.

gitnewcomer:
> @gitnewcomer Could you try artifacts from https://github.com/lstipakov/openvpn-gui/actions/runs/3764681539 ? This is based on master + openssl version bump.

Thanks @lstipakov, it works now! :)
Do you know when would be released a new MSI package with this fix?
Just another quick question: when we choose to install the PLAP on the installer, wouldn't it be better to activate the automatic startup of the OpenVPNService? This would allow a quicker deploy on many clients, because after copying the ovpn profile and password file to the `config-auto`, the PLAP would activate automatically on reboot.
Thanks again for all the work you all have done! I wish you all a Merry Christmas! :)

lstipakov:
We should have beta3 / rc1 before the end of year. 

Yes I agree, it makes sense to auto-start the service for clean install, should be fixed in https://github.com/OpenVPN/openvpn-build/pull/308.

selvanair:
I suggest to change the way we activate the service install feature. Always install the service and plap binaries, but leave the service feature in the installer unselected by default. Then enabling the service feature during install would mean that service is started (and set to auto-start) even on fresh install, choosing PLAP activation would imply service start.

PR #578, opened 2023-01-11
Handle CONNECTED,ROUTE_ERROR from openvpn.exe

selvanair:
OpenVPN 2.6 now reports `CONNECTED,ROUTE_ERROR` when connection completed. Currently in the GUI we will treat this as an error and may be seen as a regression by some users.

Until we decide on how to treat route errors in the GUI, this PR is a quick fix to handle it almost the same way as the _status quo_. What is added is some extra error logging and automatic popup of the status window to draw attention to the errors.

**Change in behaviour when using 2.6 master:**
If openvpn.exe reports CONNECTED,SUCCESS or CONNECTED,ERROR: no changes
If openvpn.exe reports CONNECTED,ROUTE_ERROR : 
   - an error is written to the status log (in red)
   - the status window is displayed (if not already open)
   - the status window icon stays yellow
   - the status text says "Connected with route errors"
   - the tray notification indicates "route errors".
   Apart from that the connection works same as before, the tray icon turns green.

As this is an interim fix, new resource strings are defined only in the fall-back language (English). Both the error/info messages and the tray icon state could be improved based how fatal route errors are in real use.

To test, use openvpn.exe from 2.6 master and force a route error --- either use a bogus route like `--route 192.168.10.0 255.255.255.0 1.1.1.1` with `dco` or stop `interactive service` and set driver to `tap-windows6` (without the service running, `dco` will error out early at address setting stage).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,11 @@
+Version 11.35.0
+===============
+
+Updates
+-------
+* Translations (Italian, Chinese-Simplified)
+* Notify user if connection completes with route addition errors
+
 Version 11.34.0
 ===============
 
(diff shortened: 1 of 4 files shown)


lstipakov (approved):
Looks good. Tested and it works as expected.

PR #583, opened 2023-01-19
Bug-fix: Connection list

selvanair:
Currently we use an array of connection pointers which needs
to be reallocated when space runs out. But, that happens from
the main thread while the status thread may be referring to those
pointers. This can cause invalid memory access in some situations:
e.g., a connection is active and user adds tens of configs 
requiring a re-alloc during next re-scan.

Use a list instead of an array  so that connection pointer never
changes once created.

### DIFF
--- a/main.c
+++ b/main.c
@@ -358,17 +358,17 @@ StopAllOpenVPN()
      * at their current state. Use the disconnect menu to put them into
      * hold state before exit, if desired.
      */
-    for (i = 0; i < o.num_configs; i++)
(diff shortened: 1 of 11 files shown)


lstipakov:
We have `SetMenuStatusById` leftovers, let's remove those:

**stub.c**
```
void SetMenuStatusById(UNUSED int id, UNUSED conn_state_t state)
{
    return;
}
```

**tray.h**
`void SetMenuStatusById(int, conn_state_t);`

PR #584, opened 2023-01-19
Bugfix AutoCloseHandler(): accommodate wraparound of 32 bit time

selvanair:
- GetTickcount() and current-time passed-in to the callback are both 32 bit which wraps around every ~50 days. In the stop condition compare the elasped ticks and timeout value instead of end-time and now. The latter can wraparound in a long-running process.

### DIFF
--- a/openvpn.c
+++ b/openvpn.c
@@ -453,15 +453,15 @@ AutoCloseHandler(HWND hwnd, UINT UNUSED msg, UINT_PTR id, DWORD now)
     if (!ac)
         return;
 
-    UINT end = ac->start + ac->timeout;
-    if (now >= end)
+    DWORD elapsed = now - ac->start;
+    if (elapsed >= ac->timeout)
     {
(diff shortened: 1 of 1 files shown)


lstipakov:
Any reasons not to use `GetTickCount64` ? This is what VS recommends:

`GetTickCount overflows roughly every 49 days.  Code that does not take that into account can loop indefinitely.  GetTickCount64 operates on 64 bit values and does not have that problem.`

`GetTickCount64` is supported starting from Vista, so we should be good.

selvanair:
Sure, we can use GetTickCount64 and ignore the "now" passed-in through the Timer callback which is always 32bit. But 32 bit works fine here as the "timeout" itself is never as long as even a minute, let alone 49 days. Only the "start" and "now" wraps, not their difference.

So, with this change, wraparound should not affect us, but if you "thumb-down" this, I'll rewrite :)

lstipakov (approved):

Issue #576, opened 2023-01-08
OpenVPN daemon left running on GUI exit if a modal dialog is pending


selvanair:
Openvpn.exe started by the GUI fails to terminate if the user tries to exit while a modal dialog is open. The GUI exits after timing out waiting for openvpn to terminate.

**To reproduce**
Start a connection using the GUI with `--auth-user-pass` in the config file. When the user-pass dialog pops up, move it aside, right click on the tray icon and press `exit`. A message box will popup, click `yes` to proceed with the exit process. The GUI will appear to stall for about 5 seconds and then exit. Open taskmgr and to see that "openvpn.exe" is left running.

Sometimes one has to try this a couple of times to reproduce, indicating some timing issue that causes the WM_OVPN_STOP message to get lost.

Happens only if the modal dialog is running on the UI thread (like user-pass or some other config-specific dialog). Dailogs like "import from url" that runs in the main thread does not cause this.

PR #587, opened 2023-01-20
Misc fixes

selvanair:
This completes all "bugs" I found while looking for what causes #577. None of these would explain the crash in #577 though. I also have a fix for #576, but it requires more testing. For lack of time, leaving it for 2.6.1.

### DIFF
--- a/pkcs11.c
+++ b/pkcs11.c
@@ -38,6 +38,7 @@
 #include <assert.h>
 
 extern options_t o;
+static const wchar_t *hfontProp;
 
 /* state of list array */
 #define STATE_GET_COUNT 1
@@ -417,7 +418,7 @@ pkcs11_listview_init(HWND parent)
         if (hfb)
         {
             SendMessage(ListView_GetHeader(lv), WM_SETFONT, (WPARAM)hfb, 1);
-            SetProp(parent, L"header_font", (HANDLE)hfb);
+            SetProp(parent, hfontProp, (HANDLE)hfb);
         }
     }
 
@@ -658,12 +659,12 @@ QueryPkcs11DialogProc(HWND hwndDlg, UINT msg, WPARAM wParam, LPARAM lParam)
 
         case WM_NCDESTROY:
             RemoveProp(hwndDlg, cfgProp);
-            HFONT hf = (HFONT) GetProp(hwndDlg, L"header_font");
+            HFONT hf = (HFONT) GetProp(hwndDlg, hfontProp);
             if (hf)
             {
                 DeleteObject(hf);
             }
-            RemoveProp(hwndDlg, cfgProp);
+            RemoveProp(hwndDlg, hfontProp);
             break;
     }
     return FALSE;
--- a/pkcs11.c
+++ b/pkcs11.c
@@ -35,7 +35,6 @@
 #include <commctrl.h>
 #include <cryptuiapi.h>
 #include <shlwapi.h>
-#include <assert.h>
 
 extern options_t o;
 static const wchar_t *hfontProp;
--- a/main.c
+++ b/main.c
@@ -79,6 +79,9 @@ options_t o;
 /* Workaround for ASLR on Windows */
 __declspec(dllexport) char aslr_workaround;
 
+/* globals */
+static HWND settings_window;   /* Handle of Settings window */
+
 static int
 VerifyAutoConnections()
 {
@@ -758,12 +761,31 @@ AboutDialogFunc(UNUSED HWND hDlg, UINT msg, UNUSED WPARAM wParam, LPARAM lParam)
   return FALSE;
 }
 
+static int CALLBACK
+SettingsPsCallback(HWND hwnd, UINT msg, UNUSED LPARAM lParam)
+{
+    switch(msg)
+    {
+        case PSCB_INITIALIZED:
+            settings_window = hwnd;
+            break;
+    }
+
+    return 0;
+}
 
 static void
 ShowSettingsDialog()
 {
-  PROPSHEETPAGE psp[4];
-  int page_number = 0;
+    PROPSHEETPAGE psp[4];
+    int page_number = 0;
+
+    if (settings_window && IsWindow(settings_window))
+    {
+        SetForegroundWindow(settings_window);
+        ShowWindow(settings_window, SW_SHOW);
+        return;
+    }
 
   /* General tab */
   psp[page_number].dwSize = sizeof(PROPSHEETPAGE);
@@ -807,7 +829,7 @@ ShowSettingsDialog()
 
   PROPSHEETHEADER psh;
   psh.dwSize = sizeof(PROPSHEETHEADER);
-  psh.dwFlags = PSH_USEHICON | PSH_PROPSHEETPAGE | PSH_NOAPPLYNOW | PSH_NOCONTEXTHELP;
+  psh.dwFlags = PSH_USEHICON | PSH_PROPSHEETPAGE | PSH_NOAPPLYNOW | PSH_NOCONTEXTHELP | PSH_USECALLBACK;
   psh.hwndParent = o.hWnd;
   psh.hInstance = o.hInstance;
   psh.hIcon = LoadLocalizedIcon(ID_ICO_APP);
@@ -815,7 +837,7 @@ ShowSettingsDialog()
   psh.nPages = page_number;
   psh.nStartPage = 0;
   psh.ppsp = (LPCPROPSHEETPAGE) &psp;
-  psh.pfnCallback = NULL;
+  psh.pfnCallback = SettingsPsCallback;
 
   PropertySheet(&psh);
 }
--- a/openvpn_config.c
+++ b/openvpn_config.c
@@ -26,7 +26,6 @@
 #endif
 
 #include <windows.h>
-#include <assert.h>
 
 #include "main.h"
 #include "openvpn-gui-res.h"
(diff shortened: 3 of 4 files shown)


selvanair:
strawberry-perl is a real pita.. And, all these repeated openssl builds are anyway a huge waste of time and resources. All for a "password-change" feature of limited use that works only for private key files on disk, not inline.

Shall we get rid of it, or replace with a script that would depend on openssl.exe? We could install the latter by default.

shipitsin-ilia:
changing password for on disk keys is considered as somewhat useful, however, I'm not sure that it belongs to the scope of vpn software.

also, I found that inlined key (embedded into config file) cannot be enctypted that way.

I vote for getting rid of it completely, there are many tools for encrypting private keys anyway.

cron2:
Hi,

On Thu, Jan 19, 2023 at 09:37:00PM -0800, Selva Nair wrote:
> strawberry-perl is a real pita.. And, all these repeated openssl builds are anyway a huge waste of time and resources. All for a "password-change" feature of limited use that works only for private key files on disk, not inline.
> 
> Shall we get rid of it, or replace with a script that would depend on openssl.exe? We could install the latter by default.

If that is really the only use ("change private key pass phrase for a
non-embedded .pem file on disk") I say get rid of it, and replace with
a help popup that shows a sample command line.

Lots of complication for a rarely-used feature.

Having a good mechanism to such keys into windows key store and then 
just throw away the original file would be much more secure anyway...
(we do not have this, right?)

gert

-- 

chipitsine:
пт, 20 янв. 2023 г. в 13:53, Gert Doering ***@***.***>:

> Hi,
>
> On Thu, Jan 19, 2023 at 09:37:00PM -0800, Selva Nair wrote:
> > strawberry-perl is a real pita.. And, all these repeated openssl builds
> are anyway a huge waste of time and resources. All for a "password-change"
> feature of limited use that works only for private key files on disk, not
> inline.
> >
> > Shall we get rid of it, or replace with a script that would depend on
> openssl.exe? We could install the latter by default.
>
> If that is really the only use ("change private key pass phrase for a
> non-embedded .pem file on disk") I say get rid of it, and replace with
> a help popup that shows a sample command line.
>
> Lots of complication for a rarely-used feature.
>
> Having a good mechanism to such keys into windows key store and then
> just throw away the original file would be much more secure anyway...
> (we do not have this, right?)
>

those private keys are usually generated on openvpn server side and
distributed to user as config/embedded or zip archive with several files.

I'm afraid that "generate private key in windows key store and let openvpn
server to sign that key"  is never used in real world.

cron2:
Hi,

On Fri, Jan 20, 2023 at 12:10:35AM -0800, Ilya Shipitsin wrote:
> > Having a good mechanism to such keys into windows key store and then
> > just throw away the original file would be much more secure anyway...
> > (we do not have this, right?)
> 
> those private keys are usually generated on openvpn server side and
> distributed to user as config/embedded or zip archive with several files.
> 
> I'm afraid that "generate private key in windows key store and let openvpn
> server to sign that key"  is never used in real world.

The sentence above was missing an "... to *import* such keys...", so 
no matter where the keys come from, if we can import them into a secure
key storage and then remove the file, it would be nice.

gert
-- 

pkcs11.c
```diff
@@ -663,7 +663,7 @@ QueryPkcs11DialogProc(HWND hwndDlg, UINT msg, WPARAM wParam, LPARAM lParam)
             {
                 DeleteObject(hf);
             }
-            RemoveProp(hwndDlg, cfgProp);
+            RemoveProp(hwndDlg, L"header_font");
```
lstipakov, line 665: For `cfgProp` we use `const TCHAR *cfgProp = _T("conn");`, shouldn't we do the same for `header_font`?

selvanair:
Hi

On Fri, Jan 20, 2023 at 3:21 AM Gert Doering ***@***.***>
wrote:

> Hi,
>
> On Fri, Jan 20, 2023 at 12:10:35AM -0800, Ilya Shipitsin wrote:
> > > Having a good mechanism to such keys into windows key store and then
> > > just throw away the original file would be much more secure anyway...
> > > (we do not have this, right?)
> >
> > those private keys are usually generated on openvpn server side and
> > distributed to user as config/embedded or zip archive with several files.
> >
> > I'm afraid that "generate private key in windows key store and let
> openvpn
> > server to sign that key" is never used in real world.
>
> The sentence above was missing an "... to *import* such keys...", so
> no matter where the keys come from, if we can import them into a secure
> key storage and then remove the file, it would be nice.
>

"Import" and then use cryptoapicert is what I do with almost all Windows
users I deal with. Very easy with pkcs12 files -- just double click and an
import dialog opens up. PEM certificate files can also be imported
similarly, but AFAIK, there is no easy way to import into Windows PEM key
files that OpenSSL generates (encrypted or not).

Selva

lstipakov (approved):
Multiple settings dialog is something I was also annoyed by. Slightly tested, looks good.

chipitsine:
пт, 20 янв. 2023 г. в 14:21, Gert Doering ***@***.***>:

> Hi,
>
> On Fri, Jan 20, 2023 at 12:10:35AM -0800, Ilya Shipitsin wrote:
> > > Having a good mechanism to such keys into windows key store and then
> > > just throw away the original file would be much more secure anyway...
> > > (we do not have this, right?)
> >
> > those private keys are usually generated on openvpn server side and
> > distributed to user as config/embedded or zip archive with several files.
> >
> > I'm afraid that "generate private key in windows key store and let
> openvpn
> > server to sign that key" is never used in real world.
>
> The sentence above was missing an "... to *import* such keys...", so
> no matter where the keys come from, if we can import them into a secure
> key storage and then remove the file, it would be nice.
>

the sentence is not missing anything ))

yes, an average user can import p12 and import openvpn config, but it is
near to impossible to teach user to adopt
openvpn config to use imported file

however, such a feature can be really nice

cron2:
Hi,

On Fri, Jan 20, 2023 at 01:09:25AM -0800, Selva Nair wrote:
> "Import" and then use cryptoapicert is what I do with almost all Windows
> users I deal with. Very easy with pkcs12 files -- just double click and an
> import dialog opens up. PEM certificate files can also be imported
> similarly, but AFAIK, there is no easy way to import into Windows PEM key
> files that OpenSSL generates (encrypted or not).

This is somewhat outside the realms of *this* ticket... but, this is 
good to know.  So I might change to "tell my windows users to just import
the pkcs12 files they get from the inhouse CA anyway, and have OpenVPN
reference it" (--cryptoapicert "SUBJ:$myname").

Is there a good trick to get something generic into cryptoapicert,
like the CA name ("when connecting to $customer, use any non-expired
cert in the CA store, but only those from $customerca") so users do
not have to modify the .ovpn file?

gert
-- 

selvanair:
> near to impossible to teach user to adopt
> openvpn config to use imported file

What I do is to distribute configs with `cryptoapi-cert THUMB:<thumbprint>` already filled in,  and a p12 file that the user has to just double click and import. The thumbprint is known as soon as the certificate is generated. For really clueless users I provide a script that does this.

That said, improving our import feature to handle multiple files (dependencies) and do this on the fly is a nice project for those anxious to contribute :)

pkcs11.c
selvanair: We can but its only used once so far...

selvanair:
> Is there a good trick to get something generic into cryptoapicert,
> like the CA name ("when connecting to $customer, use any non-expired
> cert in the CA store, but only those from $customerca") so users do
> not have to modify the .ovpn file?

Not currently. I find SUBJ: error-prone/ambiguous, so use thumb, but that requires per-user config generation. There was a patch to add a selector based on "templates" but that's a Windows CA thing and not much use for most of us. 

I think we can easily add a CA-based selection though its also as ambiguous (or even more) as SUBJ:. But could work for some.

chipitsine:
пт, 20 янв. 2023 г. в 15:20, Selva Nair ***@***.***>:

> near to impossible to teach user to adopt
> openvpn config to use imported file
>
> What I do is to distribute configs with cryptoapi-cert THUMB:<thumbprint>
> already filled in, and a p12 file that the user has to just double click
> and import. The thumbprint is known as soon as the certificate is
> generated. For really clueless users I provide a script that does this.
>

personally, I am interested in using computer self issued certificates for
PLAP.
but haven't found spare time yet ((

pkcs11.c
lstipakov:
```
SetProp(parent, L"header_font", (HANDLE)hfb);	C:\Users\lev\Projects\openvpn-build\src\openvpn-gui\pkcs11.c	419	31
HFONT hf = (HFONT) GetProp(hwndDlg, L"header_font");	C:\Users\lev\Projects\openvpn-build\src\openvpn-gui\pkcs11.c	660	51
RemoveProp(hwndDlg, L"header_font");	C:\Users\lev\Projects\openvpn-build\src\openvpn-gui\pkcs11.c	665	35
```

selvanair: I know, "once" was used in plural :) (or all in one dialog)...  Well, I'm convinced: will define a variable.

lstipakov:
> strawberry-perl is a real pita.. And, all these repeated openssl builds are anyway a huge waste of time and resources. 

I very much agree.  In openvpn-build (which is triggered on every commit to openvpn) we build openssl twice (because of vcpkg triplet thing) - for gui and for openvpn. This makes everything slow.

PR #589, opened 2023-01-20
Suppress certain warnings

lstipakov:
We have tons of warnings like

  warning C4244: '=': conversion from 'DWORD' to 'LANGID', possible loss of data
  warning C4267: 'initializing': conversion from 'size_t' to 'int', possible loss of data
  warning C4996: 'wcsncpy': This function or variable may be unsafe

We should fix those at some point, but meanwhile let's suppress those since they pollute build logs and make it very hard to analyse.


selvanair (approved):
Build logs are now clean with all those pesky warnings suppressed  -- well, we'll fix them one day. So much easier to see errors and warnings that matter.

selvanair:
test_plap.exe build still emits C4996 --- probably it requires set(CMAKE_CXX_FLAGS "${CMAKE_C_FLAGS}") ? Or we can start by fixing plap_common.c to not use the so called "unsafe" functions.

PR #591, opened 2023-01-22
Check return value of SetProp()

selvanair:
- If SetProp() is unsuccessful, we'll crash later when GetProp() returns null. Add a check, log the error and close the dialog.

  A check on GetProp() is also added, though not strictly necessary.

  We could abort here, but closing the current dialog and possibly the corresponding connection, provides a chance for the user to fix the OOM condition which is the most likely cause of SetProp() failure.

  Github: Fixes OpenVPN/openvpn-gui#577

### DIFF
--- a/as.c
+++ b/as.c
@@ -230,7 +230,7 @@ CRDialogFunc(HWND hwndDlg, UINT msg, WPARAM wParam, LPARAM lParam)
     switch (msg) {
         case WM_INITDIALOG:
             param = (auth_param_t*)lParam;
-            SetProp(hwndDlg, cfgProp, (HANDLE)param);
+            TRY_SETPROP(hwndDlg, cfgProp, (HANDLE)param);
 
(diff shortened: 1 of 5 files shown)


lstipakov:
That `CHECK_NULL_PARAM` thing hasn't made code easier to read, hasn't it? 

It should not be the case for `GetProp()` to return `NULL` if `SetProp()` succeeded. Cannot we just remove `CHECK_NULL_PARAM`? I don't this that its benefits justify costs.

selvanair:
It came out of some debug runs as I was initially concerned with some out of order calls due to race conditions between main thread and status thread. On further look, that appears impossible. Multiple UI threads is hard to get right and almost always avoidable but we inherited this design.

You are right, the GetProp null check is more trouble than worth it. Will remove.

lstipakov (approved):

Issue #592, opened 2023-01-26
Starting connection with single VPN profile doesn't work


lstipakov:
Steps to reproduce:

 - have a single VPN profile
 - click on tray icon -> Connect

Expected results:

 - connection starts

Actual results:

 - nothing happens

Same goes with "disconnect". However, if you open status window after connect attempt, it starts to connect. See the screencast.

Problem doesn't reproduce with more than one VPN profile.

![gui_bug](https://user-images.githubusercontent.com/643110/214829570-2f649614-ed9e-457a-80e3-0e5e575401e1.gif)

lstipakov:
Also it works with 2.5.8, /me is bisecting.

lstipakov:
This one.

```
94179911682503e916b9d105f100c85964f711be is the first bad commit
commit 94179911682503e916b9d105f100c85964f711be
Author: Selva Nair <selva.nair@gmail.com>
Date:   Sat Jan 14 10:34:11 2023 -0500

    Use a list instead of array for connections list
```

PR #593, opened 2023-01-26
Fix broken menu functionality with single profile

lstipakov:
To get connection for menu command, we use "dwMenuData" property of a menu. With single connection profile we read "dwMenuData" of a main menu, not a submenu. There has been a bug in code which hasn't set "dwMenuData" for the main menu but it has worked until recently, since this property stored index of connection array and index 0 has always worked.

Since commit 94179911 ("Use a list instead of array for connections list") we have switched to a linked list and store list pointer in dwMenuData. However due to bug dwMenuData has always being 0, and logic which fetches connections doesn't work:

        minfo.fMask = MIM_MENUDATA;
        GetMenuInfo((HMENU) lParam, &minfo);
        c  = (connection_t *) minfo.dwMenuData;
        if (!c)
            break; /* ignore invalid connection */

Fix by assigning main menu's dwMenuData to a value of submenu's dwMenuData in case of we have single connection - in that case we have submenu's actions (connect/disconnect) in a main menu. This way we can fetch connection from main menu's dwMenuData.

Fixes https://github.com/OpenVPN/openvpn-gui/issues/592

### DIFF
--- a/tray.c
+++ b/tray.c
@@ -169,7 +169,7 @@ CreatePopupMenus()
         hMenuConn[c->id] = CreatePopupMenu();
         /* Save the connection index in the menu.*/
         minfo.fMask = MIM_MENUDATA;
-        minfo.dwMenuData = (UINT_PTR) c;
+        minfo.dwMenuData = (ULONG_PTR) c;
         SetMenuInfo(hMenuConn[c->id], &minfo);
     }
     for (int i = 0; i < o.num_groups; i++)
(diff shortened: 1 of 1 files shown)


tray.c
```diff
@@ -182,13 +182,22 @@ CreatePopupMenus()
...
     minfo.fMask = MIM_STYLE;
     GetMenuInfo(hMenu, &minfo);
     minfo.dwStyle |= MNS_NOTIFYBYPOS;
     SetMenuInfo(hMenu, &minfo);
 
     if (o.num_configs == 1 && o.chead) {
+        /* Set main menu's menudata to first connection's submenu menudata */
+        minfo.fMask = MIM_MENUDATA;
+        GetMenuInfo(hMenu, &minfo);
+        minfo.dwMenuData = menuData;
```
flichtenheld, line 198: just use o.chead here. The result is the same and it is much more obvious what happens.

tray.c
```diff
@@ -182,13 +182,22 @@ CreatePopupMenus()
 
     hMenu = o.groups[0].menu; /* the first group menu is also the root menu */
 
+    /* Save menudata from submenu */
+    UINT_PTR menuData = minfo.dwMenuData;
```
flichtenheld, line 186: Setting this to "the last value of the loop" feels a bit hacky. See below for an alternative approach.

flichtenheld (changes requested):

flichtenheld (approved):

PR #599, opened 2023-02-07
Dialog owner, replacement for Sleep etc.

selvanair:
Fix for #576 and some related changes.

### DIFF
--- a/misc.c
+++ b/misc.c
@@ -1046,3 +1046,39 @@ RunAsAdmin(const WCHAR *cmd, const WCHAR *params)
     }
     return status;
 }
+
+/* Like sleep but service messages. If hdlg is not NULL
+ * dialog messages for it are checked. Also services
+ * MSG_FILTER hooks if caller wants further special processing.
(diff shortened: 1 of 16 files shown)


misc.c
```diff
@@ -1046,3 +1046,39 @@ RunAsAdmin(const WCHAR *cmd, const WCHAR *params)
...
+        {
+            MSG msg;
+            while(PeekMessage(&msg, NULL, 0, 0, PM_REMOVE))
+            {
+                if (msg.message == WM_QUIT)
+                {
+                    PostQuitMessage((int) msg.wParam);
+                    return false;
+                }
+                else if (!CallMsgFilter(&msg, MSGF_OVPN_WAIT)
```
lstipakov, line 1073: Sorry it took so long.

Could you explain what is this for?

I cannot find any existing hooks procedures, is this logic used?

Also, how dialog messages are served? We seem to call translate/dispatch if `hdlg` is `NULL` or the message is not dialog message. Is this intentional?

lstipakov (approved):

PR #606, opened 2023-02-24
additional token field in user/ password dialog

Timollo78:
We in our company have the need to add an additional totp-token field to the user/password authentication dialog depending on the configuration file. OpenVPN (or the plugin used for the 2 factor authentication?) expects a token to be sent attached to the password. In the GUI attaching the token to the password is not very intuitive and feels a bit unwieldy. I added an approach on how a separate token field could be displayed in the dialog depending on the client configuration. Therefore I added the possibility to add a specific comment as some kind of annotation to the configuration file. If the annotation is present, the token field appears in the user dialog, otherwise the behavior regarding the totp token remains unchanged.


schwabe:
I have not really reviewed the PR but we already have the `--static-challenge` configuration directive,  which seems to serve the need that you are having. I am currently not aware how well `--static-challenge` support is implemented in openvpn-gui but I rather would like to see the existing feature being (or an explaination why it does not work for your use case). And even if we have a new feature, it should not be specific to a single UI.

Timollo78:
Thank you for your very fast feedback  @schwabe. I`ll discuss the option to use the --static-challenge configuration directive with my colleagues on monday to figure out why it does not work for us.
I completely agree that a more general solution (if really needed) would be the best choice. Any support on that will be appreciated.

Timollo78:
Thank you very much for the feedback @selvanair. I`ll try to use the existing options.

Issue #607, opened 2023-03-09
Running command "disconnect_all" removes tray icon from taskbar


Markus74a:
I'm running Windows 11 22H2 and just switched from `OpenVPN-2.5.8-I604-amd64.msi` to the newest version `OpenVPN-2.6.1-I001-amd64.msi`. I often have multiple VPN connections open which I want to close all at once. 

Therefore, I execute: `"C:\Program Files\OpenVPN\bin\openvpn-gui.exe" --command disconnect_all`

- With `2.5.8-I604`, this command only disconnected all currently active OpenVPN connections.
- With `2.6.1-I001`, this command also does the disconnect, but it **additionally** removes the OpenVPN-GUI tray icon from the Windows 11 taskbar, while the `openvpn-gui.exe` process keeps running in the background.

Now I can't access the GUI any more since its tray icon is missing and I can't start the GUI again since it is detected as already running.

selvanair:
A fix is in PR #608

> Now I can't access the GUI any more since its tray icon is missing and I can't start the GUI again since it is detected as already running.

When tray goes missing, you can do "openvpn-gui.exe --command exit" to terminate the GUI and then start again. Sorry for the trouble.

PR #608, opened 2023-03-10
Fix --command disconnect_all

selvanair:
See also #607


lstipakov (approved):
Started at the code and tested new and old behavior - with this change tray icon doesn't disappear as a result of `disconnect-all` command.

Issue #595, opened 2023-01-26
tray icon should bear some visual similarity to desktop icon


matthewjameslord:
Hey there,

I wanted to follow up on the Trac ticket #1276 regarding the tray icon concept

I'm not TCT! I'm an employee at OpenVPN. 

I've attached an overview of the revised design, which includes five states: Idle, Idle with error, Connecting, Active, and Active with error. 
![SysTrayStates](https://user-images.githubusercontent.com/123665594/214926950-ad866433-83fa-46f4-aded-d5368c078abf.png)

You can also view these in Figma at the link provided.
https://www.figma.com/file/r5TAXjYHshNwFjoFltbm22/System-Tray-Icon?node-id=0%3A1&t=QqM1gSdEpphhDHYN-1

In addition, I've included zips containing the icons in PNG and SVG formats for all sizes.

[SysTrayIcons.zip](https://github.com/OpenVPN/openvpn-gui/files/10512572/SysTrayIcons.zip)
[SysTrayIconsPNG.zip](https://github.com/OpenVPN/openvpn-gui/files/10512573/SysTrayIconsPNG.zip)

Please take a look and let me know your thoughts on how we should proceed from here.

Thanks,
Matt (OpenVPN employee)

selvanair:
@matthewjameslord Thanks for these. 
A minor suggestion: what about moving the red dot to top right corner like overlays on icons for messages, email etc. used in iOS/android? While I get the idea of adding the dot at the end as a moniker for  "broken" connection, an overlay at the top may be better at drawing user's attention and more familiar?

Also, the solid dot could be replaced by a white "X" in red background? Not sure whether that would look too busy in a smallish tray icon.

selvanair:
A familiar warning sign like that (or error sign) at top left or right corner would be good. Wonder whether we can do this decoration as an overlay icon at run time... should be possible with taskbar icons, not sure of notification icons which are too small to start with.

fghzxm:
I personally would like to be able to distinguish all different states without relying on color.  The "Connecting" and "Active" states right now could be particularly problematic if you're colorblind.

matthewjameslord:
The dot at the bottom left is a familiar signal to alert user to problem in the system tray. If you use Slack, you will see they use the same concept with notifications and states: 
<img width="36" alt="image" src="https://user-images.githubusercontent.com/123665594/215557037-7d3179e6-ce20-4ec4-91dd-b6d3d969f9dd.png">

[...]

PR #612, opened 2023-03-16
WIP: Add a new set of icons for connection states and notifications

selvanair:
This is still WIP (icon files need some fixes), but I would appreciate feedback on some questions I have:

- Isn't it too intrusive to change icons in a minor or fixup release?
Knowing how hard such changes could be for some of users I interact with, I can empathize with admins who may complain.
- In this version I've retained both old and new icons and added an option to switch. It defaults to "legacy icons"  for now. Those who want to switch right away can do so by clicking a checkbox. Is this acceptable?

The new icons are as discussed in this thread: #595


selvanair:
Changes:
- Manually edited the new svg icon files to fix odd sizes and vew-ports (25x26 pixels --> 24x24 pixels etc.).  The graphics itself does not overflow those updated sizes, so no scaling is done.
- Updated Makefile.am
- The status window now shows "connected_error.ico" in case of route errors

Some issues still remain: the orange dots in "connecting.ico" are drawn with white "stroke" and orange "fill". Just setting the background to transparent does not remove those white lines, and  they look odd when displayed in dark background. Could be changed by editing the svg's but I'm not going to get into graphic design. If anyone wants to edit these the svgs (from #595) with my changes are in the zip file below.

[new-icons.zip](https://github.com/OpenVPN/openvpn-gui/files/11537365/new-icons.zip)

PR #615, opened 2023-03-26
Remove private-key password change feature

selvanair:
This removes dependency on OpenSSL

If there is a demand, we could add support for password change of pkcs12 files using Windows Crypto API.


CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ option(CLI_OVPN3 "Build ${PROJECT_NAME} with OpenVPN3 support" OFF)
 cmake_minimum_required(VERSION 3.10)
 
 if(NOT VCPKG_MANIFEST_DIR)
-    set(VCPKG_MANIFEST_DIR ${CMAKE_SOURCE_DIR}/vcpkg_manifests/openssl_3)
+    set(VCPKG_MANIFEST_DIR ${CMAKE_SOURCE_DIR}/vcpkg_manifests)
```
lstipakov, line 6: Since we do not need two separate manifests, no need put the only manifest into custom directory. 

 - move `vcpkg.json` file to the root source directory, where `vcpkg` searches for it by default
 - remove `vcpkg_manifests` directory
 - remove these lines from `CMakeLists.txt`

misc.c
```diff
@@ -752,31 +752,6 @@ ImportConfigFile(const TCHAR* source, bool prompt_user)
     RecreatePopupMenus();
 }
 
-void
-set_openssl_env_vars()
```
lstipakov, line 756: Let's remove unused declaration from `misc.h`.

lstipakov (changes requested):

selvanair:
I squashed the fixup commit, pushed here and merged, but it seems the PR is still open in its pre-squashed state. Not sure what got merged, if anything... Will wait for github to recover.

PR #617, opened 2023-04-01
Password reveal feature

selvanair:
A button to reveal masked password input is common these days, but it seems new style password box is only available in managed code (not in win32 API). This is an attempt to get something close to what one sees in modern UI. Suggestions for improvement welcome.

Following security features are included:
- pre-filled passwords (from saved/cached value) cannot be revealed
- when focus moves away from the password edit field, any text entered can no longer be revealed
  the user has to delete the current text and start over for the reveal "button" to be available again
- if a group policy setting has disabled password reveal in credential UI, it is respected

The images used for the reveal icon are my "poor man's" constructs using arcs and circles using SVG. Will be happy to substitute if anyone contributes better icons with a compatible license.


selvanair:
Screenshot from 2023-04-01 17-09-13

chipitsine:
Wow!

selvanair:
Good catch -- looks like its modelled after my eye :)
Actually I just manually wrote down some co-ordinates for the svg file -- may be the centre is off by a pixel (or half pixel?). Will check.

selvanair:
Is this better?

lstipakov (approved):
Looks good, apart from new icons lack transparency. I took a liberty to fix that, will open a separate PR.

selvanair:
Yeah, I forgot to set the transparent colour during conversion to ico. Thanks for fixing.

lstipakov:
@selvanair after merging your PR (with rebase and merge, without merge commit) I noticed that "fixup!" commits weren't squashed. Did I do it wrong?

selvanair:
Its not automatic. github can only squash all commits if you choose squash and merge. I usually do an --autosquash in my local machine after approval and force-push.

Issue #619, opened 2023-04-13
sometimes passphrase request window is not displayed


fixed77:
Bug in openvpn gui 2.6.1-2.6.2. Windows 10
key with passphrase.

faced with the problem that sometimes after entering the login and password, the interface does not request the passphrase of the key and hangs on it.
After much testing, we managed to reproduce the problem. when starting prime95 and trying to connect, the passphrase request window is not displayed.
![fail](https://user-images.githubusercontent.com/8656458/231758723-a3699713-57de-4686-b6de-3e8e103bfb2d.PNG)

client config: 
```
client
dev tun
dev-type tun
proto udp

remote remote.server 43333

setenv opt block-outside-dns
nobind
auth-user-pass
auth-nocache
reneg-sec 0
remote-cert-tls server
topology subnet
explicit-exit-notify

keepalive 10 60
ping-timer-rem
persist-tun
persist-key

cipher AES-256-GCM
data-ciphers AES-256-GCM:CHACHA20-POLY1305

<ca>
</ca>

<cert>
</cert>

<key>
# key with passphrase
</key>
```

using verb11 received logs from the client. On successful passphrase request:
```
2023-04-13 13:11:48 us=187000   pkcs11_cert_private = DISABLED
2023-04-13 13:11:48 us=187000   pkcs11_cert_private = DISABLED
2023-04-13 13:11:48 us=187000   pkcs11_cert_private = DISABLED
2023-04-13 13:11:48 us=187000   pkcs11_cert_private = DISABLED
2023-04-13 13:11:48 us=187000   pkcs11_cert_private = DISABLED
2023-04-13 13:11:48 us=187000   pkcs11_cert_private = DISABLED
2023-04-13 13:11:48 us=187000   pkcs11_cert_private = DISABLED
2023-04-13 13:11:48 us=187000   pkcs11_cert_private = DISABLED
2023-04-13 13:11:48 us=187000   pkcs11_cert_private = DISABLED
2023-04-13 13:11:48 us=187000   pkcs11_cert_private = DISABLED
2023-04-13 13:11:48 us=187000   pkcs11_cert_private = DISABLED
2023-04-13 13:11:48 us=187000   pkcs11_cert_private = DISABLED
2023-04-13 13:11:48 us=187000   pkcs11_pin_cache_period = -1
2023-04-13 13:11:48 us=187000   pkcs11_id = '[UNDEF]'
[...]

PR #621, opened 2023-04-17
Bring back delay before management is ready

lstipakov:
Commit 2b1e5867

  "Replace Sleep by a Wait function that pumps messages"

replaced Sleep() with a wait function which also processes messages. However WM_MANAGEMENT message in some cases requires actual delay to be processed. To acheve that, filter out WM_MANAGEMENT from message dispatching.

Fixes https://github.com/OpenVPN/openvpn-gui/issues/619

### DIFF
--- a/manage.c
+++ b/manage.c
@@ -345,7 +345,8 @@ OnManagement(SOCKET sk, LPARAM lParam)
                 else if (strncmp(pos, "INFO:", 5) == 0)
                 {
                     /* delay until management interface accepts input */
-                    OVPNMsgWait(100, c->hwndStatus);
+                    /* use real sleep here, since WM_MANAGEMENT might arrive before management is ready */
+                    Sleep(100);
                     c->manage.connected = 2;
                     if (rtmsg_handler[ready_])
                         rtmsg_handler[ready_](c, pos + 5);


misc.c
```diff
@@ -1038,7 +1038,7 @@ OVPNMsgWait(DWORD timeout, HWND hdlg)
         if (MsgWaitForMultipleObjectsEx(0, NULL, end - now, QS_ALLINPUT, MWMO_INPUTAVAILABLE) == WAIT_OBJECT_0)
         {
             MSG msg;
-            while(PeekMessage(&msg, NULL, 0, 0, PM_REMOVE))
+            while(PeekMessage(&msg, NULL, 0, WM_MANAGEMENT - 1, PM_REMOVE))
```
selvanair, line 1041: I haven't tested, but it looks like this may affect exit processing. We wait for openvpn.exe to disconnect during exit, waiting here for up to 5 seconds. If management events are not processed during that wait, we'll not get OnStop() called.

Simplest option may be to revert to Sleep() in

```
else if (strncmp(pos, "INFO:", 5) == 0)
{
    /* delay until management interface accepts input */
    OVPNMsgWait(100, c->hwndStatus);
...
```
A 100 msec sleep may not appear  a stall in the UI. Also its limited to a single status thread, not the main thread.

selvanair (commented):
Good analysis.

lstipakov:
Hi,Yeah Sleep(100) also works, that was my first solution. But I wanted to do something more comprehensive and came up with this one, but didn’t think of all possible implications. Will change to Sleep. Lähetetty iPhonestaSelva Nair ***@***.***> kirjoitti 17.4.2023 kello 15.39:﻿
@selvanair commented on this pull request.

Good analysis.

In misc.c:
> @@ -1038,7 +1038,7 @@ OVPNMsgWait(DWORD timeout, HWND hdlg)
         if (MsgWaitForMultipleObjectsEx(0, NULL, end - now, QS_ALLINPUT, MWMO_INPUTAVAILABLE) == WAIT_OBJECT_0)
         {
             MSG msg;
-            while(PeekMessage(&msg, NULL, 0, 0, PM_REMOVE))
+            while(PeekMessage(&msg, NULL, 0, WM_MANAGEMENT - 1, PM_REMOVE))

I haven't tested, but it looks like this may affect exit processing. We wait for openvpn.exe to disconnect during exit, waiting here for up to 5 seconds. If management events are not processed during that wait, we'll not get OnStop() called.
Simplest option may be to revert to Sleep() in
else if (strncmp(pos, "INFO:", 5) == 0)
{
    /* delay until management interface accepts input */
    OVPNMsgWait(100, c->hwndStatus);
...

A 100 msec sleep may not appear  a stall in the UI. Also its limited to a single status thread, not the main thread.

selvanair:
On looking further, what I pointed out may be a non-issue: the wait during exit is in the main thread while WM_MANAGEMENT is handled by the connection-thread. That said, Sleep() after receiving INFO: is simpler and arguably more logical than unconditionally suspending the message in OVPNMsgWait.

selvanair (approved):

PR #630, opened 2023-05-25
CMakeLists.txt: generate PDB for release builds

lstipakov:


CMakeLists.txt
```diff
@@ -158,6 +158,11 @@ if(MSVC)
         RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/$<0:>
         LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/$<0:>
         ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/$<0:>)
+
+    target_compile_options(${PROJECT_NAME} PRIVATE "$<$<CONFIG:Release>:/Zi>")
+    target_link_options(${PROJECT_NAME} PRIVATE "$<$<CONFIG:Release>:/DEBUG>")
```
selvanair, line 163: I'm not sue, but we may want /DEBUG:FULL here so that the pdb can be used for debugging outside the build machine as well? We do not have to distribute the pdb, but could make it available.

lstipakov: Yeah makes sense.

CMakeLists.txt
```diff
@@ -158,6 +158,11 @@ if(MSVC)
         RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/$<0:>
         LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/$<0:>
         ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/$<0:>)
+
+    target_compile_options(${PROJECT_NAME} PRIVATE "$<$<CONFIG:Release>:/Zi>")
+    target_link_options(${PROJECT_NAME} PRIVATE "$<$<CONFIG:Release>:/DEBUG:FULL>")
+    target_link_options(${PROJECT_NAME} PRIVATE "$<$<CONFIG:Release>:/OPT:REF>")
```
flichtenheld, line 164: There should be no need to have this as three lines with `target_link_options(${PROJECT_NAME} PRIVATE` repeated. `target_link_options` takes a list. Also you could probably even combine them into one `$<$<CONFIG:Release>:/DEBUG:FULL /OPT:REF /OPT:ICF>` (although I haven't tested that)

selvanair (approved):

flichtenheld (approved):

Issue #445, opened 2021-08-23
coding style


selvanair:
Moving to a new thread

Hi,

On Mon, Aug 23, 2021 at 01:42:46AM -0700, Lev Stipakov wrote:
> > Does our windows code not follow our coding style wrt "no conditionals without {} brackets"? 
> 
> openvpn-gui style is somewhat mixed - I've seen examples with and without brackets.

I consider Selva the current final authority on OpenVPN gui coding
style, so it's not my decision.

If I *were* to decide, I'd ask for all new code to be written according
to "main openvpn" coding style (which especially means "always {} brackets"),
possibly amended by "the way windows is coded" (// comments or such).

But this is strictly not a topic for this PR, but for a separate discussion.

gert

_Originally posted by @cron2 in https://github.com/OpenVPN/openvpn-gui/issues/436#issuecomment-904019407_

selvanair:
Yes the "coding style" here, if at all we can call it a style, is all over the place. Partly due to following the same style as in the context, partly laziness, and mainly for not imposing a style. 

I'm okay with just following the same style as OpenVPN core from now own. We do not really need "//" for comments -- easier to stick to a documented style than making exceptions.

cron2:
Hi,

On Mon, Aug 23, 2021 at 01:01:19PM -0700, Selva Nair wrote:
> Yes the "coding style" here, if at all we can call it a style, is all over the place. Partly due to following the same style as in the context, partly laziness, and mainly for not imposing a style. 
> 
> I'm okay with just following the same style as OpenVPN core from now own. We do not really need "//" for comments -- easier to stick to a documented style than making exceptions. 

So let's do this, then :-)

How to document this?

The https://community.openvpn.net/openvpn/wiki/CodeStyle page already
has a section about "Windows specific", so maybe just adding a remark
there about "OpenVPN GUI: for all new code, the same style as for
openvpn core applies" might suffice...  what do you think?

gert

-- 
"If was one thing all people took for granted, was conviction that if you 
 feed honest figures into a computer, honest figures come out. Never doubted 
 it myself till I met a computer with a sense of humor."
                             Robert A. Heinlein, The Moon is a Harsh Mistress

Gert Doering - Munich, Germany                             ***@***.***

selvanair:
Depends on whether we do a "Grand Reformat" or not. If we do not reformat, we'll have to allow some exceptions to match the context as required for readability.

cron2:
[...]

PR #637, opened 2023-07-07
Uncrustify

flichtenheld:
Since we didn't have a lot of activity in the last weeks, maybe it is a good time to address #445 ?

Closes: #445


selvanair:
Skimming through the sources, all look good.  Also compared by generating the assembly code before after this grand-reformatting: all match except for a few instances where we print `__LINE__` as expected.

selvanair (approved):

PR #646, opened 2023-08-29
Build improvements

lstipakov:


selvanair (commented):
This looks good. Would it be useful to make ASAN builds the default for GHA?

lstipakov:
I would just add ASAN configuration to GHA jobs, but for that one needs to look closer on matrix/jobs thing.

CMakeLists.txt
```diff
@@ -29,10 +29,14 @@ add_executable(${PROJECT_NAME} WIN32
...
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /guard:cf /Qspectre")
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /guard:cf")
+if (NOT ASAN)
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /Qspectre")
+endif ()
+
 set(CMAKE_EXE_LINKER_FLAGS  "${CMAKE_EXE_LINKER_FLAGS} /guard:cf /DYNAMICBASE")
 
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /wd4996 /wd4267 /wd4244")
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /wd4996 /wd4267 /wd4244 /wd5072")
```
flichtenheld, line 39: I really think we should document what these numbers mean here. Sure, I can google it. But a comment would be better.

CMakeLists.txt
```diff
@@ -29,10 +29,14 @@ add_executable(${PROJECT_NAME} WIN32
 
 set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -DDEBUG")
 
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /guard:cf /Qspectre")
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /guard:cf")
+if (NOT ASAN)
```
flichtenheld, line 33: If you're using an option/variable here, you should declare it to get proper documentation. But I'm not sure a variable is the correct approach here? In OpenVPN 3 we used a build type instead for ASAN. Any reason not to do that?

CMakeLists.txt
```diff
@@ -132,6 +136,8 @@ target_compile_definitions(${PROJECT_NAME_PLAP} PRIVATE
     WIN32_LEAN_AND_MEAN
     HAVE_CONFIG_H)
 
+set_target_properties(${PROJECT_NAME_PLAP} PROPERTIES LINK_FLAGS " /ignore:4302 ")
```
flichtenheld, line 139: again, please document what "4302" means

flichtenheld: Just a question, but why are we using `set_target_properties` instead of the dedicated options?

flichtenheld (changes requested):

CMakeLists.txt
lstipakov: I see, but `CMAKE_BUILD_TYPE` is not used with MSVC generator. Looks like I need to look into CMAKE_CONFIGURATION_TYPES.

CMakeLists.txt
lstipakov: I changed the way how ASAN is enabled and this is not needed anymore.

CMakeLists.txt
lstipakov: Changed to `target_link_options`.

flichtenheld (approved):

PR #649, opened 2023-08-30
localization: shortern tray icon tooltip for RU

lstipakov:


lstipakov:
ping @chipitsine

chipitsine:
I'm not sure why do we have a space for some languages

image


(also, I think this should be "OpenVPN GUI" for all languages without customization, but ok, as a first step we can just cut down russian translation)

chipitsine (approved):

selvanair:
As @chipitsine pointed out, I think we need a trailing space as we append the connected message to this when there are active connections. Not sure why not all language strings have that space.

Just replacing IDS_TIP_DEFAULT by "OpenVPN GUI " in tray.c without localization may be better.

lstipakov:
Yeah makes sense, I'll do the change.

lstipakov:
Let's use PACKAGE_NAME instead of hardcoded "OpenVPN GUI".

selvanair (approved):

Issue #656, opened 2023-10-09
"Could not parse --management option" when using config-auto folder.


Samg381:
When moving my **working** OpenVPN config + certificates from the `config` to the `config-auto` folder, I receive the following error after clicking connect:

![Screenshot_8](https://github.com/OpenVPN/openvpn-gui/assets/3127698/b6c6c3d5-2d18-42de-9eec-9b6cbaf4c9fc)

This should function no differently than it does in the `config` folder, with the only difference being connection at system startup.

cron2:
If you want to access configs in `auto` from the GUI, they need to have a `management` statement in them - otherwise the GUI won't know where to find them.

For normal "started by GUI", the GUI will setup `management` when running `openvpn`, but for auto the ordering is different so the config also need to be different (*iff* you want GUI access).  For many users the whole point of having the config in `auto` is "no clicking, no GUI", though.

Samg381:
Thanks for getting back. Could you elaborate on the `management` statement? Should I simply place that in the configuration, or is more configuration required?

I am trying to have my OpenVPN GUI connect to my server automatically when the system boots up. I like the taskbar icon, as it allows me to easily disconnect if I need to, along with letting me know the connection is active. If I add the `management` statement, will that achieve my goal?

cron2:
Adding `management 127.0.0.1 41000` should be sufficient ("listen on localhost, Port number 41000").  If that port is used by something else - "any random number between 1024 and 65535" will do the job.

For added security (so "no random process" can connect to the management interface and interfere) a password can be set - which needs to go to a file ("just one line with a password") which is then referenced as `management 127.0.0.1 41000 mysecret.txt`.  If no path is specified, I assume that "putting it into the `config-auto` directory will be the place where it is looked for.

(I'm not the resident expert on this, @selvanair is, but I *think* I got this right ;-) )

Samg381:
That did the trick! Thanks a ton!

PR #657, opened 2023-10-09
Add clarity for error on missing management paramater.

Samg381:
Configuration files located in `config-auto` require a management parameter to be declared. I added an additional line of detail in the error message to this effect. This error has historically caused confusion, and this additional messaging will clarify the fact that a management interface is _required_ to resolve the error.


Samg381:
Linking #656

selvanair:
+1 to improving this message, but wording could be better: What about "Attaching to auto-started connections require --management option in the config file" or something similar that conveys that this is necessary only if controlling via the GUI is desired.

Samg381:
Agreed and updated.

selvanair (approved):

PR #658, opened 2023-10-10
Update Italian language

bovirus:
@selvanair 

Please check and merge, Thanks.

### DIFF
--- a/res/openvpn-gui-res-it.rc
+++ b/res/openvpn-gui-res-it.rc
@@ -411,7 +411,8 @@ Per maggiori dettagli consulta il file log (%ls) ."
     IDS_NFO_SERVICE_STOPPED "Servizio OpenVPN arrestato."
     IDS_NFO_ACTIVE_CONN_EXIT "Ci sono ancora connessioni attive che verranno chiuse se esci dall'interfaccia di OpenVPN.\
 \n\nSei sicuro di voler uscire?"
-    IDS_ERR_PARSE_MGMT_OPTION "Impossibile analizzare l'opzione --management in <%ls%ls>."
+    IDS_ERR_PARSE_MGMT_OPTION "Impossibile analizzare l'opzione --management in <%ls%ls>. \
+L'allegato alla connessione automatica richiede l'opzione --management nel file config."
     IDS_NFO_SERVICE_ACTIVE_EXIT "Al momento sei connesso (il servizio di OpenVPN è in funzione). \
 Se esci dall'interfaccia di OpenVPN rimarrai comunque connesso.\n\n\
 Vuoi continuare e uscire dall'interfaccia di OpenVPN?"


res/openvpn-gui-res-it.rc
```diff
@@ -411,7 +411,8 @@ Per maggiori dettagli consulta il file log (%ls) ."
     IDS_NFO_SERVICE_STOPPED "Servizio OpenVPN arrestato."
     IDS_NFO_ACTIVE_CONN_EXIT "Ci sono ancora connessioni attive che verranno chiuse se esci dall'interfaccia di OpenVPN.\
 \n\nSei sicuro di voler uscire?"
-    IDS_ERR_PARSE_MGMT_OPTION "Impossibile analizzare l'opzione --management in <%ls%ls>."
+    IDS_ERR_PARSE_MGMT_OPTION "Impossibile analizzare l'opzione --management in <%ls%ls>. \
+L'allegato alla connessione automatica richiede l'opzione --management nel file config."
```
ordex, line 415: Hi @bovirus , what is "L'allegato alla connessione automatica" ? Does it come from the English version? (I haven't checked it yet)

selvanair: English version is 
`IDS_ERR_PARSE_MGMT_OPTION "Could not parse --management option in <%ls%ls>. \
Attaching to auto-started connections require --management option in the config file."`

The second line was added yesterday which prompted this update.

ordex: Thanks @selvanair , then this translation is wrong. The word "allegato", being used in the proposed translation, means "(email) attachment" and this is not the meaning that "attach" has in this context.

To be honest I can't think of what the correct translation should be right now.
@bovirus in this context you should read "Attaching" more like "Collegarsi" or "Legarsi", but none of them would work imho. Maybe you have a better suggestion?

ordex: @selvanair shouldn't the English version be "requireS" rather than "require"?

selvanair: yes :)

selvanair: > Thanks @selvanair , then this translation is wrong. The word "allegato", being used in the proposed translation, means "(email) attachment" and this is not the meaning that "attach" has in this context.
> 
> To be honest I can't think of what the correct translation should be right now. @bovirus in this context you should read "Attaching" more like "Collegarsi" or "Legarsi", but none of them would work imho. Maybe you have a better suggestion?

"Attaching to auto-started connections" could be interpreted as "Controlling auto-started connections using the GUI" if that helps in translation.

ordex: > > Thanks @selvanair , then this translation is wrong. The word "allegato", being used in the proposed translation, means "(email) attachment" and this is not the meaning that "attach" has in this context.
> > To be honest I can't think of what the correct translation should be right now. @bovirus in this context you should read "Attaching" more like "Collegarsi" or "Legarsi", but none of them would work imho. Maybe you have a better suggestion?
> 
> "Attaching to auto-started connections" could be interpreted as "Controlling auto-started connections using the GUI" if that helps in translation.

Yeah! I think that helps. @bovirus how about starting from this one?
@selvanair imho that'd be a better English text too. The average Joe is less likely to understand "Attaching" here.

selvanair: This is shown when the user clicks "connect" in the GUI menu against a config that is started by the "openvpnservice" at boot. So somehow indicating that `--management` option is required in this case unlike connections started and stopped by the GUI is an improvement. "Attaching", "Controlling" or something else all refer to what the user is trying to do: which is "try to see the connection status using the GUI". So I think any of those words would work here.

Issue #668, opened 2024-02-09
Tray icon and text not updated in some cases


selvanair:
When a persistent/prestarted connection is connected from the GUI, the tray icon does not change colour nor update the tip text until the connection is completed. The "connecting" state does not reflect in the tray icon. Works correctly for non-persistent connections.

To reproduce, have a persistent connection profile in the GUI menu, disconnect it and reconnect. The tray icon does not change to yellow. The icon does turn green when connection is completed.
![screen1](https://github.com/OpenVPN/openvpn-gui/assets/3981391/2d0b9da5-74af-42b9-a873-eb1427135506)

PR #669, opened 2024-02-10
Always update tray icon on state change

selvanair:
The tray icon and its tip text get updated to the connecting state when starting a new status window thread. This is not enough for persistent connections as these can be restarted from the hold state which does not go through a new thread creation.

To see the bug, disconnect and reconnect a persistent connection and have it wait in the connecting state. The tray icon colour does not change to yellow and the tray tip message does not show the "Connecting to:" text.

Fix by ensuring that  CheckAndSetTrayIcon() or SetTrayIcon() gets called whenever a state change message is received.

Fixes issue #668

### DIFF
--- a/openvpn.c
+++ b/openvpn.c
@@ -438,6 +438,10 @@ OnStateChange(connection_t *c, char *data)
             SetStatusWinIcon(c->hwndStatus, ID_ICO_CONNECTING);
         }
     }
+    else
+    {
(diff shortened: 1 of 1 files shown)


cron2:
Looks good to me (looked at some other uses of `CheckAndSetTrayIcon()` and at the function itself).

Issue #666, opened 2024-01-17
Can't see ip address in tray tooltip


raddyst:
When i connect with https://github.com/zhovner/zaborona_help/blob/master/openvpn-client-config/srv0.zaborona-help-UDP-no-encryption_maxroutes.ovpn profile, 

![image](https://github.com/OpenVPN/openvpn-gui/assets/8584929/86484e0a-41f9-4b7d-95d5-707e4e530403)

i can't see ip address in tray tooltip.

![image](https://github.com/OpenVPN/openvpn-gui/assets/8584929/58112905-f968-49fd-8319-df73d6ab27df)

ps. With other profiles - all ok, win7 x64

cron2:
@selvanair do we have a message buffer for the whole message which fills up for over-long profile names?

selvanair:
> When i connect with https://github.com/zhovner/zaborona_help/blob/master/openvpn-client-config/srv0.zaborona-help-UDP-no-encryption_maxroutes.ovpn profile,
> 

The total text length allowed in Windows "notifyicon" tooltip is limited to 128 characters so truncation happens when the profile name is too long. As the balloon message has more space (256 characters) and less information, truncation is less likely there. 

Double click on tray icon to open the status window and see the full IP address.

@cron2 
> @selvanair do we have a message buffer for the whole message which fills up for over-long profile names?

As space s limited, we do truncate some strings used to build up the tooltip, but in this case the truncation is due to Windows limit. 

When there are multiple profiles connected, we try to list all connected/connecting profiles truncating both lists at ~100 characters. The total message would still get truncated by Windows to 128. When there is only one profile connected (the case above), we add "IP address" and "connected since" info as well. 

In the latter case, to avoid truncating the IP, we could truncate the profile name to about 32 characters but that works for v4 addresses only.

selvanair:
PR #670 proposes a fix to this using a custom tooltip window to show the text. If the text is too long [*], the profile name part is truncated to ensure the IP address and time will be always displayed in full.

Tested on Win10 with status bar at bottom, right, top and left. Untested on multiple monitors.

[*] I set it at 500 characters but could be made longer. The built-in tooltip is restricted to 128 characters which is too small to show the IP, time and anything more than a short profile name.

lstipakov:
Hi,

On Windows 11 there are two issues;

 - When there is no connection, there is no tooltip at all. Currently we display `OpenVPN GUI` there.
[...]

PR #670, opened 2024-02-10
Use a custom tooltip window for the tray icon

selvanair:
Built-in tray notification icon has a tip text length limit of 128 characters which is often limited for showing the connected profile name, connected since time and IP addresses. If the profile name is long the IP numbers could get truncated.

Fix by using a custom tooltip window and display it when mouse hovers over the icon. As the status bar need not be at the bottom of the screen (could be at right, left or top as well), the location of the window is chosen based on the mouse co-ordinates that trigger the hover event.

In case of errors while setting up the tooltip window, fall back to the current behaviour.

If the message is too long to include time and IP, truncate the profile name part of the message.

Fixes issue #666

Edit: force pushed after fixing some typos in commit message and buggy text truncation logic.

old

Screenshot from 2024-02-10 17-11-34

new

Screenshot from 2024-02-10 17-14-56

### DIFF
--- a/tray.c
+++ b/tray.c
@@ -39,6 +39,11 @@
 #include "localization.h"
 #include "misc.h"
 
+#ifndef GUID_NULL
+#include <initguid.h>
+DEFINE_GUID(GUID_NULL, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0);
+#endif
+
 /* Popup Menus */
 HMENU hMenu;
 HMENU *hMenuConn;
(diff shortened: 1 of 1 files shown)


cron2:
I like the idea, but do not feel qualified today to ACK or NAK the code. @lstipakov something for you?

lstipakov:
Hi,

On Windows 11 there are two issues;

 - When there is no connection, there is no tooltip at all. Currently we display `OpenVPN GUI` there.
 - When there is active connection, tooltip is displayed on top left. Currently we display it next to the tray.
Näyttökuva 2024-02-12 110020

Näyttökuva 2024-02-12 110139

selvanair:
linking to my comment https://github.com/OpenVPN/openvpn-gui/issues/666#issuecomment-1938927344

selvanair:
Force pushed with changes:
 - For some reason wParam does not contain the position in NIN_POPUOPEN message. Fixed by using GetCrursorPos() to determine the mouse location. 
 - Also fixed the empty message bug.

selvanair:
(Moving this to the right place)
Pushed two commits:
(i) Remove title and prepend it to the message
(ii) Position the message at a fixed distance above the icon instead of tracking the mouse position
(figured how to do this :)

On Win11 the tip window should probably have rounded corners, but will leave it at that.

lstipakov (approved):
Tested on Windows 11. Both tooltip and code look good.

Issue #548, opened 2022-11-29
Openvpn GUI crashes. The connection remains.


MDcod:
Drops randomly. Didn't notice the patterns. The tray icon disappears and the OpenVPN GUI for Windows process disappears.
Network connectivity remains good. But now there is no way to break the connection. Just restart the computer. I don't see gui logs anywhere, only the network itself.

> OS: Windows 11 Pro 21H2 (22000.1219), also crashed on earlier versions
> OpenVpn: (v2.5.8), also crashed on earlier versions
> OpenVpn GUI: (v11.31.0.0), also crashed on earlier versions

Please help, it's incredibly annoying

MDcod:
Maybe I can give some more useful information about it?

selvanair:
> Maybe I can give some more useful information about it?

Yes please !

selvanair:
OpenVPN logs may help for a start. When the GUI crashes leaving openvpn.exe running, go to `OpenVPN\log` in your user profile folder and look for a file named `<foo>.log`  where `<foo>` is the name of the connection that was started. Also look into Windows event logs.

Alternatively, when the GUI crashes, got to task-manager and kill the openvpn.exe process left behind. That will terminate the VPN connection -- no need to reboot. Then restart the GUI and from its menu open the log file of the connection that crashed (connection-name->View Log). Upload it somewhere and post a link here.

Another thing to do is to narrow down the issue to a particular OpenVPN/OS version combination. Your initial post says
 
> OS: Windows 11 Pro 21H2 (22000.1219), also crashed on earlier versions
> OpenVpn: (v2.5.8), also crashed on earlier versions
> OpenVpn GUI: (v11.31.0.0), also crashed on earlier versions

That's not helpful. If every version is crashing on all versions of the OS, what can I say.

mduivcap:
I'm having the same issue as the poster mentioned but with a newer version.

The app does crash when I try to connect en when I try to disconnect.
In between it usually does not crash.
Unfortunatelly I cannot end the task (company policy).

The logfile:

> 2023-03-10 15:10:46 WARNING: Compression for receiving enabled. Compression has been used in the past to break encryption. Sent packets are not compressed unless "allow-compression yes" is also set.
[...]

PR #680, opened 2024-04-17
Fix off-by-one error in escape_string()

selvanair:
Github: fixes openvpn/openvpn-gui#548


lstipakov (approved):
Tested that 
```
    char* s = escape_string("hello ");
    free(s);
```
causes heap corruption (we move more bytes than needed), and this indeed fixes it.

PR #691, opened 2024-06-19
Validate the service pipe to avoid connecting to a bogus pipe

selvanair:
If an attacker with SeImeprsonatePrivilege manages to create a namedpipe server with a name matching that used by the "Interactive Service", the GUI connecting to it could allow the attacker to impersonate the GUI user.

Fix by validating the service pipe by comparing the pid of the pipe server with that of the "Interactive Service".

Note: GetNamedPipeServerProcessId() returns the pid of the process that created the first instance of the pipe. So, this patch only guards against a rogue pipe instance created before the service has started. This has to work in combination with a patch for the service that disallows creation of additional pipe instances when the service is running.

CVE: CVE-2024-4877
Reported by: Zeze with TeamT5 <zeze7w@gmail.com>

Acked-by: Lev Stipakov <lstipakov@gmail.com>

### DIFF
--- a/openvpn.c
+++ b/openvpn.c
@@ -60,6 +60,7 @@
 #include "env_set.h"
 #include "echo.h"
 #include "pkcs11.h"
+#include "service.h"
 
 #define OPENVPN_SERVICE_PIPE_NAME_OVPN2 L"\\\\.\\pipe\\openvpn\\service"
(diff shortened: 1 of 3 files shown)


selvanair:
changes since last review:
- formatting fixes to pass uncrustify

cron2:
Looks good to me, and matches the patch proposed and ACKed by @lstipakov  on the security@ list.

PR #693, opened 2024-07-10
openvpn.c: handle strdup failure

chipitsine:


openvpn.c
```diff
@@ -1417,6 +1417,11 @@ OnPassword(connection_t *c, char *msg)
             param->flags |= FLAG_CR_TYPE_SCRV1;
             param->flags |= (*(chstr + 3) != '0') ? FLAG_CR_ECHO : 0;
             param->str = strdup(chstr + 5);
+            if (!param->str)
+            {
+                WriteStatusLog(c, L"GUI> ", L"Error: Out of memory - ignoring dynamic challenge", false);
```
selvanair, line 1422: This is actually static challenge :) That said, param->str is later used with a `if (param->str) { <display the sttring> }` so failure here only leads to the challenge prompt string in the dialog displayed as empty, not a fatal error.

That said, erroring out here like this is fine too -- once out of memory the process is going to die sooner or later anyway.

Issue #697, opened 2024-07-21
Blurry eye icons when UI scaling is set


inelisoni:
Eye icons are blurry when UI scaling is set (125%, 150%, 175%, or even 200%).
I believe this is because icons are only available in 16x16 size. For example, tray icon is available in sizes of 16x16, 20x20, 24x24, and 32x32.

125%
![image](https://github.com/user-attachments/assets/0a2eb2af-059e-4f84-aa3c-1947307b1a5b)

150%
![image](https://github.com/user-attachments/assets/b4fb8bc8-3206-4711-9778-2e13d0d0eeb4)

175%
![image](https://github.com/user-attachments/assets/525f5845-0041-4afc-9009-c930eea9013a)

200%
![image](https://github.com/user-attachments/assets/6f962175-c819-497e-b5d1-965e6bb2467b)

selvanair:
@inelisoni could you please test whether the GUI executable from the link below work better? https://github.com/selvanair/openvpn-gui/actions/runs/10030273814#artifacts

This one has eye icons with 16, 20, 24 and 32 bit variants.

inelisoni:
Yes, it looks much better now. If necessary, I can provide screenshots.

PR #698, opened 2024-07-22
Add higher resolution eye icons

selvanair:
20, 24 and 32 bit eye icons added.

Fixes issue #697


lstipakov (approved):
Checked that new .ico files indeed contain 16x16, 20x20, 24x24 and 32x32 icons.

PR #699, opened 2024-08-19
Update openvpn-gui-res-ua.rc

toxi22:


lstipakov:
May I ask why you closed this?

PR #701, opened 2024-09-09
Support static-challenge "concat" option

selvanair:


openvpn.c
```diff
@@ -706,9 +706,19 @@ UserAuthDialogFunc(HWND hwndDlg, UINT msg, WPARAM wParam, LPARAM lParam)
                         {
                             SaveAuthPass(param->c->config_name, password);
                         }
+                        if (param->flags & FLAG_CR_TYPE_CONCAT)
+                        {
+                            GetDlgItemTextW(hwndDlg, ID_EDT_AUTH_CHALLENGE, password + wcslen(password), _countof(password)-wcslen(password));
+                            SetDlgItemTextW(hwndDlg, ID_EDT_AUTH_PASS, password);
+                            /* erase potentially secret contents in the response text box */
+                            memset(password, L'x', wcslen(password));
+                            SetDlgItemTextW(hwndDlg, ID_EDT_AUTH_CHALLENGE, password);
```
cron2, line 715: not sure I understand this code here - the `GetDlgItemTextW()` seems clear to me ("if we have CONCAT, read the text field from the window and put it after the existing content in `password` = concat).

What I'm unclear on is the `SetDlgItemTextW()` part - so we put the concatenated password into the ÌD_EDT_AUTH_PASS` field of the window, and a full set of `x` (of length password+challenge) into the `ID_EDT_AUTH_CHALLENGE` field?

lstipakov: I guess that's because later in `ManagementCommandFromInput()` we get the value of password+auth_challenge from `ID_EDT_AUTH_PASS` field - so we need that field to contain those concatenated values. And we do erase the content in `ID_EDT_AUTH_CHALLENGE` because `ManagementCommandFromInput()` does it only for `ID_EDT_AUTH_PASS`.

selvanair: Yes, as lev said.  This allows one to reuse the password submission code by re-writing the password filed with password+OTP. A less "hacky" way would be to add a new function "ManagementComamndFromTwoInputs()" duplicating most of "ManagementCommandFromInput()"

cron2: Yeah, thanks.  I only half-understood this - I saw "ok, it's appending the CR to the password, to be used later down" and then totally overlooked that `password[]` is actually x-ed out right after :-) - makes sense.

LGTM.

Issue #663, opened 2023-12-04
Textbox for 2fa security code


nognomar:
Could you please add an optional additional text box for one-time passwords for cases where connecting to the VPN requires a combination of a permanent password and OTP? It would be convenient to have the option to save the permanent part of the password and only input the OTP.

schwabe:
That seems like your setup uses a non-standard way to request 2FA and having it as part of the password. The normal OpenVPN methods (static-challenge, dynamic-challenge and crtext) do not need to split password method.

nognomar:
@schwabe This is the second company where I've encountered such an approach. Moreover, some VPN clients, like Tunnelblick, implement this functionality, and when I was using a mac, it was a great help to me. I think that since the authors of Tunnelblick have implemented support for combining a password and OTP, this functionality is still in demand, even though it's not encountered frequently.

zsolks:
I have the same problem with one company. All the time we used OpenVPN GUI for Windows to connect to them remotelly. Everything was fine.
Now they enabled 2FA. During loging to them I always see info AUTH_FAILED. There is no window to put 2FA code, no popup window.
So we have to use Stormshield SSL VPN Client to connect to them, but we prefer OpenVPN GUI

schwabe:
@zsolks that seems to be more that those products seem to blatenly ignore the normal methods to use 2FA with OpenVPN.

cron2:
client logfile or it did not happen... 2FA in a combined password (secret+token) works just fine, you just need to ensure that you do not check "[x] save password" in GUI

cron2:
@schwabe well... so do we... it's an old setup, and the OTP handling is done in the PAM stack, so it wants "secret+token in the password field".  Of course it could be modernized, but since it works as is and users know how to get in, I never felt like investing time.

schwabe:
The best way we can probably do is to have a --static-challenge flag or similar that produces the password+otp token format instead of the SCRV1:<password_base64>:<response_base64> that is normally used when you use static-challenge

zsolks:
> @zsolks that seems to be more that those products seem to blatenly ignore the normal methods to use 2FA with OpenVPN.

I do not know, but Stormshield works. :/

Im, not admin and never setup openvpn server yet so I, can't polemize with youre technical stuff.
But i know, that client-company give me files to OpenVPN like:
```
CA.cert.pem
config.ini
openvpnclient.cert.pem
openvpnclient.pkey.pem
[...]

PR #702, opened 2024-09-09
Add a global option to always prompt for OTP

selvanair:
This is more like a quick solution for user requests such as #663.
This is built on top of #701, by adding a global option to always prompt for OTP and submit it concatenated with password. Submitting as a separate PR as I'm not fully convinced of the need for this.

The newly extended static-challenge config option (GUI support is in PR #701)  that can optionally specify concatenation format should be preferred over this. Always prompting for OTP could confuse users when none are required.  But the former will be available only in OpenVPN 2.7 while this is immediately usable.

### DIFF
--- a/localization.c
+++ b/localization.c
@@ -640,6 +640,10 @@ GeneralSettingsDlgProc(HWND hwndDlg, UINT msg, UNUSED WPARAM wParam, LPARAM lPar
             {
                 Button_SetCheck(GetDlgItem(hwndDlg, ID_CHK_AUTO_RESTART), BST_CHECKED);
             }
+            if (o.auth_pass_concat_otp)
+            {
+                Button_SetCheck(GetDlgItem(hwndDlg, ID_CHK_CONCAT_OTP), BST_CHECKED);
+            }
 
             break;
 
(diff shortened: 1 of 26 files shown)


selvanair:
If you review this please ignore #701 as it's a subset of this.

selvanair:
Anyone thinks this is useful? After hearing that tunnelblick has had such a feature for years I'm more sympathetic to users request for this.

lstipakov (approved):
Looks good. I compiled it and tested that OTP prompt indeed appears if checkbox is set.

selvanair:
>At work, all our VPN connections that use OTP require it to be added at the beginning of the password.

This is why I was extremely reluctant in adding this option. Over the years I have seen reports of a number of ad-hoc implementations that send OTP appended to password, and we finally added support for it although I consider it a hack. I'm hearing about "prepending" for the first time, but I'm not surprised. Soon there will be some who use a ":" to separate the two and so on.

The correct approach is to do server side implementation correctly by following the static challenge protocol. If that is not an option, lobby for a new protocol in OpenVPN core and we will support it in the GUI. Right now the core supports the original static-challenge protocol and concatenation option. So does the GUI.

r4dius:
Ahah, I'm no VPN expert, I thought both would be standard practice. I'd rather have both options than none 🤣

cron2:
Someone has to maintain all the different code paths for "options", and test them.

Much better is to adhere to a standard, and not invent new standards and require changes to pre-existing software to make that work.

selvanair:
> Ahah, I'm no VPN expert, I thought both would be standard practice. I'd rather have both options than none

Neither are standard practice but unnecessary hacks invented by server administrators. 

Anyway, this menu item to prompt for OTP is an interim solution until OpenVPN 2.7  is out. Once 2.7 is released, the correct way to use the concatenation option is to add `--static-challenge` directive to the client config file with flag set to 2 or 3 to indicate "concatenation" as the protocol. Adding new meaning to this setting that is not in the specifications is not acceptable. In fact, I think your patch would break `--static-challenge` option processing.

So, please follow the protocol as documented in management-notes.txt in OpenVPN 2 code base.

PR #705, opened 2024-09-19

justwho:


res/openvpn-gui-res-zh-hans.rc
```diff
@@ -561,7 +561,7 @@ Wintun驱动程序将无法正常工作。"
     IDS_NFO_AUTO_CONNECT    "在 %u 秒后自动连接..."
     IDS_NFO_CLICK_HERE_TO_START "OpenVPN GUI 已经运行. 右击任务栏图标启动."
     IDS_NFO_BYTECOUNT "接收字节: %ls  发送字节: %ls"
-    IDS_NFO_OTP_PROMPT "Input OTP or passcode"
+    IDS_NFO_OTP_PROMPT "输入OTP或密码"
```
selvanair, line 564: @justwho  Thanks for the PR.

I asked a native speaker for comments and got an OK, but I have a concern: 
Where this is used, the dialog asks for password (密码) and  OTP with this pompt. Wouldn't the use of the same word "密码" in both places confuse users?

If there is no good translation for "passcode",  would it be better to show this prompt as just "输入OTP" ?

justwho: I also think using '输入OTP' is more appropriate.Thanks!!! @selvanair

selvanair: @justwho Could you please add a commit on top of this with that change? I can then squash during merge.

justwho: OK！ @selvanair

PR #712, opened 2024-11-23
Changing OTP prompt option to allow for "Prepend OTP" method too

r4dius:
This is a modification of the OTP prompt introduced in [#702](https://github.com/OpenVPN/openvpn-gui/pull/702).

It replaces the OTP prompt checkbox with radio buttons, allowing users to choose between the original "Append OTP," "Prepend OTP," and "Disable" options.

Screenshot 2024-11-23 223718


selvanair:
See my comment in #702.

PR #718, opened 2025-01-15
Switch to clang-format

flichtenheld:
The uncrustify tool is relatively unstable, the configuration is difficult to understand.

Switch to clang-format instead, which has better support in terms of maintenance and integrations.

The clang-format configuration I have developed to be as close to old "OpenVPN" style as enforced by uncrustify as possible. While switching openvpn2 itself to it is a bigger project, I think this project should be easier.

I have skimmed over the reformat commit and didn't see any real cases where it made the code worse, but feel free to point out any specific instances.


selvanair:
This does do a few arguably weird things: E.g.,
(i) split function arguments to one per line in some places (e.g., line ~151, ~260 in access.c), but consolidate them to one line in other (line ~230 in access.c) 
(ii) line break right after opening bracket in some function calls (e.g., line ~385 in access.c) 
But I do not have a strong opinion. Works for me.

I compared the resulting object files --- no differences found other than expected from the use of __LINE__ and assert() in a few places. 

BTW: 
(i) ubuntu 22.04 has `clang-format` version14 which does not support `AlignConsecutiveMacros`
(ii) This PR created a new branch named `clang-format` here -- is that by mistake?

lstipakov:
VS supports clang format natively, so it is definitely an improvement compared to uncrustify.

lstipakov (approved):

flichtenheld:
> This does do a few arguably weird things: E.g., (i) split function arguments to one per line in some places (e.g., line ~151, ~260 in access.c), but consolidate them to one line in other (line ~230 in access.c) (ii) line break right after opening bracket in some function calls (e.g., line ~385 in access.c) But I do not have a strong opinion. Works for me.

Are you sure these line numbers are from access.c? I do not see the examples you point to there?

> I compared the resulting object files --- no differences found other than expected from the use of `__LINE__` and assert() in a few places.
> 
> BTW: (i) ubuntu 22.04 has `clang-format` version14 which does not support `AlignConsecutiveMacros`

Yeah, in other projects we tried to work with the clang-format from distros, but that is very limiting, in how old that always ends up being. And while clang-format is generally stable, it still sometimes changes behavior between versions, e.g. due to bug fixes. This is why I did go the route of adding the pre-commit support, which allows you to specify a specific clang-format version that it will then always use. I think it installs it via a pypi package? Would need to check. This way works for me, for people with IDEs they will need to check how to get the latest clang-format. But definitely no to using the clang-format from a specific Ubuntu version. 

> (ii) This PR created a new branch named `clang-format` here -- is that by mistake?

Hmm, yeah, I could have used a fork. I had the permissions so I did it without much thought.

access.c
```diff
@@ -151,15 +150,16 @@ AddUserToGroup(const WCHAR *group)
...
 
 #ifdef DEBUG
-    if (status == (DWORD) -1)
+    if (status == (DWORD)-1)
     {
         PrintDebug(L"RunAsAdmin: failed to execute the command [%ls %ls] : error = 0x%x",
-                   cmd, params, GetLastError());
+                   cmd,
+                   params,
+                   GetLastError());
```
selvanair, line 158: Here is a case where it splits args each into a new line.

access.c
```diff
@@ -229,22 +229,20 @@ AuthorizeConfig(const connection_t *c)
...
-            MessageBoxW(NULL, L"Failed to determine process owner SID", L""PACKAGE_NAME, MB_OK);
+            MessageBoxW(NULL, L"Failed to determine process owner SID", L"" PACKAGE_NAME, MB_OK);
         }
         return FALSE;
     }
     groups = GetProcessTokenGroups();
-    if (IsUserInGroup(sid, groups, admin_group)
-        || IsUserInGroup(sid, groups, o.ovpn_admin_group))
+    if (IsUserInGroup(sid, groups, admin_group) || IsUserInGroup(sid, groups, o.ovpn_admin_group))
     {
```
selvanair, line 238: and here it combines args into one line.
There is some logic to it, and I won't fret too much about it.

access.c
```diff
@@ -385,8 +386,8 @@ IsUserInGroup(PSID sid, const PTOKEN_GROUPS token_groups, const WCHAR *group_nam
...
         DWORD nread, nmax;
         LOCALGROUP_MEMBERS_INFO_0 *members = NULL;
-        err = NetLocalGroupGetMembers(NULL, group_name, 0, (LPBYTE *) &members,
-                                      MAX_PREFERRED_LENGTH, &nread, &nmax, &resume);
+        err = NetLocalGroupGetMembers(
+            NULL, group_name, 0, (LPBYTE *)&members, MAX_PREFERRED_LENGTH, &nread, &nmax, &resume);
```
selvanair, line 390: Here is an example of line break after opening "("

selvanair (approved):
My line number indications were poor. Marked some inline this time.
That said, LGTM given uncrustify is more of a pain.

access.c
flichtenheld: Yeah, one can play around with the different penalty values like `PenaltyBreakBeforeFirstCallParameter` to influence this. And obviously the `ColumnLimit` plays a huge role here.

PR #725, opened 2025-01-21
Avoid Compiler warnings

flichtenheld:
Try to avoid various compiler warnings both in MinGW and MSVC builds.


openvpn.c
```diff
@@ -167,19 +167,18 @@ void
...
-    if (flags - 1 == NULL)
+    char *flags = strchr(line, ',');
+    if (flags == NULL)
     {
         return;
     }
 
-    message = strchr(flags, ',') + 1;
-    if (message - 1 == NULL)
+    char *message = strchr(flags + 1, ',');
```
lstipakov, line 180: Looks like we got off-by-one error here

`Tue Feb 18 14:26:09 2025 ,Timers: ping 8, ping-restart 40`

lstipakov (changes requested):

openvpn.c
flichtenheld: Fixed

lstipakov (approved):
I haven't tested MinGW, but stared at the code and tested msvc build - looks good (as before).

Issue #729, opened 2025-02-19
removed config files do not disappear from the GUI list


cron2:
So, @lstipakov made me fight windows OpenVPN to test the DCO/server functionality, and I was copying .ovpn files all around the place and then cleaning up stuff again... and I found something which I think is a bug.

This happens with the GUI bundled in our official 2.6.13-I001 installer (I002 is not on swupdate yet), so no private builds involved yet.  No upgrade installation either, removed the old install and installed fresh, with default options.

There are a number of .ovpn files already in c:\program files\openvpn\config, and openvpn-gui.exe is running.

Add a new .ovpn file, right-click GUI, new file is visible.

Remove an .ovpn file (either the new one or one of the others), right-click GUI, freshly-removed config is still visible - trying to start it will of course cause an error (`IService > OpenVPN exited with error: exit code = 1`).

I do not understand enough windows coding to figure out how the GUI notices that "new files have appeared" and why it does not notice "files have been removed".

lstipakov:
I am on it.

lstipakov:
```
commit 6932c5e7109330d8cc75afca8980c80e457ef275
Author: Selva Nair <selva.nair@gmail.com>
Date:   Wed Jul 20 20:58:04 2022 -0400

    Change the logic of releasing persistent connections in OnHold

    (i)
    State is changed to detached before auto-starting
    so that OnHold() will see state = resuming and keep the hold.

    State is set to disconnected instead of detached on detach
    so that manual starts will release the hold automatically.

    End result: While connecting automatically, do not release if
    management-hold is on. But while started manually, release
    from hold so that connection can complete without further
    user action.

    In normal use of automatic service, one would not add management
    hold into the config. However, if the user disconnects the connection
    the GUI puts it on hold, and we do not want to auto-start it after a
    lock-unlock or some other automatic action.

    (ii)
    Also, currently, for persistent connections, the status
    window is not shown automatically which feels unnatural in
    real use. Instead, popup the status window when connection
    is manually initiated. Its not popped up when automatically
    attached to or if silent_connection is on.

    Only persistent connections are affected by the change.

    fixup: config file list is not recreated from scratch when
    enable_persistent == 2 (auto attach mode) to avoid losing info
    such as auto_connect = false on detached connections.

    Signed-off-by: Selva Nair <selva.nair@gmail.com>
```

cron2:
[...]

PR #732, opened 2025-02-19
Disable config in menu listing if its ovpn file becomes inaccessible

selvanair:
Since the introduction of persistent connections, we no longer recreate the entire config menu but only add newly added connection profiles during each rescan. This leaves any deleted configs actively displayed in the menu until the GUI is restarted.

Improve this situation by graying out entries corresponding to profiles not readable from file system. If the file reappears, the item gets automatically enabled again.

Also, if the corresponding connection is active, the item is not grayed out to allow the user to disconnect or reconnect it. Otherwise the corresponding OpenVPN core process will hang around with no way to control it from the GUI. Here "active" includes those on management-hold in case of persistent connections, as those can be started even if the underlying config file has gone missing.

Addresses github issue #729

### DIFF
--- a/openvpn_config.c
+++ b/openvpn_config.c
@@ -280,6 +280,15 @@ ActivateConfigGroups(void)
             cg->active = true;
             cg = PARENT_GROUP(cg);
         }
+        /* also deactivate any configs that are no longer readable */
+        if (CheckReadAccess(c->config_dir, c->config_file))
+        {
+            c->flags &= ~FLAG_CONFIG_DISABLED;
+        }
+        else
+        {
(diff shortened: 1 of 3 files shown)


selvanair:
This may not be the best "fix", but its certainly an easy one. Suggestions for improvement most welcome.

cron2:
I like the approach.  This will make it clear that "you can't run this anymore (now)" while keeping the other options.

Will look more closely tomorrow.  Thanks ;-)

selvanair:
> Will look more closely tomorrow. Thanks ;-)

:)

lstipakov (approved):
Checked the code and tested, looks good!

PR #738, opened 2025-04-05
Italian language update

bovirus:


ordex (approved):

ordex:
The PR should probably be rebased, but as for the introduced change, everything looks good.

PR #739, opened 2025-04-06
Hide password reveal eye icon when no response box is shown

selvanair:
When CR_TEXT or dynamic CR prompt is triggered with 'response not required' ('R' not present in flags), we hide the response box, but not the password reveal eye icon. Fix it.

Avoids this lonely eye:
lonely-eye


lstipakov:
Indeed single eye icon looks a bit creepy.

lstipakov (approved):

PR #744, opened 2025-04-21
Keep PLAP dialog alive by simulating user activity

lstipakov:
Add a timer to periodically call SetThreadExecutionState and simulate mouse movement using SendInput, preventing the pre-logon UI from being dismissed due to inactivity during mobile QR code authentication.


selvanair (commented):
This looks good and works!
It appears call to SetThreadExecutionState() is not  strictly necessary as simulating mouse activity is automatically detected by the system and keeps the screen active.

lstipakov:
@selvanair Thanks, will check and send v2. Meanwhile would you mind to look at https://gerrit.openvpn.net/c/openvpn/+/943 ?

selvanair (approved):

PR #753, opened 2025-06-10
add ability to save pin for smartcards

tantra35:
This patch adds ability to save smartcards(i.e. yubikey) pins

### DIFF
--- a/openvpn.c
+++ b/openvpn.c
@@ -843,7 +843,6 @@ GenericPassDialogFunc(HWND hwndDlg, UINT msg, WPARAM wParam, LPARAM lParam)
         case WM_INITDIALOG:
             param = (auth_param_t *)lParam;
             TRY_SETPROP(hwndDlg, cfgProp, (HANDLE)param);
-            BOOL lenableOKBtn = FALSE;
 
             WCHAR *wstr = Widen(param->str);
             if (!wstr)
@@ -978,7 +977,7 @@ GenericPassDialogFunc(HWND hwndDlg, UINT msg, WPARAM wParam, LPARAM lParam)
             else
             {
                 /* disable OK button until response is filled-in */
-                EnableWindow(GetDlgItem(hwndDlg, IDOK), lenableOKBtn);
+                EnableWindow(GetDlgItem(hwndDlg, IDOK), FALSE);
                 ResetPasswordReveal(GetDlgItem(hwndDlg, ID_EDT_RESPONSE),
                                     GetDlgItem(hwndDlg, ID_PASSWORD_REVEAL),
                                     0);


tantra35:
@Thanks for your review, i fixed as i think all you  remarks, also i want to coment something:

>I'm not sure whether saving a smartcard/token PIN is a good idea. Many tokens lock-up if a wrong PIN >is entered multiple times. Its easy for this to happen if we use a saved PIN and somehow miss to >promptly clear it on failure.

Yeah you right but for example [viscosity](https://www.sparklabs.com/viscosity/) client have ability to save pin, so i think openvpn-gui also must have this...

>We can intercept PIN prompts only when a token is used through pkcs11-helper or pkcs11-provider. On >windows, its much better to use the cryptoapicert option which can cover certificates/keys in tokens as >well. In that case PIN prompt is triggered by Windows itself and not seen by OpenVPN-GUI.

I don't think using windows cryptoapi is a good idea, since the dialog it shows is completely different from the openvpn gui, and may confuse users - in other words, they may not understand what it is all about.

openvpn.c
```diff
@@ -880,18 +880,79 @@ GenericPassDialogFunc(HWND hwndDlg, UINT msg, WPARAM wParam, LPARAM lParam)
...
             else if (param->flags & FLAG_PASS_TOKEN)
             {
                 SetWindowText(hwndDlg, LoadLocalizedString(IDS_NFO_TOKEN_PASSWORD_CAPTION));
                 SetDlgItemText(hwndDlg,
                                ID_TXT_DESCRIPTION,
                                LoadLocalizedString(IDS_NFO_TOKEN_PASSWORD_REQUEST, param->id));
+
+                if (param->c->flags & FLAG_DISABLE_SAVE_PASS) 
+                {
+                    ShowWindow(GetDlgItem(hwndDlg, ID_CHK_SAVE_PASS), SW_HIDE);
```
selvanair, line 894: Isn't it better to make this hiding of the save password checkbox the default outside of all if-else clauses? That will simplify the logic and have less deeply nested code.
Like, the following line could become `else if (!(param->c->flags & FLAG_DISABLE_SAVE_PASS) && RecallSmart...())`

selvanair (commented):
Have you confirmed that `c->failed_auth_attempts` does get incremented when token password fails?
PS. Also, shouldn't it be `c->failed_psw_attempts` ?

selvanair:
> Have you confirmed that `c->failed_auth_attempts` does get incremented when token password fails?

In fact, a wrong PIN causes pkcs11-helper to re-prompt for it three times without a restart. When a SIGUSR1 restart happens finally, its triggered as a generic "tls-error" leading to a further round of wrong PIN submission and so on.

This is not acceptable. So, NAK from me.

tantra35:
> > Have you confirmed that `c->failed_auth_attempts` does get incremented when token password fails?
> 
> In fact, a wrong PIN causes pkcs11-helper to re-prompt for it three times without a restart. When a SIGUSR1 restart happens finally, its triggered as a generic "tls-error" leading to a further round of wrong PIN submission and so on.

Thanks for you investigations, its very helpful for me(I not fully yet understand how full process works), and I have few questions:

1. if I understand all correctly openvpn it case of wrong ping just reprompt it throw management socket?
2. About restart I not fully understand, why you write about `SIGUSR1`? Its a *nix signal handler as I think...
3. If openvpn just reprompt pin within existent management connection, it's very easy enter additional counter for pin attempt in `connection` structure(https://github.com/OpenVPN/openvpn-gui/blob/master/options.h#L137), or just increment already existent counters `failed_psw_attempts` or `failed_auth_attempts`(not sure that this is acceptable), if so, it's looks like a solution

PR #770, opened 2025-10-23
Fix two off-by-one memory assignments

MegaManSec:
These bugs were found with ZeroPath.


openvpn.c
```diff
@@ -2188,7 +2188,7 @@ OnNeedOk(connection_t *c, char *msg)
     }
 
     /* allocate space for response : "needok param->id cancel/ok" */
-    resp = malloc(strlen(param->id) + strlen("needok \' \' cancel"));
+    resp = malloc(strlen(param->id) + strlen("needok \' \' cancel") + 1);
     wstr = Widen(param->str);
```
selvanair: Same here: the space gets replaced by param->id.

PR #772, opened 2025-10-27
Fix some issues found by ZeroPath

selvanair:


as.c
```diff
@@ -39,19 +39,35 @@
...
+    const wchar_t *special[] = { L"CON",  L"PRN",  L"AUX",  L"NUL",  L"COM1", L"COM2",
+                                 L"COM3", L"COM4", L"COM5", L"COM6", L"COM7", L"COM8",
+                                 L"COM9", L"LPT1", L"LPT2", L"LPT3", L"LPT4", L"LPT5",
+                                 L"LPT6", L"LPT7", L"LPT8", L"LPT9" };
+
+    for (DWORD i = 0; i < _countof(special); i++)
+    {
+        if (wcscmp(fname, special[i]) == 0)
+        {
+            wcsncpy_s(fname, wcslen(fname) + 1, L"TMP", _TRUNCATE);
```
lstipakov: I am a bit unsure on `+ 1` here. We should pass the capacity of the buffer,  not the property of the content. 

The callers of `SanitizeFilename` have the length og the buffer - maybe pass it to `SanitizeFilename`?

selvanair: +1.  v2 is coming.

selvanair: Done.

PR #776, opened 2025-11-02
Add a message source for event logging

selvanair:
- message resource is embedded into the exe

~TODO: openvpn-build: register openvpn-gui.exe while installing~

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,6 +79,15 @@ target_include_directories(${PROJECT_NAME} PUBLIC ${CMAKE_CURRENT_BINARY_DIR})
 set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /MANIFEST:NO")
 set(CMAKE_RC_FLAGS "/D \"HAVE_CONFIG_H\" /D \"_MSC_VER\"")
 
+add_custom_command(
+    OUTPUT  eventmsg.h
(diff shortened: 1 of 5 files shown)


cron2:
If I understood @lstipakov correctly, the reason why openvpnsrv.exe is using a separate .dll is because the event viewer will block restarting the program if embedded message resources are used... (I hope I understood that correctly).  Not sure this is relevant for the GUI, just wanted to point it out.

cron2:
I refrain from having an opinion here (not enough windows clue) - thanks for pointing out what options we have, and what consequences they have.  So I'm fine to go along with whatever you and @lstipakov agree on.  Having better event logging is certainly welcome ;-)

lstipakov:
> Restarting the GUI with event viewer left open works. But upgrades will be blocked. Even if we use a separate dll, upgrading that dll will be blocked by an open event viewer. 

Hm, I just tested the upgrade from beta2 to beta3 with Event Viewer being open and it went fine. I remember when I first put event source into .exe for the service, the upgrade got blocked - this was the reason I moved it to .dll.

> Or we could just point to `openvpnservmsg.dll` as the source and use the same event-id. Our message format is the same --- `%1%n%2`.

I would go this way.

selvanair:
> Hm, I just tested the upgrade from beta2 to beta3 with Event Viewer being open and it went fine. I remember when I first put event source into .exe for the service, the upgrade got blocked - this was the reason I moved it to .dll.

For me it came  up with "reboot to complete the upgrade".. It seems msiexec has a mind of its own.

> > Or we could just point to openvpnservmsg.dll as the source and use the same event-id. Our message format is the same --- %1%n%2.
> 
> 
> I would go this way.

Will do.

PR #780, opened 2025-11-03
Check correct management daemon path when OpenVPN3 is enabled

### LINKED ISSUE #779: 11.57.00 breaks (experimental) openvpn3 support on windows
i haven't triaged yet, but openvpn3 support no longer works on Windows. Client windows gets stuck `waiting for OpenVpnService to terminate`

facboy:
If OpenVPN3 is enabled then the check needs to verify against the `omi_exe_path` registry key, not the `exe_path` key.

Fixes #779.

### DIFF
--- a/misc.c
+++ b/misc.c
@@ -1457,7 +1457,12 @@ ValidateManagementDaemon(connection_t *c)
     wchar_t win32_path[MAX_PATH];
     swprintf(win32_path, _countof(win32_path), L"%ls%ls", L"\\\\?\\GLOBALROOT", nt_path);
 
+#ifdef ENABLE_OVPN3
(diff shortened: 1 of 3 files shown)


misc.c
```diff
@@ -1457,7 +1457,12 @@ ValidateManagementDaemon(connection_t *c)
     wchar_t win32_path[MAX_PATH];
     swprintf(win32_path, _countof(win32_path), L"%ls%ls", L"\\\\?\\GLOBALROOT", nt_path);
 
+#ifdef ENABLE_OVPN3
+    TCHAR* exe_path = o.ovpn_engine == OPENVPN_ENGINE_OVPN3 ? o.omi_exe_path : o.exe_path;
```
selvanair: Please use WCHAR or wchar_t instead of TCHAR for all new code. Anyway this needs a v2 to make clang-format happy.

PR #782, opened 2025-11-05
Update CHANGES.rst

selvanair:
Merge just before next version upgrade (11.58.0.0)


selvanair:
We struggle to co-ordinate this every time. Let's just merge this now so that this will be at least semi up-to-date when 2.7 is out. It has been a while since this was updated.

PR #784, opened 2025-11-12
Update config path validation code to match the service

selvanair:
We use this to show a dialog to fix the issue before validation failure in the service. None of this is critical, but required to be in sync with recent/upcoming changes in the service.

### DIFF
--- a/access.c
+++ b/access.c
@@ -168,24 +168,34 @@ AddUserToGroup(const WCHAR *group)
  * interactive service.
  */
 static BOOL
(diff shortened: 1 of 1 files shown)


lstipakov (approved):

lstipakov:
Please note that due to the use of the new API, this breaks compatibility with Windows 7. I think 2.6 still runs on Windows 7 (although we don't regularly test it). If we are going to use this as a submodule in openvpn-build 2.6 branch, then poor Windows 7 users are out of luck. @flichtenheld

PR #787, opened 2025-11-14
Check the return value of GetProp()

selvanair:
Most of GetProp() calls are replaced by a macro that logs a message and returns on error.

### DIFF
--- a/openvpn.c
+++ b/openvpn.c
@@ -695,7 +695,7 @@ UserAuthDialogFunc(HWND hwndDlg, UINT msg, WPARAM wParam, LPARAM lParam)
             break;
 
         case WM_COMMAND:
-            param = (auth_param_t *)GetProp(hwndDlg, cfgProp);
+            TRY_GETPROP(hwndDlg, cfgProp, param, FALSE);
             switch (LOWORD(wParam))
             {
                 case ID_EDT_AUTH_PASS:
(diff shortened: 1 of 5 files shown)


pkcs11.c
```diff
@@ -681,9 +684,9 @@ QueryPkcs11DialogProc(HWND hwndDlg, UINT msg, WPARAM wParam, LPARAM lParam)
             break;
 
         case WM_CLOSE:
-            c = (connection_t *)GetProp(hwndDlg, cfgProp);
-            StopOpenVPN(c);
             EndDialog(hwndDlg, wParam);
+            TRY_GETPROP(hwndDlg, cfgProp, c, TRUE);
+            StopOpenVPN(c);
```
lstipakov: Is StopOpenVPN() supposed to be here?

lstipakov: ah I am blind, we just changes the places of EndDialog and StopOpenVPN

openvpn.c
```diff
@@ -519,11 +516,8 @@ AutoCloseCancel(HWND hwnd)
 static void CALLBACK
 AutoCloseHandler(HWND hwnd, UINT UNUSED msg, UINT_PTR id, DWORD now)
 {
-    autoclose *ac = (autoclose *)GetProp(hwnd, L"AutoClose");
-    if (!ac)
-    {
-        return;
-    }
+    autoclose *ac;
+    TRY_GETPROP(hwnd, L"AutoClose", ac, );
```
lstipakov: Same as above.

openvpn.h
```diff
@@ -30,11 +30,25 @@
...
+        MsgToEventLog(EVENTLOG_ERROR_TYPE, L"%hs:%d SetProp returned null", __func__, __LINE__); \
         EndDialog(hwnd, IDABORT);                                                                \
         return false;                                                                            \
     } while (0)
 
+/* A macro to set lval = GetProp(hwnd, name) or log an error and return err_return */
+#define TRY_GETPROP(hwnd, name, lval, err_return)                                          \
+    do                                                                                     \
+    {                                                                                      \
+        HANDLE handle_##lval = GetProp(hwnd, name);                                        \
```
irwir: Inside a block, any reasonably unique name would do for a local variable. Concatenation here is an unnecessary complication.

PR #788, opened 2025-11-17
TRY_GETPROP: make it more pretty for VOID callers

lstipakov:
Because

  TRY_GETPROP(hwnd, cfgProp, c, );

looks a bit scary.

### DIFF
--- a/openvpn.h
+++ b/openvpn.h
@@ -35,8 +35,8 @@
         return false;                                                                            \
     } while (0)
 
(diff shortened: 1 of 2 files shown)


selvanair:
> Because
> 
> TRY_GETPROP(hwnd, cfgProp, c, );
> 
> looks a bit scary.

Doesn't look scary to me, but I'm fine with the change.

Issue #789, opened 2025-11-18
Openvpn-gui 11.58 non-work in Win7


one-way-ticket:
Openvpn-gui 11.58 from latest OpenVPN-2.6.16-I001-amd64.msi non-work in Win7.
Need api-ms-win-core-path-l1-1-0.dll, but this file does not exist in Win7.
Imported functions:
PathCchAddBackslash
PathCchCanonicalize
PathCchCombine

Openvpn-gui 11.56 from OpenVPN-2.6.15-I001-amd64.msi work fine in Win7.

cron2:
I have no strong opinion on this.  Windows 7 with 2.6.15 is still as unsupported as before, but if you feel changing to the other API is a good thing, I'm happy to merge whatever comes along :-) (and it will take out all the configure/CMakeFile things again...)

selvanair:
If Windows 7 is not a supported platform for 2.6.xx, we can keep this as is. 

I only wanted to point out that, as there is no `pathcch.lib` dependency in openvpn.exe v2.6, if there is a strong desire to have Windows 7 support, we could remove it from the GUI. I'm not keen on it.

nikita-leontiev:
Download [page](https://openvpn.net/community) states that Windows 7 is still supported.
No info about Windows 7 support dropping in 2.6.16 release notes.
Installer doesn't issue an error stating that OS is not supported.

PR #790, opened 2025-11-26
Remove dependence on pathcch.dll not in Windows 7

selvanair:
Fixes Github: OpenVPN/openvpn-gui#789


lstipakov (approved):
Looks good - basically we replaced Cch functions with somewhat less safe API, but it is okay - we added asserts and this is the only way if we want to support Windows 7. Slightly tested on Windows 11.

PR #791, opened 2025-11-27
Avoid delay in sending startup data to iservice

selvanair:
The interactive service now expects to receive the start up data soon after opening the pipe. Move potentially blocking AuthorizeConfig() to before opening the pipe.


selvanair:
We use the same GUI in both 2.6 and 2.7, and this should work with both old and new versions of the service.

PR #799, opened 2026-01-16
Update .gitignore to include bonding-specific build artifacts and add…

PointerSoftware:
… bonding subdirectory to CMakeLists.txt


cron2:
I guess this escaped, since none of the bonding stuff exists upstream...

PR #806, opened 2026-04-02
res: Major cleanup

flichtenheld:
There are many, many strings that are not used since a decade or more. Do a general cleanup.

Note: cleanup is incomplete. So created as a draft PR.


openvpn-gui-res.h
```diff
@@ -333,12 +316,7 @@
 #define IDS_LANGUAGE_NAME               1650
 
 /* Win32 Service Related */
-/* historic: #define IDS_ERR_OPEN_SCMGR_ADMIN        1701 */
```
selvanair, line 336: Last time when we removed some such defines, the consensus was to keep them commented out to avoid reuse (see #80 ). I do not personally feel that is necessary (i.e., reuse is fine), but just clarifying the origin of such comments, in case anyone cares to preserve them.

flichtenheld: I don't think these numbers are ever transmitted somewhere, are they? They are only used by the compiler? So I think it should be perfectly safe to reuse them.

selvanair (approved):
Wow, this is some cleanup! Looks good.

A future enhancement could be to extend the script to also look for missing entries in non-English resource files.

PR #808, opened 2026-04-03
Italian language update

bovirus:
Please check and merge. Thanks.


selvanair:
@bovirus  Please fix the CI build errors.

cron2:
> @bovirus Please fix the CI build errors.

I don't think the CI build errors can be blamed on this PR - as far as I can see, @bovirus is updating `res/openvpn-gui-res-it.rc`, while the build errors are from a (somewhat misguided?) attempt to block left-right-writing-order shenannigans, complaining about `openvpn-gui-res-fa.rc` (Farsi, I guess, which sounds like it would use this).

Now, why this is happening *now* and never before?  Github updates behind our backs?

@ordex can you have a look at the italian updates, please?

cron2:
> > @bovirus Please fix the CI build errors.
> 
> I don't think the CI build errors can be blamed on this PR - as far as I can see, @bovirus is updating `res/openvpn-gui-res-it.rc`, while the build errors are from a (somewhat misguided?) attempt to block left-right-writing-order shenannigans, complaining about `openvpn-gui-res-fa.rc` (Farsi, I guess, which sounds like it would use this).

... this said, there are different CI build errors here.  The msvc-x64 instance complains about

```
  Error: D:\a\openvpn-gui\openvpn-gui\plap\../res/openvpn-gui-res-it.rc(369): error RC2104: undefined keyword or key name: IDS_ERR_ONE_CONN_OLD_VER [D:\a\openvpn-gui\openvpn-gui\out\build\x64-ovpn3\libopenvpn_plap.vcxproj]
```

which very much is caused by `openvpn-gui-res-it.rc`.

The patch introduces two new strings, of which this is one.  @bovirus please fix this part (do not introduce new messages unless this is something that is actually missing in the italian translation), and we look at the farsi part.

selvanair:
> I don't think the CI build errors can be blamed on this PR

I was referring to the undefined key erros due to re-introduction of strings that we removed in a recent patch.  The error message in the mingw build is very cryptic though..
 
The one about bidirectional control characters in the farsi files are warnings that I've been ignoring for a while now...

cron2:
Thanks for explaining.  This was non-obvious to me at first, so might have confused others too (and the way Github presents "warnings" and "failures" isn't exactly helping)

ordex (approved):

ordex:
I re-opened the PR and looked at the changes. LGTM.
Although I wonder why we still carry around messages for OpenVPN pre-2.0 and 1.5/1.6 (but this is unrelated to this PR)

ordex:
Ah This might be about what @selvanair said..

ordex (changes requested):
Need to remove non-existent keys.

cron2:
> I re-opened the PR and looked at the changes. LGTM. Although I wonder why we still carry around messages for OpenVPN pre-2.0 and 1.5/1.6 (but this is unrelated to this PR)

you can't approve something that fails CI (due to introduction of two new keys that are unused)... :-)

ordex:
Since I couldn't modify this PR, I opened #811 with an edited commit, still authored by @bovirus , but without the non-existing keys.

Issue #809, opened 2026-04-10
"disable_save_passwords" setting also disables saving the username now


MaartenJB:
Hi,

I've been using "HKLM\Software\OpenVPN\disable_save_passwords" to disable password caching in the openvpn client. In "OpenVPN-2.6.14-I004-amd64" this option did what the name suggests: not saving the password. But after installing a newer version of the openvpn client now it also doesn't save the username.

"OpenVPN-2.6.14-I004-amd64" works as expected (only saves username)
"OpenVPN-2.6.19-I001-amd64" does NOT work as expected (doesn't save username or password)
"OpenVPN-2.7.0-I017-amd64" does NOT work as expected (doesn't save username or password)

It would be nice to have the option again to only disable saving passwords, and if this isn't a bug maybe have an extra option like "disable_save_credentials" for when you don't want the username and password to be saved.

selvanair:
Username is saved even if "disable_saved_passwords" is set. How exactly are you concluding that username is not saved: the auth dialog does not show up with pre-filled username, or something else?
One difference between 2.6.14 and 2.7 is that saved username is now encrypted in storage.

MaartenJB:
In my case the username is saved until I actually close the openvpn-gui. (so not just disconnect/connect)

Did you close the gui when testing?

selvanair:
There is no need to close the GUI --- just enter the username/password, connect, disconnect and reconnect: you should see the username field to appear pre-filled with previously used value.
As I asked above, tell us how you are concluding whether username is saved or not -- by looking at auto-fill of dialog or something else?

MaartenJB:
I know the username is saved when connecting, and yes, if you disconnect and connect again, the username appears pre-filled. But when you close the gui, and open it again, and try to connect, for me the username is then gone.

selvanair:
Thanks for the clarification: now I can reproduce this. 
Looks like a bug introduced when username storage was changed to encrypted -- `disable_save_passwords` causes any saved passwords to be deleted, and along with goes some extra data needed for decryption. This was okay when username was saved in plain, not any longer.

Please check whether executable here fixes this: https://github.com/OpenVPN/openvpn-gui/actions/runs/24290203788#artifacts

MaartenJB:
I've tested the fix, and this seems to work properly now. The username is saved after restarting the application and appears pre-filled when attempting a connect.

cron2:
[...]

PR #810, opened 2026-04-11
Do not delete entropy data when clearing passwords

selvanair:
This data is required for decrypting saved username. Fixes: issue #809


flichtenheld (approved):
LGTM

PR #819, opened 2026-06-29
New sorting option

aiv:
## Summary

VPN profiles are listed in the tray menu in the order returned by the
filesystem (`FindFirstFile`/`FindNextFile`). On NTFS this is usually
alphabetical, but it is not guaranteed - FAT/exFAT and network shares can
return creation order - and there was no explicit sorting anywhere.

This PR adds an opt-in setting to sort configurations alphabetically in the
menu, controlled by a new checkbox in **General → Preferences**. The option is
**off by default**, so existing behavior is unchanged unless the user enables it.

## Behavior notes

- The menu is rebuilt on every open (`RecreatePopupMenus` → `BuildFileList` →
  `CreatePopupMenus`), so toggling the checkbox takes effect immediately on the
  next menu open - no restart required.
- Toggling the option **off** fully reverts to the on-disk order, because each
  config keeps its immutable `c->id` (assigned at load time).
- `c->id` is never modified, so the management port mapping
  (`mgmt_port_offset + id`) and the per-config menu handles (`hMenuConn[id]`)
  remain intact.
- Sorting applies to config entries only. In the nested/grouped menu view,
  configs within each folder are sorted while the order of the folders
  themselves is left unchanged.

### DIFF
--- a/localization.c
+++ b/localization.c
@@ -655,6 +655,10 @@ GeneralSettingsDlgProc(HWND hwndDlg, UINT msg, UNUSED WPARAM wParam, LPARAM lPar
             {
                 Button_SetCheck(GetDlgItem(hwndDlg, ID_CHK_CONCAT_OTP), BST_CHECKED);
             }
+            if (o.sort_configs)
+            {
+                Button_SetCheck(GetDlgItem(hwndDlg, ID_CHK_SORT_CONFIGS), BST_CHECKED);
+            }
 
             break;
 
@@ -723,6 +727,8 @@ GeneralSettingsDlgProc(HWND hwndDlg, UINT msg, UNUSED WPARAM wParam, LPARAM lPar
                     (Button_GetCheck(GetDlgItem(hwndDlg, ID_CHK_AUTO_RESTART)) == BST_CHECKED);
                 o.auth_pass_concat_otp =
                     (Button_GetCheck(GetDlgItem(hwndDlg, ID_CHK_CONCAT_OTP)) == BST_CHECKED);
+                o.sort_configs =
+                    (Button_GetCheck(GetDlgItem(hwndDlg, ID_CHK_SORT_CONFIGS)) == BST_CHECKED);
 
                 SaveRegistryKeys();
 
--- a/openvpn-gui-res.h
+++ b/openvpn-gui-res.h
@@ -173,6 +173,7 @@
 /* General settings contd.. */
 
 #define ID_CHK_CONCAT_OTP               470
+#define ID_CHK_SORT_CONFIGS             471
 
 /*
  * String Table Resources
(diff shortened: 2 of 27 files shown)


selvanair:
(i) The patch feels "incomplete" as it would leave nested folder names unsorted. As a consistent sorting order would really matter only to those with many configs, this is not a minor defect.

(ii) While the argument about immutability of `c->id` feels vaguely familiar from memory, wont it be much easier to just make this the default and not configurable? Assuming vast majority of users will not see any difference. Then there is no re-ordering on the fly and no need to double check this wont cause any such side effects. Also no localization or UI changes needed, easier to review.

That said, is this complication worth it --- I have never seen NTFS return file list in non alphabetcial order. What fraction of users are affected?

aiv:
Thanks for the review!

**(i) Unsorted nested folder names**

That was a deliberate choice - I left the folders unsorted so they stay grouped at the top rather than being interleaved alphabetically. But I agree that for someone with many configs a fully consistent order is what they'd expect, so I'm happy to extend the sorting to the group/folder names as well. I'll update the patch to sort both configs and folders.

**(ii) Make it the default instead of an option**

Fair point. I went with an option mainly because I assumed I don't know the project as well as you do, and a non-default opt-in felt like the safer/less intrusive way to propose this. But if it were my call I'd actually prefer to make alphabetical sorting the default and non-configurable - for the vast majority of users it changes nothing, and as you note it removes the on-the-fly reordering, the side-effect concerns, and all the localization/UI churn, which makes the change much smaller and easier to review. This ties into the NTFS point below, which is really the motivation for the whole change.
I'm glad to rework it that way if you prefer.

**On NTFS ordering / who is affected**

You're right that NTFS itself returns entries in alphabetical order. The case that bit me, and the reason I ended up here, is a config folder kept on a cloud-sync service (Google Drive / Dropbox / OneDrive etc.) that mounts as a regular folder in Windows. Enumeration order there is *not* alphabetical, and the menu ends up in a fairly random order. With a sync folder full of profiles this gets quite messy in practice.

I'm attaching a screenshot showing the chaos :)

image

selvanair:
>That was a deliberate choice - I left the folders unsorted so they stay grouped at the top rather than being interleaved alphabetically.

Right, keeping the prestarted profiles at the top as we have now is important. Also global and local config folders ordered the same way as now. So, it may be better not to touch folder ordering at all.

> The case that bit me, and the reason I ended up here, is a config folder kept on a cloud-sync service (Google Drive / Dropbox / OneDrive etc.)

A very special use-case...
I'm very busy these days, do not expect a quick response from me if you resubmit..  A thorough review is needed, more so if we make this the default.  Hopefully some one else may show interest...

aiv:
No problem at all, I completely understand.

No rush on the review whatsoever - take your time. If it eventually lands, great, since IMHO it's a genuinely useful improvement for anyone whose config folder doesn't enumerate in a predictable order. 

Thanks again for taking the time to look at it!

PR #826, opened 2026-08-07
Add Hungarian translation resource fileAdd files via upload

Internetezoo:
Add Hungarian translation resource file


cron2:
please remove all commits that are not related to the hungarian translation from this branch, and re-push.

cron2:
this is not how it works.  We now have 6 commits, but we must have exactly *one* with the desired content.

Please read up on "squashing", and produce 1 single commit with just the translation, and force-push that.

Internetezoo:
[openvpn-gui-res-hu.rc.txt](https://github.com/user-attachments/files/30838153/openvpn-gui-res-hu.rc.txt)

